PR #41, opened 2021-04-27
Base field

mfrachet:
Storybook available on https://design-system-git-base-field-strapijs.vercel.app/?path=/story/field--default-story

These components aim to be building blocks for higher-level components such as `TextInput` but also the select, textarea, dropdown etc...

## APIs
Most simple case

```jsx
 <Field name="email">
  <Stack size={1}>
    <FieldLabel>Email</FieldLabel>
    <FieldInput type="text" placeholder="Placeholder" value={value} onChange={(e) => setText(e.target.value)} />
  </Stack>
</Field>
```

Most complex case

```jsx
 <Field name="password" hint="Description line" error={email.length > 5 ? 'Email is too short' : undefined}>
  <Stack size={1}>
    <Grid cols="auto auto 1fr" gap={1}>
      <FieldLabel>Password</FieldLabel>
      <Tooltip content="Content of the tooltip">
        <button
          style={{ border: 'none', padding: 0, background: 'transparent' }}
          aria-label="Information about the email"
        >
          <HelpIcon aria-hidden={true} />
        </button>
      </Tooltip>
      <a href="#" style={{ marginLeft: 'auto' }}>
        Details
      </a>
    </Grid>
    <FieldInput
      placeholder="example@domain.com"
      value={email}
      onChange={(e) => setEmail(e.target.value)}
      Action={
        <FieldAction aria-label="Show password">
          <Password aria-hidden={true} />
        </FieldAction>
      }
    />
    <FieldHint />
    <FieldError />
  </Stack>
</Field>
```

## With VoiceOver

Note for designers: the text field takes as much space as it's able to take. In our scenario, inside the application, this will be constrained by the grid (or any layout) mechanisms.

Usage of the Field components using VoiceOver

### DIFF
--- a/packages/strapi-design-system/src/Field/FieldAction.js
+++ b/packages/strapi-design-system/src/Field/FieldAction.js
@@ -5,8 +5,8 @@ export const FieldAction = styled.button`
   background: transparent;
   // TODO: Make sure to use the theme when it's ready
   font-size: 1.6rem;
-  margin-left: -100%;
-  transform: translateX(-${({ theme }) => theme.spaces[3]});
+  margin-right: ${({ theme }) => theme.spaces[3]};
+  margin-left: ${({ theme }) => theme.spaces[3]};
   width: auto;
   padding: 0;
   display: flex;
--- a/packages/strapi-design-system/src/Field/FieldAction.js
+++ b/packages/strapi-design-system/src/Field/FieldAction.js
@@ -1,0 +1,14 @@
+import styled from 'styled-components';
+
+export const FieldAction = styled.button`
+  border: none;
+  background: transparent;
+  // TODO: Make sure to use the theme when it's ready
+  font-size: 1.6rem;
+  margin-left: -100%;
+  transform: translateX(-${({ theme }) => theme.spaces[3]});
+  width: auto;
+  padding: 0;
+  display: flex;
+  align-items: center;
+`;
--- a/packages/strapi-design-system/src/Field/FieldInput.js
+++ b/packages/strapi-design-system/src/Field/FieldInput.js
@@ -1,0 +1,67 @@
+import React, { forwardRef } from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import { useField } from './FieldContext';
+import { Row } from '../Row';
+
+const Input = styled.input`
+  border: 1px solid ${({ theme, hasError }) => (hasError ? theme.colors.danger600 : theme.colors.neutral200)};
+  border-radius: ${({ theme }) => theme.borderRadius};
+  padding: ${({ theme }) => `${theme.spaces[3]} ${theme.spaces[4]}`};
+  color: ${({ theme }) => theme.colors.neutral800};
+  background: ${({ theme }) => theme.colors.neutral0};
+  font-weight: 400;
+  // TODO: Make sure to use the theme when it's ready
+  font-size: ${14 / 16}rem;
+  display: block;
+  width: 100%;
+
+  ::placeholder {
+    color: ${({ theme }) => theme.colors.neutral500};
+  }
+
+  :disabled {
+    color: ${({ theme }) => theme.colors.neutral500};
+    background: ${({ theme }) => theme.colors.neutral150};
+  }
+`;
+
+export const FieldInput = forwardRef(({ Action, ...props }, ref) => {
+  const { id, error, hint, name } = useField();
+
+  let ariaDescription;
+
+  if (error) {
+    ariaDescription = `field-error-${id}`;
+  } else if (hint) {
+    ariaDescription = `field-hint-${id}`;
+  }
+
+  const fieldId = `field-${id}`;
+  const hasError = Boolean(error);
+
+  return (
+    <Row justifyContent="space-between">
+      <Input
+        id={fieldId}
+        name={name}
+        ref={ref}
+        hasError={hasError}
+        aria-describedby={ariaDescription}
+        aria-invalid={hasError}
+        {...props}
+      />
+      {Action}
+    </Row>
+  );
+});
+
+FieldInput.displayName = 'FieldInput';
+
+FieldInput.defaultProps = {
+  Action: undefined,
+};
+
+FieldInput.propTypes = {
+  Action: PropTypes.node,
+};
--- a/packages/strapi-design-system/src/Field/FieldInput.js
+++ b/packages/strapi-design-system/src/Field/FieldInput.js
@@ -26,7 +26,7 @@ const Input = styled.input`
   }
 `;
 
-export const FieldInput = forwardRef(({ Action, ...props }, ref) => {
+export const FieldInput = forwardRef(({ action, ...props }, ref) => {
   const { id, error, hint, name } = useField();
 
   let ariaDescription;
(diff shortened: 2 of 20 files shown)


rgesulfo (approved):
LGTM! :)

packages/strapi-design-system/src/Field/FieldAction.js
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components';
+
+export const FieldAction = styled.button`
+  border: none;
+  background: transparent;
+  font-size: 1.6rem;
```
HichamELBSI, line 6: Can we add a "Don't forget to use the theme" please 🙏🏽

packages/strapi-design-system/src/Field/FieldAction.js
```diff
@@ -0,0 +1,13 @@
...
+  margin-left: -100%;
```
HichamELBSI, line 7: I hate negative margin 😂  can you explain, why do you need it, please?

packages/strapi-design-system/src/Field/FieldInput.js
```diff
@@ -0,0 +1,71 @@
...
+import { useField } from './FieldContext';
+
+const Input = styled.input`
+  border: 1px solid ${({ theme, hasError }) => (hasError ? theme.colors.danger600 : theme.colors.neutral200)};
+  border-radius: ${({ theme }) => theme.borderRadius};
+  padding: ${({ theme }) => `${theme.spaces[2]} ${theme.spaces[4]}`};
+  color: ${({ theme }) => theme.colors.neutral800};
+  background: ${({ theme }) => theme.colors.neutral0};
+  font-weight: 400;
+  font-size: ${14 / 16}rem;
```
HichamELBSI, line 13: Same here, can you add a "Use the theme" todo comment for the font values please 😄

packages/strapi-design-system/src/Field/FieldInput.js
```diff
@@ -0,0 +1,71 @@
...
+    color: ${({ theme }) => theme.colors.neutral500};
+    background: ${({ theme }) => theme.colors.neutral150};
+  }
+`;
+
+const InputWrapper = styled.div`
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+`;
```
HichamELBSI, line 31: What do you think about creating something like a Flex component to avoid those CSS duplication?

HichamELBSI (changes requested):

packages/strapi-design-system/src/Field/FieldAction.js
mfrachet: The Input itself is wrapped in a parent container called "InputWrapper". This input wrapper has "flex" display. We can represent the big picture as following:

Untitled-2021-03-27-2011

In order for the Icon to be inside the Input, it needs a way to be moved from the right edge. I know two ways to do that: make the InputWrapper relatively positioned, make the Action absolutely position, and use absolute coordinates or use a negative margin.

I honestly don't have a strong preference here.

Do you have a preferred way? Or even an option that I don't know about? 😊

packages/strapi-design-system/src/Field/FieldInput.js
mfrachet: I'm pretty sure I can fix this with a grid 🤔

maevalienard:
- It seems the Details link doesn't follow the design rules to me, does it? I'm not sure if this is the right moment to mention it though, let me know if this specific part will be improved in another iteration :D 
Capture d’écran 2021-04-29 à 16 39 19
- Also related to this complex input, the icon color seems to be the wrong one. It looks like a neutral800 but should actually be a neutral500 (it's currently a neutral300 in the design but since we changed the other colors for an accessibility matters, we have to stay consistent). 

- The height of the input is currently 34px in Storybook but should be 40px -> The padding top and bottom currently applied are 8px each and should be at 12px (or rem sorry about that, I just inspected the element and got this information). 

- About the password input, we're supposed to have an eye icon on the right of the input. Does it appear? (I have two different plugin taking this space, so it's hard to say but I feel like it hasn't been implemented yet)

mfrachet:
(@maevalienard this is a perfect moment, thanks for the feedbacks, will take a look tomorrow 😊 )

mfrachet:
> It seems the Details link doesn't follow the design rules to me, does it? I'm not sure if this is the right moment to mention it though, let me know if this specific part will be improved in another iteration :D

The Link component has not been implemented yet. This is more a representation of the "layout" of a Field. This Field is not a "final" input, but it will aim to provide building blocks for other inputs 😊 . So, that's "normal" the link is not valid here

> - Also related to this complex input, the icon color seems to be the wrong one. It looks like a neutral800 but should actually be a neutral500 (it's currently a neutral300 in the design but since we changed the other colors for an accessibility matters, we have to stay consistent). 

Since we have not taken a strong decision on this icon, I've decided to put a "raw" icon in a "raw button". Again, it's a matter of layouts and positioning, not the style for this element exactly

> The height of the input is currently 34px in Storybook but should be 40px -> The padding top and bottom currently applied are 8px each and should be at 12px (or rem sorry about that, I just inspected the element and got this information).

I've moved to 12px 👍🏻 . Thanks for the review 😄 

> About the password input, we're supposed to have an eye icon on the right of the input. Does it appear? (I have two different plugin taking this space, so it's hard to say but I feel like it hasn't been implemented yet)

Same about the two first points :). This only aims to show what it may look like with an icon in it (layout). Also, it seems we don't have the "eye" icon in the strapi-icons package (or in figma?) and this is the reason why it's not this one in the example 😊 

When building a real text input with a password, the "eye" will be there 😊 

---

Also, it means that we can replace the "information" icon and the icon at the end of the input with any kind of other actions (for maybe, other field needs 😊 )

Thank you for the review 🙏🏻

packages/strapi-design-system/src/Field/FieldAction.js
mfrachet: (InputWrapper is now a "row")

packages/strapi-design-system/src/Field/FieldInput.js
mfrachet: Done with `Row` 😭

maevalienard:
Looks amazing now, thanks for the changes @mfrachet and also thanks for all the explanations, that makes definitely sense! :D

maevalienard (approved):

packages/strapi-design-system/src/Field/FieldAction.js
HichamELBSI: TBH I prefer the `position relative` (I do it in the input UID), the margin is a bit weird when you inspect the dom (There is no impact in other elements). 

Screenshot 2021-04-30 at 10 56 48

packages/strapi-design-system/src/Field/FieldInput.js
```diff
@@ -0,0 +1,67 @@
...
+      />
+      {Action}
+    </Row>
+  );
+});
+
+FieldInput.displayName = 'FieldInput';
+
+FieldInput.defaultProps = {
+  Action: undefined,
```
soupette, line 62: Shouldn't it be null?

packages/strapi-design-system/src/Field/FieldAction.js
mfrachet: I took another approach and move the border to the parent. Then, the FieldAction only has to "push" things out . It fixes the negative margin issue and the input under the text issue 👍🏻

HichamELBSI (commented):
-

HichamELBSI (approved):
LGTM

PR #44, opened 2021-04-28
Add a CRA example app

HichamELBSI:
This is just a CRA example to test the DS build and components.


HichamELBSI:
This is just a CRA example to test the DS build and components.

examples/app/README.md
```diff
@@ -0,0 +1,70 @@
+# Getting Started with Create React App
```
soupette, line 1: Maybe we could change the README a bit?

examples/app/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "app",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@buffetjs/core": "^3.3.5",
```
soupette, line 6: Do you plan on removing them later?

examples/app/public/index.html
```diff
@@ -0,0 +1,43 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#000000" />
+    <meta
+      name="description"
+      content="Web site created using create-react-app"
+    />
+    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
```
soupette, line 12: Maybe we could change the favicon to use the strapi one? (Not needed)

examples/app/public/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+  "short_name": "React App",
```
soupette, line 2: Can we change this so it is Strapi branded?

examples/app/src/App.css
```diff
@@ -0,0 +1,37 @@
+.App {
```
soupette, line 1: Not sure this is needed, WDYT?

examples/app/src/App.js
```diff
@@ -0,0 +1,31 @@
+import { useState } from "react";
+import { Checkbox, Box } from "@strapi/design-system";
+import { Checkbox as BuffetCheckbox } from "@buffetjs/core";
+import "./App.css";
+
+function App() {
+  const [val, setValue] = useState();
```
soupette, line 7: It's a nice start, but what do you think later to create a more dynamic form so it is easier to test the other input components?

examples/app/src/index.js
```diff
@@ -0,0 +1,28 @@
...
+import { ThemeProvider, lightTheme } from "@strapi/design-system";
+import { GlobalStyle } from "@buffetjs/styles";
+import "./index.css";
+import Main from "./App";
+import reportWebVitals from "./reportWebVitals";
+
+const App = () => {
+  return (
+    <ThemeProvider theme={lightTheme}>
+      <GlobalStyle />
```
soupette, line 12: Maybe you can create 2 pages, one for buffet one for the new DS so we can easily see the difference between them?

package.json
```diff
@@ -20,6 +20,7 @@
     "specify:pull": "specify pull"
   },
   "devDependencies": {
+    "@craco/craco": "^6.1.2",
```
soupette, line 23: This dependency needs to be in the examples/app so if we choose to remove the exemple app we don't have to remember to clean the repo dependencies.

soupette: Also what do you think about create a new script:

`setup: "yarn && yarn bootstrap && yarn build"` so we have only command that does everything?

yarn-error.log
```diff
@@ -0,0 +1,6813 @@
+Arguments:
```
soupette, line 1: Maybe we should add this file to the .gitignore WDYT?

soupette (commented):
Good job we can clearly see the difference between the 2 component libraries!

examples/app/src/App.css
HichamELBSI: Indeed, it is not :)

examples/app/src/App.js
HichamELBSI: Yes, I will do it when we will have more components. Not sure it is relevant to do it actually with a checkbox and radio only

examples/app/src/index.js
HichamELBSI: Not sure it is relevant to keep buffet in this app. I will remove it.

package.json
HichamELBSI: Why not 😄

yarn-error.log
HichamELBSI: yep

soupette (approved):
LGTM this is awesome! 👏

PR #46, opened 2021-04-30
Force local package install

soupette:

### DIFF
--- a/examples/app/package.json
+++ b/examples/app/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@craco/craco": "^6.1.1",
-    "@strapi/design-system": "../../packages/strapi-design-system",
+    "@strapi/design-system": "^0.0.1",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-scripts": "4.0.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -6,7 +6,7 @@
   "license": "MIT",
   "scripts": {
     "analyze:bundle": "lerna run analyze:bundle --stream",
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "lerna bootstrap --force-local",
```
mfrachet, line 9: force local should not be necessary 🤔 . design-system and icons are working together without that specific need

examples/app/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@craco/craco": "^6.1.1",
-    "@strapi/design-system": "../../packages/strapi-design-system",
+    "@strapi/design-system": "0.0.1",
```
mfrachet, line 7: Also, in order to make the dependency resolve, have you used:

```sh
$ npx lerna add @strapi/design-system --scope=strapi-ds-app-example
```

???

soupette: Nope! I'll update my PR

mfrachet (approved):
LGTM, thank you 😊

PR #47, opened 2021-04-30
Add Row layout

mfrachet:
Add a row layout component.

I'm not a huge fan of this one since it may break responsiveness in the future. But since I don't find any good solutions to handle row base layout, I'm opening this one.

It only contains a minimalist set of CSS rules. If we feel the need to add more, let's add them in another pr 😊

### DIFF
--- a/packages/strapi-design-system/src/Row/Row.js
+++ b/packages/strapi-design-system/src/Row/Row.js
@@ -1,0 +1,24 @@
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+export const Row = styled.div`
+  display: flex;
+  flex-direction: ${({ reverse }) => reverse};
+  justify-content: ${({ justifyContent }) => justifyContent};
+  align-items: ${({ alignItems }) => alignItems};
+  flex-wrap: ${({ wrap }) => wrap};
+`;
+
+Row.defaultProps = {
+  reverse: false,
+  justifyContent: undefined,
+  alignItems: 'center',
+  wrap: undefined,
+};
+
+Row.propTypes = {
+  reverse: PropTypes.bool,
+  justifyContent: PropTypes.string,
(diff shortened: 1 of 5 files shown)


packages/strapi-design-system/src/Row/Row.js
```diff
@@ -0,0 +1,24 @@
...
+
+Row.defaultProps = {
+  reverse: false,
+  justifyContent: undefined,
+  alignItems: 'center',
+  wrap: undefined,
+};
+
+Row.propTypes = {
+  reverse: PropTypes.bool,
```
soupette, line 20: Why this is a boolean?

packages/strapi-design-system/src/Row/Row.js
```diff
@@ -0,0 +1,24 @@
...
+  justifyContent: PropTypes.string,
```
soupette, line 21: Can you order the propTypes alphabetically ?

packages/strapi-design-system/src/Row/Row.js
mfrachet: Cause I failed 😆

packages/strapi-design-system/src/Row/Row.js
mfrachet: come oooon, right, I can do that 😆

soupette: Maybe we could add this https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-prop-types.md ?

HichamELBSI (approved):
LGTM

packages/strapi-design-system/src/Row/Row.js
mfrachet: Yes, why not 😊 . Can you do it?

PR #49, opened 2021-04-30
Adding Pagination component

mfrachet:
This PR brings the Pagination component to the table.

https://design-system-git-add-pagination-strapijs.vercel.app/?path=/story/pagination--base

NOTE: since we can't rely on strong values for the size of the element (32x32) I've tried my best to make it work using padding. Also notice that the size of the content will depend on its value (multiple characters will take more space)

Example:

## API

```jsx
<Pagination activePage={2}>
  <PreviousLink href="/1">Go to previous page</PreviousLink>
  <PageLink number={1} href="/1">
    Page 1
  </PageLink>
  <PageLink number={2} href="/2">
    Page 2
  </PageLink>
  <PageLink number={3} href="/3">
    Page 3
  </PageLink>
  <Dots>There are pages in between</Dots>
  <PageLink number={4} href="/4">
    Page 4
  </PageLink>
  <NextLink href="/3">Go to next page</NextLink>
</Pagination>
```

### DIFF
--- a/packages/strapi-design-system/src/Pagination/Pagination.js
+++ b/packages/strapi-design-system/src/Pagination/Pagination.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import { Row } from '../Row';
+import { PaginationContext } from './PaginationContext';
+
+const PaginationWrapper = styled.nav``;
+const PaginationList = styled(Row)`
+  & > * + * {
+    margin-left: ${({ theme }) => theme.spaces[1]};
+  }
+`;
+
+export const Pagination = ({ children, label, activePage }) => {
+  return (
+    <PaginationContext.Provider value={activePage}>
+      <PaginationWrapper aria-label={label}>
+        <PaginationList as="ul">{children}</PaginationList>
+      </PaginationWrapper>
+    </PaginationContext.Provider>
+  );
+};
+
+Pagination.defaultProps = {
+  label: 'pagination',
+};
+
+Pagination.propTypes = {
+  children: PropTypes.node.isRequired,
+  label: PropTypes.string,
+  activePage: PropTypes.number.isRequired,
+};
--- a/packages/strapi-design-system/src/Pagination/components.js
+++ b/packages/strapi-design-system/src/Pagination/components.js
@@ -1,0 +1,83 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+/** TODO: make sure to use the react component when they are available in strapi-icons */
+import prevLink from './assets/previous.svg';
+import nextLink from './assets/next.svg';
+import { VisuallyHidden } from '../VisuallyHidden';
+import { useActivePage } from './PaginationContext';
+import { Text } from '../Text';
+
+// TODO: make sure to use the Link exposed by the chosen router
+const Wrapper = styled.a`
+  padding: ${({ theme }) => theme.spaces[3]};
+  background: ${({ theme, active }) => (active ? theme.colors.neutral0 : theme.colors.neutral100)};
+  border-radius: ${({ theme }) => theme.borderRadius};
+  color: ${({ theme, active }) => (active ? theme.colors.primary600 : theme.colors.neutral800)};
+  box-shadow: ${({ active }) => (active ? `0px 1px 4px rgba(26, 26, 67, 0.1)` : undefined)};
+  text-decoration: none;
+  display: flex;
+`;
+
+const PaginationText = styled(Text)`
+  line-height: revert;
+`;
+
+export const PreviousLink = ({ children, ...props }) => (
+  <li>
+    <Wrapper {...props}>
+      <VisuallyHidden>{children}</VisuallyHidden>
+      <img src={prevLink} aria-hidden={true} />
+    </Wrapper>
+  </li>
+);
+
+export const NextLink = ({ children, ...props }) => (
+  <li>
+    <Wrapper {...props}>
+      <VisuallyHidden>{children}</VisuallyHidden>
+      <img src={nextLink} aria-hidden={true} />
+    </Wrapper>
+  </li>
+);
+
+export const PageLink = ({ number, children, ...props }) => {
+  const activePage = useActivePage();
+
+  const isActive = activePage === number;
+
+  return (
+    <li>
+      <Wrapper {...props} active={isActive} aria-current={isActive}>
+        <VisuallyHidden>{children}</VisuallyHidden>
+        <PaginationText aria-hidden={true} small={true} highlighted={isActive}>
+          {number}
+        </PaginationText>
+      </Wrapper>
+    </li>
+  );
+};
+
+export const Dots = ({ children, ...props }) => (
+  <li>
+    <Wrapper {...props} as="div">
+      <VisuallyHidden>{children}</VisuallyHidden>
+      <PaginationText aria-hidden={true} small={true}>
+        …
+      </PaginationText>
+    </Wrapper>
+  </li>
+);
+
+PageLink.propTypes = {
+  children: PropTypes.node.isRequired,
+  number: PropTypes.number.isRequired,
+};
+
+const sharedPropTypes = {
+  children: PropTypes.node.isRequired,
+};
+
+Dots.propTypes = sharedPropTypes;
+NextLink.propTypes = sharedPropTypes;
+PreviousLink.propTypes = sharedPropTypes;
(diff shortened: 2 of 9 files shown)


maevalienard:
I'm obsessed with this vercel app combined with the storybook, it's sooo handy.

- The hover state looks really slight but it's my fault, I didn't mention any in the UI doc. Would it be possible to add one? :) 

- The text doesn't look aligned to me. It seems we have more padding on top of the text. Is it something we can improve?

packages/strapi-design-system/src/Pagination/components.js
```diff
@@ -0,0 +1,82 @@
...
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+/** TODO: make sure to use the react component when they are available in strapi-icons */
+import prevLink from './assets/previous.svg';
+import nextLink from './assets/next.svg';
+import { VisuallyHidden } from '../VisuallyHidden';
+import { useActivePage } from './PaginationContext';
+import { Text } from '../Text';
+
+const Wrapper = styled.a`
```
soupette, line 11: Are you sure that using an `a` tag will cause the browser to refresh when navigating from one page to another?

mfrachet: Yes :).

But it might be a problem if we only want partial rendering of the page in a router (like react-router or reach-router)

mfrachet:
cc @maevalienard 



>    The hover state looks really slight but it's my fault, I didn't mention any in the UI doc. Would it be possible to add one? :)

Yes, we can add one 😊 . What's the expected hover state?

>    The text doesn't look aligned to me. It seems we have more padding on top of the text. Is it something we can improve?

Woops, for some reasons, the text was a bit below. I've fixed to center it correctly 😅

HichamELBSI (changes requested):

packages/strapi-design-system/src/Pagination/Pagination.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+import { Row } from '../Row';
+import { PaginationContext } from './PaginationContext';
+
+const PaginationWrapper = styled.nav``;
```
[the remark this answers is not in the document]
mfrachet: I'm not sure to understand what you mean by using the keyboard nav?

However, it's okay to have multiple "nav" elements in a page **if they have a dedicated label** (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role#labeling_landmarks)

😊

packages/strapi-design-system/src/Pagination/components.js
soupette: Do you think we should use a `<button>` instead?

mfrachet: Since the pagination is a nav, I think buttons are not suited (semantically) for navigation

soupette: You might be right, and what would be the difference with using the following?

```
<nav arial-label="pagination navigation">
<ul>
 <li> <button aria-label="Go to page X">X</button></li>
</ul>
</nav>
```

mfrachet: buttons are for actions, not for navigation. https://marcysutton.com/links-vs-buttons-in-modern-web-applications

The problem is the same as using a <p> instead of a <h1>: it's the its purpose

soupette: I think that we should anticipate the integration with our router, either in this PR or when we start migrating the admin panel to the new DS. I know that MUI for instance offers some [router integration](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-prop-types.md) do you think that we should also provide such support on our side?

mfrachet: I think we can start with a simple anchor and add a TODO flag. Then, add a peer dependency to the router we use and smoothly move to the anchor of our routers. That way, we still have time to choose a valid and accessible router without breaking the speed of the DS implem 😊 . Does it sound fair to you?

packages/strapi-design-system/src/Pagination/components.js
```diff
@@ -0,0 +1,83 @@
...
+import styled from 'styled-components';
+import PropTypes from 'prop-types';
+/** TODO: make sure to use the react component when they are available in strapi-icons */
+import prevLink from './assets/previous.svg';
+import nextLink from './assets/next.svg';
+import { VisuallyHidden } from '../VisuallyHidden';
+import { useActivePage } from './PaginationContext';
+import { Text } from '../Text';
+
+// TODO: make sure to use the Link exposed by the chosen router
```
mfrachet, line 11: cc @soupette 😊

maevalienard:
> Yes, we can add one 😊 . What's the expected hover state?
Awesome thank you! @rgesulfo has had a look, he'll let you know :) 

> Woops, for some reasons, the text was a bit below. I've fixed to center it correctly 😅
You're the one, thanks!

soupette (commented):
LGTM!

---

On a side note which is not required in order to merge this PR:

Do we plan on making this component a bit more dynamic?


I am starting to see the need of another package where we would have such components WDYT of the following encapsulation:


```
const DynamicPagination = pages => {
  const [activePage, setActivePage] = usePagination(pages);

   return (
     <Pagination activePage={activePage}>
       {pages.map(page => <PageLink {...page} />}
    </Pagination>
}
```



}

Do you think it could be a good idea to update the example app so we can see the component in use?

HichamELBSI (approved):
LGTM good job !

mfrachet:
Thanks for the comments 😊 

> Do we plan on making this component a bit more dynamic?

They already are :). I know that you would prefer to pass objects instead of using components but passing objects does not make the components more dynamic. It enforces the usage of a specific "shape" (aka the object properties) instead of using the component "shape" (the props).

I don't think we've planned to use this approach, at least, not for now (correct me if I'm wrong @HichamELBSI )

soupette:
> Thanks for the comments 😊
> 
> > Do we plan on making this component a bit more dynamic?
> 
> They already are :). I know that you would prefer to pass objects instead of using components but passing objects does not make the components more dynamic. It enforces the usage of a specific "shape" (aka the object properties) instead of using the component "shape" (the props).
> 
> I don't think we've planned to use this approach, at least, not for now (correct me if I'm wrong @HichamELBSI )
> 
> For instance, what is the real difference between doing:
> 
> ```js
> const Page = ({ pages }) => {
>   const [activePage, setActivePage] = usePagination(pages);
> 
>    return (
>      <Pagination activePage={activePage}>
>        {pages.map(page => <PageLink {...page} />}
>     </Pagination>
> }
> ```
> 
> and
> 
> ```js
> const DynamicPagination = ({ pages }) => {
> const [activePage, setActivePage] = usePagination(pages);
> 
>    return (
>      <Pagination activePage={activePage}>
>        {pages.map(page => <PageLink {...page} />}
>     </Pagination>
>    )
> }
> 
> const Page = ({ pages }) => {
>   const [activePage, setActivePage] = usePagination(pages);
> 
>    return (
>     <DynamicPagination pages={pages} />
>    )
> }
> ```
> 
> except one level of indirection?

The difference will be that we would need to develop hooks/components related to the pagination in the helper plugin in order to avoid redeveloping the same logic between the same kind of pages.

mfrachet:
> The difference will be that we would need to develop hooks/components related to the pagination in the helper plugin in order to avoid redeveloping the same logic between the same kind of pages.

What do you think of creating this specific abstraction directly inside Strapi until people request for it? So that we don't take too much on us for now, and wait for the community to drive us?

soupette:
> > The difference will be that we would need to develop hooks/components related to the pagination in the helper plugin in order to avoid redeveloping the same logic between the same kind of pages.
> 
> What do you think of creating this specific abstraction directly inside Strapi until people request for it? So that we don't take too much on us for now, and wait for the community to drive us?

Sure this is what I meant by this is not required in order to merge this PR.

However what do you think about testing the component in the example app? Do you think it's the good time to do it?

mfrachet:
> Sure this is what I meant by this is not required in order to merge this PR.

Great we're aligned buddiiiiiie 😊 

> However what do you think about testing the component in the example app? Do you think it's the good time to do it?

I was thinking of building a full layout on an example app. But I can definitly integrate examples per components if necessary 😊

soupette:
> > Sure this is what I meant by this is not required in order to merge this PR.
> 
> Great we're aligned buddiiiiiie 😊
> 
> > However what do you think about testing the component in the example app? Do you think it's the good time to do it?
> 
> I was thinking of building a full layout on an example app. But I can definitly integrate examples per components if necessary 😊

It's completely up to you but I do believe it's easier to build some kind of examples incrementally.

rgesulfo:
@mfrachet I finally wrapped up the styling specs for this component (hover state included): https://www.figma.com/file/PICiE8O4NLrHO1lhJftLdE/Design-System-%E2%9C%85?node-id=3400%3A0
What's the best way to proceed since this branch is closed?
Thanks for your help :)

mfrachet:
Thanks @rgesulfo ! We'll simply open a new PR with the modifications 😊

PR #50, opened 2021-04-30
Button component

HichamELBSI:
Add a button component

Example on: https://design-system-git-atom-button-strapijs.vercel.app/?path=/story/button--variants

## API

Existing variants : `['default', 'tertiary', 'secondary', 'danger', 'success', 'success-light', 'danger-light'];`
Existing sizes : `S, L`
```jsx
<Button size="L" onClick={handler} aria-label="Create something">
   Create
</Button>
```

With icon
```jsx
import { Plus, Trash } from '@strapi/icons';
 
<Button aria-label="New row" leftIcon={<Plus />} rightIcon={<Trash />}>
   New row
</Button>
```

## Voiceover

buttonvoice

### DIFF
--- a/packages/strapi-design-system/src/Button/Button.js
+++ b/packages/strapi-design-system/src/Button/Button.js
@@ -6,7 +6,7 @@ import { Box } from '../Box';
 import { getDisabledStyle, getHoverStyle, getActiveStyle, getVariantStyle, getIconPosition } from './utils';
 import { VARIANTS, BUTTON_SIZES } from './constants';
 
-export const ButtonStyle = styled.button`
+export const ButtonWrapper = styled.button`
   display: flex;
   cursor: pointer;
   padding: ${({ theme }) => `${theme.spaces[2]} ${theme.spaces[4]}`};
@@ -31,8 +31,9 @@ export const ButtonStyle = styled.button`
     }
   }
   &[aria-disabled='true'] {
+    pointer-events: none;
     ${getDisabledStyle}
-    &:hover {
+    &:active {
       ${getDisabledStyle}
     }
   }
@@ -47,11 +48,19 @@ export const ButtonStyle = styled.button`
 
 export const Button = ({ variant, leftIcon, rightIcon, disabled, children, size, ...props }) => {
   return (
-    <ButtonStyle aria-pressed={!disabled} aria-disabled={disabled} size={size} variant={variant} {...props}>
-      {leftIcon && <Box paddingRight={2}>{leftIcon}</Box>}
+    <ButtonWrapper aria-disabled={disabled} size={size} variant={variant} {...props}>
+      {leftIcon && (
+        <Box aria-hidden={true} paddingRight={2}>
+          {leftIcon}
+        </Box>
+      )}
       <Text small={size === 'S'}>{children}</Text>
-      {rightIcon && <Box paddingLeft={2}>{rightIcon}</Box>}
-    </ButtonStyle>
+      {rightIcon && (
+        <Box aria-hidden={true} paddingLeft={2}>
+          {rightIcon}
+        </Box>
+      )}
+    </ButtonWrapper>
   );
 };
 
--- a/packages/strapi-design-system/src/Button/Button.js
+++ b/packages/strapi-design-system/src/Button/Button.js
@@ -2,19 +2,33 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { Text } from '../Text';
-import { getDisabledStyle, getHoverStyle, getActiveStyle, getVariantStyle } from './utils';
+import { Box } from '../Box';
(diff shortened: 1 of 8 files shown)


packages/strapi-design-system/src/Button/Button.js
```diff
@@ -1,5 +1,70 @@
...
+  svg {
+    height: ${({ theme }) => theme.spaces[3]};
+  }
+  svg {
+    > g,
+    path {
+      fill: ${({ theme }) => theme.colors.neutral0};
+    }
+  }
+  &:disabled {
```
mfrachet, line 33: I'm not 100% sure how to deal with the disabled state of things when dealing with accessibility. Because it prevents people from even knowing that there is a button (remember the submit form example?). But for today, I don't have strong appreciations :(

packages/strapi-design-system/src/Button/Button.js
```diff
@@ -1,5 +1,70 @@
...
+import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import { Text } from '../Text';
+import { Box } from '../Box';
+import { getDisabledStyle, getHoverStyle, getActiveStyle, getVariantStyle, getBoxPosition } from './utils';
+import { VARIANTS, BUTTON_SIZES } from './constants';
 
-export const Button = styled.button`
-  background: ${(props) => props.theme.colors.primary100};
+export const ButtonStyle = styled.button`
```
mfrachet, line 9:
```suggestion
export const ButtonWrapper = styled.button`
```

Since it's not a "style" per see, but a wrapper around the button 😊

packages/strapi-design-system/src/Button/Button.js
```diff
@@ -1,5 +1,70 @@
...
+  &:active {
+    ${getActiveStyle}
+  }
+  ${getVariantStyle}
 `;
+
+export const Button = ({ variant, leftIcon, rightIcon, children, size, ...props }) => {
+  return (
+    <ButtonStyle size={size} variant={variant} {...props}>
+      {leftIcon && <Box paddingRight={2}>{leftIcon}</Box>}
```
mfrachet, line 51: What if we make this aria-hidden by default? It would allow us not to specify it on the SVG. WYT?

packages/strapi-design-system/src/Button/Button.stories.mdx
```diff
@@ -20,22 +22,54 @@ import { Button } from './Button';
...
+</Canvas>
+
+## With icon
+
+Description...
+
+<Canvas>
+  <Story name="with-icon">
+    <Box paddingBottom={2}>
+      <Button leftIcon={<ContentIcon />}>Content type</Button>
```
mfrachet, line 61: beautiful API, love it👍🏻

packages/strapi-design-system/src/Button/utils.js
```diff
@@ -0,0 +1,124 @@
+import { Text } from '../Text';
+import { LIGHT_VARIANTS } from './constants';
+
+export const getVariantColorName = (variant) => {
```
mfrachet, line 4: Could you please write a test for the utils? Looks like there's a bunch of logic and it may save guts in the future :blush: (or writing snapshots for each scenario from the component entry point?)

packages/strapi-design-system/src/Button/utils.js
```diff
@@ -0,0 +1,124 @@
...
+    default: {
+      return `
+          border: 1px solid ${theme.colors[`${getVariantColorName(variant)}600`]};
+          background: ${theme.colors[`${getVariantColorName(variant)}600`]};
+        `;
+    }
+  }
+};
+
+export const getBoxPosition = ({ size }) => {
```
mfrachet, line 118: What is the aim of this function?

packages/strapi-design-system/src/Loader/__tests__/Loader.spec.js
```diff
@@ -8,21 +8,40 @@ describe('Loader', () => {
   it('snapshots the component', () => {
     const { container } = render(
       <ThemeProvider theme={lightTheme}>
-        <Loader />
+        <Loader>Loading content...</Loader>
```
mfrachet, line 11: Woops, thank you for this 😬

mfrachet (commented):
Very nice job! I got a bunch of comments but it looks very very good 👍🏻

packages/strapi-design-system/src/Button/utils.js
HichamELBSI: This one is used to adapt the position of the icon, I will rename it 👍🏽

soupette (commented):
LGTM!

packages/strapi-design-system/src/Button/utils.js
HichamELBSI: Sure!

mfrachet (approved):
GG, Let's merge thiiiiis 🚀 🙏🏻

PR #53, opened 2021-05-04
Allow to pass rightAction and leftAction for the base field

mfrachet:
In order to create components such as the Search bar that has actions on both sides

A base field with actions on the left and on the right side


soupette (approved):
LGTM

HichamELBSI (approved):
LGTM

PR #54, opened 2021-05-04
Make divider "hr" instead of "div"

mfrachet:
Semantic 🤷🏻‍♂️

### DIFF
--- a/packages/strapi-design-system/src/Divider/Divider.js
+++ b/packages/strapi-design-system/src/Divider/Divider.js
@@ -1,10 +1,15 @@
+import React from 'react';
 import PropTypes from 'prop-types';
 import { Box } from '../Box';
 import styled from 'styled-components';
 
-export const Divider = styled(Box)`
+const DividerWrapper = styled(Box)`
   height: 1px;
+  margin: 0;
+  border: none;
 `;
+
+export const Divider = (props) => <DividerWrapper {...props} as="hr" />;
 
 /**
  * The background props is implicitly passed to the Box component
--- a/packages/strapi-design-system/src/Divider/__tests__/Divider.spec.js
+++ b/packages/strapi-design-system/src/Divider/__tests__/Divider.spec.js
@@ -19,9 +19,11 @@ describe('Divider', () => {
 
       .c1 {
         height: 1px;
+        margin: 0;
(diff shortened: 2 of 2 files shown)


packages/strapi-design-system/src/Divider/Divider.js
```diff
@@ -1,11 +1,16 @@
...
 import styled from 'styled-components';
 
-export const Divider = styled(Box)`
+const DividerWrapper = styled(Box)`
   height: 1px;
+  margin: 0;
+  border: none;
 `;
 
+export const Divider = (props) => <DividerWrapper {...props} as="hr" />;
```
soupette, line 12: WDYT of the following:

```

const Divider = styled(Box)`
  height: 1px; // should we use rem?
  margin: 0;
  border: none;
`;

Divider.defaultProps = {
as: 'hr'
}

```

HichamELBSI: That means we will be able to change the `as` props. Do we want that? IMO we should force the prop

HichamELBSI (commented):
Thanks, good job

packages/strapi-design-system/src/Divider/Divider.js
soupette: Can't we force it using the `.withConfig` ?

mfrachet: What is the easiest to read and to reason about for people?

HichamELBSI: We also can do `styled(() => <Box as="hr" \>)`. So that, we don't need to create a new component.

HichamELBSI (commented):
res

packages/strapi-design-system/src/Divider/Divider.js
mfrachet:
```jsx
() => <Box as="hr" \> 
```

is the new (anonymous) component 😊

HichamELBSI: yes a new comp will be created anyway haha

HichamELBSI (approved):
LGTM

PR #55, opened 2021-05-04
Add Tag component

mfrachet:
Add a Tag component

Example: https://design-system-git-add-tag-strapijs.vercel.app/?path=/story/tag--base

## API

```jsx
 <Tag icon={<HelpIcon aria-hidden={true} />}>Hello world</Tag>
```

## Voice over

Using the Tag component using Voiceover

### DIFF
--- a/packages/strapi-design-system/src/Tag/Tag.js
+++ b/packages/strapi-design-system/src/Tag/Tag.js
@@ -1,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Text } from '../Text';
+import { Box } from '../Box';
+import { Row } from '../Row';
+
+const TagWrapper = styled(Box)`
+  border: 1px solid ${({ theme }) => theme.colors.primary200};
+  border-radius: ${({ theme }) => theme.borderRadius};
+
+  svg path {
+    fill: ${({ theme }) => theme.colors.primary600};
+  }
+`;
+
+const TagText = styled(Text)`
+  border-right: 1px solid ${({ theme }) => theme.colors.primary200};
+  padding-right: ${({ theme }) => theme.spaces[2]};
+`;
+
+export const Tag = ({ children, icon, ...props }) => (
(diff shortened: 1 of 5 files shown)


packages/strapi-design-system/src/Tag/Tag.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Text } from '../Text';
+import { Box } from '../Box';
+import { Row } from '../Row';
+
+const TagWrapper = styled(Box)`
+  border: 1px solid ${({ theme }) => theme.colors.primary200};
+  border-radius: ${({ theme }) => theme.borderRadius};
```
soupette, line 10: I am being picky here but what do you think about creating a utility to retrieve the borderRadius that we could use in all our component? This is almost always the same code

```
const getBorderRadius = ({ theme }) => theme.borderRadius;


const Compo = styled(Box)`
  border-radius: ${getBorderRadius};
`
```

mfrachet: I understand the intent of avoiding duplication. But adding more code in other places reduces visibility to me. I can make such an utility (wyt @HichamELBSI ?) but does it worth it?

HichamELBSI: A helper for one line, not sure how useful it is. Can you explain the advantage of having this, please?

HichamELBSI (approved):
LGTM good job

packages/strapi-design-system/src/Tag/Tag.js
soupette: I let you decide if you think it's worthing it

maevalienard:
Why didn't you chose to pick the cross icon instead of the information one? 😂

mfrachet:
It's not in strapi/icons I think 😓 (that's the only reasons tbh)

maevalienard:
Except that point, LGTM :)

PR #57, opened 2021-05-04
Add icon button

HichamELBSI:
Add a icon button component

Example on: https://design-system-git-icon-button-strapijs.vercel.app/?path=/story/iconbutton--base

## API

```jsx
import { EditIcon } from '@strapi/icons';
 
<IconButton title="Edit">
    <EditIcon />
</IconButton>
```

## Voiceover

iconbuttonvoice

### DIFF
--- a/packages/strapi-design-system/src/BaseButton/BaseButton.js
+++ b/packages/strapi-design-system/src/BaseButton/BaseButton.js
@@ -1,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+export const BaseButtonWrapper = styled.button`
+  display: flex;
+  cursor: pointer;
+  padding: 10px;
+  border-radius: ${({ theme }) => theme.borderRadius};
+  background: ${({ theme }) => theme.colors.neutral0};
+  border: 1px solid ${({ theme }) => theme.colors.neutral200};
+  svg {
+    height: ${({ theme }) => theme.spaces[3]};
+    width: ${({ theme }) => theme.spaces[3]};
+  }
+  svg {
+    > g,
+    path {
+      fill: ${({ theme }) => theme.colors.neutral0};
+    }
+  }
+  &[aria-disabled='true'] {
+    pointer-events: none;
+  }
+`;
+
+export const BaseButton = React.forwardRef(({ disabled, children, ...props }, ref) => {
+  return (
+    <BaseButtonWrapper ref={ref} aria-disabled={disabled} {...props}>
+      {children}
+    </BaseButtonWrapper>
+  );
+});
+
+BaseButton.displayName = 'BaseButton';
+
+BaseButton.defaultProps = {
+  disabled: false,
+};
+BaseButton.propTypes = {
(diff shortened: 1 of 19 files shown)


maevalienard (approved):
LGTM, great job :)

packages/strapi-design-system/src/IconButton/IconButton.js
```diff
@@ -0,0 +1,50 @@
...
+        fill: ${({ theme }) => theme.colors.neutral400};
+      }
+    }
+  }
+`;
+
+export const IconButton = React.forwardRef(({ children, title, ...props }, ref) => {
+  return (
+    <Tooltip content={title}>
+      <IconButtonWrapper aria-label={title} {...props} ref={ref}>
```
mfrachet, line 35: I think aria-label and the tooltip are redundant information. For now, tooltip is used to describe the element, but we may adjust it so that it becomes the label of the element. For instance, in this example, both the label and the description of the element provide the same information. Not sure this is what we exactly want 😊

mfrachet (commented):
Most of the thing looks very good! However, I have one comment that may have a broader impact on the DS. We may have to discuss about it 😊

packages/strapi-design-system/src/IconButton/IconButton.js
HichamELBSI: Not sure to clearly understand your comment sorry Marvin 🙈.

mfrachet: When wanting to provide a meaning to element that are not written with text (e.g icons, button) or to provide more information on an element, we can use aria-label (and aria-labelledby) and aria-describedby.

In your case, both aria-labelledby and aria-describedby provide the exact same information ("Create", "Edit" etc...) I think it would be create to distiguish if we want the tooltip to be the main label (aria-label or aria-labelledby) or if it aims to provide addition information (aria-describedby).

Let me know if it's still not clear 😊

packages/strapi-design-system/src/BaseButton/BaseButton.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+export const BaseButtonWrapper = styled.button`
+  display: flex;
+  cursor: pointer;
+  padding: 10px;
```
mfrachet, line 8: I think we should use tokens here 😛

soupette (commented):
LGTM!

Good job

packages/strapi-design-system/src/Tooltip/Tooltip.js
```diff
@@ -38,6 +41,7 @@ export const Tooltip = ({ children, content, delay, position, ...props }) => {
           visible={visible}
           {...props}
         >
+          <VisuallyHidden id={descriptionId}>{description}</VisuallyHidden>
```
mfrachet, line 44:
```suggestion
          {visible && <VisuallyHidden id={descriptionId}>{description}</VisuallyHidden>}
```

PR #61, opened 2021-05-05
Fix pagination

mfrachet:
Example: https://design-system-git-fix-pagination-strapijs.vercel.app/iframe.html?id=pagination--base&viewMode=story

## Voice Over

Using the pagination in voiceover

### DIFF
--- a/packages/strapi-design-system/src/Pagination/components.js
+++ b/packages/strapi-design-system/src/Pagination/components.js
@@ -14,6 +14,7 @@ const PaginationText = styled(Text)`
 const LinkWrapper = styled.a`
   padding: ${({ theme }) => theme.spaces[3]};
   border-radius: ${({ theme }) => theme.borderRadius};
+  // TODO: make sure to use the one from the theme
   box-shadow: ${({ active }) => (active ? `0px 1px 4px rgba(26, 26, 67, 0.1)` : undefined)};
   text-decoration: none;
   display: flex;
@@ -23,18 +24,8 @@ const PageLinkWrapper = styled(LinkWrapper)`
   color: ${({ theme, active }) => (active ? theme.colors.primary700 : theme.colors.neutral800)};
   background: ${({ theme, active }) => (active ? theme.colors.neutral0 : undefined)};
 
-  &:focus {
-    ${PaginationText} {
-      font-weight: bold;
-    }
-  }
-
   &:hover {
     box-shadow: 0px 1px 4px rgba(26, 26, 67, 0.1);
-
-    ${PaginationText} {
-      font-weight: bold;
-    }
   }
 `;
 
--- a/packages/strapi-design-system/src/Pagination/components.js
+++ b/packages/strapi-design-system/src/Pagination/components.js
@@ -1,71 +1,130 @@
 import React from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
-/** TODO: make sure to use the react component when they are available in strapi-icons */
-import prevLink from './assets/previous.svg';
-import nextLink from './assets/next.svg';
+import { BackFilter, NextFilter } from '@strapi/icons';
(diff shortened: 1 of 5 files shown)


maevalienard:
I don't seem to see the shadow behind the hover boxes?

mfrachet:
> I don't seem to see the shadow behind the hover boxes?

I don't see this in the pagination spec 😬

rgesulfo:
> > I don't seem to see the shadow behind the hover boxes?
> 
> I don't see this in the pagination spec 😬

My bad @mfrachet. Could you add a `box-shadow: 0px 1px 4px rgba(26, 26, 67, 0.1);` to the hover state of the page # and the chevron icons? Thank you so much for your help on this :)

mfrachet:
> My bad @mfrachet. Could you add a box-shadow: 0px 1px 4px rgba(26, 26, 67, 0.1); to the hover state of the page # and the chevron icons? Thank you so much for your help on this :)

We have a shadow token, maybe it could be a good idea to make it part of this one? WYT? cc @rgesulfo 😊

mfrachet:
@rgesulfo @maevalienard the hover state should be fixed. Also notice that the focus state is managed in another pr (adding the blue outline) so it won't be present in this one

packages/strapi-design-system/src/Pagination/components.js
```diff
@@ -1,71 +1,133 @@
...
+  background: ${({ theme, active }) => (active ? theme.colors.neutral0 : undefined)};
+
+  &:focus {
+    ${PaginationText} {
+      font-weight: bold;
+    }
+  }
+
+  &:hover {
+    box-shadow: 0px 1px 4px rgba(26, 26, 67, 0.1);
```
HichamELBSI, line 33: Will this value be in the shadow design token?

packages/strapi-design-system/src/Pagination/components.js
```diff
@@ -1,71 +1,133 @@
...
+
+    ${PaginationText} {
+      font-weight: bold;
+    }
```
HichamELBSI, line 37: This affects the pagination wrapper width. Is it a problem @maevalienard @rgesulfo?

pagination

packages/strapi-design-system/src/Pagination/components.js
```diff
@@ -76,8 +138,12 @@ PageLink.propTypes = {
...
+  href: PropTypes.string.isRequired,
 };
 
-Dots.propTypes = sharedPropTypes;
 NextLink.propTypes = sharedPropTypes;
 PreviousLink.propTypes = sharedPropTypes;
+
+Dots.propTypes = {
+  children: PropTypes.node.isRequired,
+};
```
HichamELBSI, line 149: Clean file, good job!

HichamELBSI (changes requested):
Good job, I love the pagination api ❤️ !

soupette (approved):
LGTM

packages/strapi-design-system/src/Pagination/components.js
maevalienard: It's because we switch from regular text to a bold one. We can remove the boldness if needed

maevalienard:
Looks awesome! I second Hicham about the wrapper size, maybe we could avoid switching from regular text to bold one, so we don't have this issue anymore? wdyt?

mfrachet:
cc @HichamELBSI @maevalienard the font-weight bold issue is fixed (I removed it)

packages/strapi-design-system/src/Pagination/components.js
rgesulfo: @maevalienard to make sure I update the Figma correctly, we remove the bold weight on the hover and select states? Or just the hover? Thanks

maevalienard:
Looks good! Thank youuuuu

PR #65, opened 2021-05-06
Add AssetTag component

mfrachet:
Add AssetTag component

I've taken the freedom to use neutral600 instead of neutral500 for the text color since the color contrast was not sufficient

## API

```jsx
 <AssetTag>Doc</AssetTag>
```

## What it looks like

AssetTag in storybook


HichamELBSI (approved):
LGTM

PR #68, opened 2021-05-07
Add FocusTrap component

mfrachet:
The focus trap component aims to prevent focus from going outside the box. Practical for modals / dialogs. It also bring back the focus to the caller when unmounting

## API

```jsx
<FocusTrap>
  <button>I m done buddie</button>
</FocusTrap>
```

## Example

A storybook example of the FocusTrap component

### DIFF
--- a/packages/strapi-design-system/src/Button/Button.js
+++ b/packages/strapi-design-system/src/Button/Button.js
@@ -54,7 +54,9 @@ export const Button = ({ variant, leftIcon, rightIcon, disabled, children, size,
           {leftIcon}
         </Box>
       )}
-      <Text small={size === 'S'}>{children}</Text>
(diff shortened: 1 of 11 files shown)


packages/strapi-design-system/src/Button/Button.js
```diff
@@ -54,7 +54,9 @@ export const Button = ({ variant, leftIcon, rightIcon, disabled, children, size,
           {leftIcon}
         </Box>
       )}
-      <Text small={size === 'S'}>{children}</Text>
+      <Text small={size === 'S'} as="span">
```
mfrachet, line 57: ℹ️ I've fixed this on the fly since button can't hold a paragraph and Text is a p by default

HichamELBSI:
Awesome!

PR #71, opened 2021-05-10
Add breadcrumbs

HichamELBSI:

# Add breadcrumbs

## Description
This is a draft for the breadcrumbs component. Let's check the API before going forward 😄 

## Demo
Example on : https://design-system-git-breadcrumbs-strapijs.vercel.app/?path=/story/breadcrumbs--base

## API
```jsx
<Breadcrumbs label="Second page">
   <BreadCrumbItem>Home</BreadCrumbItem>
   <BreadCrumbItem>first</BreadCrumbItem>
   <BreadCrumbItem>second</BreadCrumbItem>
</Breadcrumbs>
```

## Screenshot

Screenshot 2021-05-12 at 10 43 23

### DIFF
--- a/packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
@@ -5,6 +5,7 @@ import { After } from '@strapi/icons';
 import { Text } from '../Text';
 import { Box } from '../Box';
 import { Row } from '../Row';
+import { VisuallyHidden } from '../VisuallyHidden';
 
 const CrumbWrapper = styled(Row)`
   svg {
@@ -38,10 +39,16 @@ Crumb.propTypes = {
 };
 const crumbType = PropTypes.shape({ type: PropTypes.oneOf([Crumb]) });
 
-export const Breadcrumbs = ({ children }) => <ol>{children}</ol>;
+export const Breadcrumbs = ({ children, label }) => (
+  <div>
+    <VisuallyHidden>{label}</VisuallyHidden>
+    <ol aria-hidden={true}>{children}</ol>
+  </div>
+);
 
 Breadcrumbs.displayName = 'Breadcrumbs';
 
 Breadcrumbs.propTypes = {
   children: PropTypes.oneOfType([PropTypes.arrayOf(crumbType), crumbType]).isRequired,
+  label: PropTypes.string.isRequired,
 };
--- a/packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
@@ -1,0 +1,66 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Text } from '../Text';
+import { Box } from '../Box';
+import { VisuallyHidden } from '../VisuallyHidden';
+import { useId } from '../helpers/useId';
+
+const BreadCrumbItemWrapper = styled.li`
+  display: inline-flex;
+`;
+
+export const BreadCrumbItem = React.forwardRef(({ children, isLastChild, separator }, ref) => {
+  return (
+    <BreadCrumbItemWrapper ref={ref}>
+      <Text>{children}</Text>
+      {!isLastChild && (
+        <Box paddingLeft={1} paddingRight={1}>
+          {separator}
+        </Box>
+      )}
+    </BreadCrumbItemWrapper>
+  );
+});
+BreadCrumbItem.displayName = 'BreadCrumbItem';
+BreadCrumbItem.defaultProps = {
+  isLastChild: false,
+  separator: '/',
+};
+BreadCrumbItem.propTypes = {
+  children: PropTypes.string.isRequired,
+  isLastChild: PropTypes.bool,
+  separator: PropTypes.string,
+};
+
+export const Breadcrumbs = React.forwardRef(({ label, description, children, ...props }, ref) => {
+  const breadCrumbsId = useId('breadcrumbs');
+  const descriptionId = useId('description');
+  const validChildren = React.Children.toArray(children).filter((child) => React.isValidElement(child));
+  const elements = validChildren.map((child, index) =>
+    React.cloneElement(child, {
+      separator: '/',
+      spacing: 1,
+      isLastChild: validChildren.length === index + 1,
+    }),
+  );
+
+  return (
+    <nav id={breadCrumbsId} aria-describedby={descriptionId} aria-label={label} ref={ref} {...props}>
+      <VisuallyHidden id={descriptionId}>{description}</VisuallyHidden>
+      <ol>{elements}</ol>
+    </nav>
+  );
+});
+
+Breadcrumbs.displayName = 'Breadcrumbs';
+
+Breadcrumbs.defaultProps = {
+  description: undefined,
+  label: undefined,
+};
+Breadcrumbs.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.element, PropTypes.arrayOf(PropTypes.element)]).isRequired,
+  description: PropTypes.string,
+  label: PropTypes.string,
+};
--- a/packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
@@ -1,36 +1,44 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import { After } from '@strapi/icons';
 import { Text } from '../Text';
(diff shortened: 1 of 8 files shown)


mfrachet:
To reduce verbosity, WYT of 

```jsx
<Breadcrumbs>
   <Crumb>Home</Crumb>
   <Crumb>first</Crumb>
   <Crumb>second</Crumb>
</Breadcrumbs>
```

Breadcrumbs means "miette de pains" and crumb means "miette"

mfrachet:
> Do you think Crumb will be understandable for the DS users?

I think it's less verbose and quite understandable, but I let you decide whether it's a good idea or not 😊

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { After } from '@strapi/icons';
+import { Text } from '../Text';
+import { Box } from '../Box';
+import { useId } from '../helpers/useId';
+
+const CrumbWrapper = styled.li`
```
mfrachet, line 9:
```suggestion
const CrumbWrapper = styled(Row)`
```

so that you benefit from the inline prop and the align-items by default and then you can eventually do `<CrumbWrapper as="li"`

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,64 @@
...
+  svg {
+    height: 10px;
+    width: 10px;
+  }
+  svg path {
+    fill: ${({ theme }) => theme.colors.neutral300};
+  }
+`;
+
+export const Crumb = React.forwardRef(({ children, isLastChild }, ref) => {
```
mfrachet, line 21: Are we sure we want to forward the ref? The wrapper is not focusable per se

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,64 @@
...
+  isLastChild: false,
+};
+Crumb.propTypes = {
+  children: PropTypes.string.isRequired,
+  isLastChild: PropTypes.bool,
+};
+
+export const Breadcrumbs = React.forwardRef(({ children, ...props }, ref) => {
+  const breadCrumbsId = useId('breadcrumbs');
+  const validChildren = React.Children.toArray(children).filter((child) => React.isValidElement(child));
```
mfrachet, line 46: What is the aim of the "isValid" call?

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,64 @@
...
+  const elements = validChildren.map((child, index) =>
+    React.cloneElement(child, {
+      isLastChild: validChildren.length === index + 1,
+    }),
+  );
+
+  return (
+    <nav id={breadCrumbsId} ref={ref} {...props}>
```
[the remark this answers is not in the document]
mfrachet: do we need the ref here?

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,64 @@
...
+      <ol>{elements}</ol>
+    </nav>
+  );
+});
+
+Breadcrumbs.displayName = 'Breadcrumbs';
+
+Breadcrumbs.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.element, PropTypes.arrayOf(PropTypes.element)]).isRequired,
```
mfrachet, line 63: isn't PropTypes.node equivalent to what you've written?

mfrachet (changes requested):
Nicely done! I put a couple of comments in the PR about some stuff 😊

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,64 @@
...
+  children: PropTypes.string.isRequired,
+  isLastChild: PropTypes.bool,
+};
+
+export const Breadcrumbs = React.forwardRef(({ children, ...props }, ref) => {
+  const breadCrumbsId = useId('breadcrumbs');
+  const validChildren = React.Children.toArray(children).filter((child) => React.isValidElement(child));
+  const elements = validChildren.map((child, index) =>
+    React.cloneElement(child, {
+      isLastChild: validChildren.length === index + 1,
```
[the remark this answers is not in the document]
HichamELBSI: Nice catch thanks!

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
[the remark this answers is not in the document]
HichamELBSI: Not really

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
HichamELBSI: Just to add validation of the children in order to have valid elements. We can remove it if you think that we don't need it.

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
HichamELBSI: `PropTypes.node` will include `numbers` and `strings`. IMO we should be more strict on that kind of API.

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
soupette: I don't think you need it here, such logic could be interesting if you were cloning the children to add extra props.

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
soupette: Here instead you can make sure that the `child` is `Crumb` or an array of `Crumb`

```
const crumbType = PropTypes.shape({ type: PropTypes.oneOf([Crumb]) });

children: PropTypes.oneOfType([
  PropTypes.arrayOf(crumbType), // array of fields
  crumbType // singular field
  
 ]).isRequired
```

soupette:
> @mfrachet Do you think `Crumb` will be understandable for the DS users? I understand the logic behind it anyway 👍🏽 .
> @soupette any thoughts on this?

I let you decide, I kinda like Crumb

soupette (commented):
LGTM!

Good job

mfrachet:
I'm having one concern concerning navigating with a screen reader. The breadcrumb itself does not mean anything significant in term of experience. I mean, what information does it provide to the user? Right now, it's "just" a list of items with no meaning at all.

What kind of information should this non-navigable breadcrumb provide to the users? (this is not a question specifically targetting @HichamELBSI but the broader team cc @rgesulfo @maevalienard @soupette )

soupette:
> I'm having one concern concerning navigating with a screen reader. The breadcrumb itself does not mean anything significant in term of experience. I mean, what information does it provide to the user? Right now, it's "just" a list of items with no meaning at all.
> 
> What kind of information should this non-navigable breadcrumb provide to the users? (this is not a question specifically targetting @HichamELBSI but the broader team cc @rgesulfo @maevalienard @soupette )

I don't really know what to answer to be honest, I agree that it does not mean anything in term of experience.

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
...
+Crumb.displayName = 'Crumb';
+Crumb.propTypes = {
+  children: PropTypes.string.isRequired,
+};
+const crumbType = PropTypes.shape({ type: PropTypes.oneOf([Crumb]) });
+
+export const Breadcrumbs = ({ children, label }) => (
+  <div aria-label={label}>
+    <ol aria-hidden>{children}</ol>
+  </div>
```
HichamELBSI, line 44: ℹ️

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,54 @@
...
+  children: PropTypes.string.isRequired,
+};
+const crumbType = PropTypes.shape({ type: PropTypes.oneOf([Crumb]) });
+
+export const Breadcrumbs = ({ children, label }) => (
+  <div>
+    <VisuallyHidden>{label}</VisuallyHidden>
+    <ol aria-hidden={true}>{children}</ol>
+  </div>
+);
```
HichamELBSI, line 47: ℹ️

PR #88, opened 2021-05-17
Add text input

mfrachet:
# TextInput

## Description

Higher level component for displaying a textinput on the screen.

NOTE: the Link component is not part of the API since it does not exist yet in the DS document

## Demo
Example on : https://design-system-git-add-text-input-strapijs.vercel.app/?path=/story/textinput--base

## API

```jsx
  <TextInput
          placeholder="This is a content placeholder"
          label="Content"
          name="content"
          hint="Description line"
          error={undefined}
          onChange={(e) => {}}
          value={'content'}
          action={<span>Hello world</span>}
        />
```

### DIFF
--- a/packages/strapi-design-system/src/TextInput/TextInput.js
+++ b/packages/strapi-design-system/src/TextInput/TextInput.js
@@ -6,28 +6,28 @@ import { Row } from '../Row';
 import { Box } from '../Box';
 import styled from 'styled-components';
 
-const TextareaWrapper = styled.div`
+const TextInputWrapper = styled.div`
   & textarea {
     height: ${80 / 16}rem;
   }
 `;
 
 export const TextInput = React.forwardRef(
-  ({ leftAction, rightAction, name, hint, error, label, action, ...props }, ref) => {
+  ({ leftAction, rightAction, name, hint, error, label, labelAction, ...props }, ref) => {
     return (
-      <TextareaWrapper>
+      <TextInputWrapper>
         <Field name={name} hint={hint} error={error}>
           <Stack size={1}>
             <Row cols="auto auto 1fr" gap={1}>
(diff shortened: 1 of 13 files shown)


packages/strapi-design-system/src/TextInput/TextInput.js
```diff
@@ -0,0 +1,53 @@
...
+
+export const TextInput = React.forwardRef(
+  ({ leftAction, rightAction, name, hint, error, label, action, ...props }, ref) => {
+    return (
+      <TextareaWrapper>
+        <Field name={name} hint={hint} error={error}>
+          <Stack size={1}>
+            <Row cols="auto auto 1fr" gap={1}>
+              <FieldLabel>{label}</FieldLabel>
+              {action && <Box paddingLeft={1}>{action}</Box>}
```
HichamELBSI, line 23: The difference between `leftAction, rightAction and action` props is not very clear for me to be honest. In which use cases should we use `leftAction/rightAction` vs `action`.

HichamELBSI (changes requested):

packages/strapi-design-system/src/TextInput/TextInput.js
mfrachet: Oh yeah I understand 😊 . Left/ Right actions are for the input component (like the "eye" for the password, or the "magnifier" for the search input"). The action is for the tooltip information.

Do you have something in mind to adjust this?

HichamELBSI: WDYT about `labelAction`

mfrachet: Let's try with that 👍🏻

HichamELBSI (approved):
LGTM

PR #91, opened 2021-05-18
Add Table component for a11y

mfrachet:
# Add Table component

## Description

Add a keyboard navigable Table component

- When pressing Arrow Right, it moves to the cell on the right
- When pressing Arrow Left, it moves to the cell on the left
- When pressing Arrow Up, it moves to the cell on the top
- When pressing Arrow Down, it moves to the cell on the bottom
- When pressing Home, it moves to the first cell of the column
- When pressing End, it moves to the last cell of the column
- When pressing End, it moves to the last cell of the column
- When pressing Control+Home, it moves to the first cell of the first row
- When pressing Control+End, it moves to the last cell of the last row
- When pressing PageUp, it moves X step to the top (passed in props)
- When pressing PageDown, it moves X step to the bottom (passed in props)

## Demo
Example on : https://design-system-git-add-table-strapijs.vercel.app/?path=/story/table--base

## API

```jsx
         <Table colCount={COL_COUNT} rowCount={ROW_COUNT}>
            <Thead>
              <Tr>
                <Th>One</Th>
                <Th>Two</Th>
                <Th>Three</Th>
                <Th>Four</Th>
                <Th>Five</Th>
              </Tr>
            </Thead>
            <Tbody>
              <Tr>
                <Td>2/1</Td>
                <Td>2/2</Td>
                <Td>2/3</Td>
                <Td>2/4</Td>
                <Td>2/5</Td>
              </Tr>
              <Tr>
                <Td>3/1</Td>
                <Td>3/2</Td>
                <Td>3/3</Td>
                <Td>3/4</Td>
                <Td>3/5</Td>
              </Tr>
            </Tbody>
          </Table>
```


packages/strapi-design-system/src/Table/Cell.js
```diff
@@ -0,0 +1,47 @@
...
+  const cellRef = useRef(null);
+
+  useEffect(() => {
+    if (mountedRef.current && isFocused) {
+      cellRef.current.focus();
+    }
+  }, [isFocused]);
+
+  useEffect(() => {
+    mountedRef.current = true;
```
soupette, line 15: Why don't we need the cleanup here?

mfrachet: not sure why you'd like to cleanup?

packages/strapi-design-system/src/Table/Table.js
```diff
@@ -0,0 +1,107 @@
...
+   */
+  const [rowIndex, setRowIndex] = useState(0);
+  const [colIndex, setColIndex] = useState(0);
+
+  useEffect(() => {
+    const handleKeyDown = (e) => {
+      switch (e.key) {
+        case KeyboardKeys.ARROW_RIGHT: {
+          e.preventDefault();
+          setColIndex((s) => (s < colCount - 1 ? s + 1 : s));
```
soupette, line 19: Can you find another naming for `s` ?

packages/strapi-design-system/src/Table/Content.js
```diff
@@ -0,0 +1,38 @@
...
+   */
+  const childrenClone = Children.toArray(children).map((child, index) =>
+    cloneElement(child, { focusedColIndex: rowIndex - 1 === index ? colIndex : undefined, 'aria-rowindex': index + 2 }),
+  );
+
+  return <tbody {...props}>{childrenClone}</tbody>;
+};
+
+Thead.propTypes = {
+  children: PropTypes.node.isRequired,
```
soupette, line 33: What do you think about making sure that the children is `Tr` ?

mfrachet: For an obscure reason, it works for tbody, but not for thead 🤔

packages/strapi-design-system/src/Table/Cell.js
soupette: Settings the `mountedRef.current`  to false

packages/strapi-design-system/src/Table/Content.js
soupette: Weird...
Also, do you think we should split theses components into 2 different files?

packages/strapi-design-system/src/Table/Cell.js
mfrachet: This still not answers "why" but how 😛

soupette: Making sure the effect above is not ran when the components unmounts. Usually I take the other approach:


```
const mountedRef = useRef(true);

useEffect(() => {
  // ...
}, [isFocused)

useEffect(() => {
 return () => (mountedRef.current = false)
}, [])
```

mfrachet: The goal of this is to make sure the .focus is not called on mount. Verifying the value when unmounting does not seem necessary to me

packages/strapi-design-system/src/Table/Content.js
mfrachet: ✅

soupette (commented):
I'll let @HichamELBSI review this PR but it is good for me!

Good job!

HichamELBSI (approved):
Good job! I really like that component 😄

PR #94, opened 2021-05-20
Improve build icons

mfrachet:
# Improving build icons

## Description

- Moving back examples to the root of the folder (cc what @HichamELBSI has done in the past 😅 )
- Making sure we can import `@strapi/icons/Icon` to reduce the bundle size
- Adjusting jest and webpack config because of the mono repo 💣 

The idea is to deploy `@strapi/icons/dist` with the package.json included so that every single components is at the root and a sibling of the "main": "index.js", a la material-ui

### Before

Bundle size owning strapi-icons in every single component

### After

Bundle size without strapi-icons in every single component


example/package.json
```diff
@@ -5,8 +5,8 @@
   "dependencies": {
     "@craco/craco": "^6.1.1",
     "@reach/router": "^1.3.4",
-    "@strapi/design-system": "^0.0.1",
-    "@strapi/icons": "^0.0.1",
+    "@strapi/design-system": "file:../packages/strapi-design-system",
+    "@strapi/icons": "file:../packages/strapi-icons/dist",
```
mfrachet, line 9: as Hicham suggested some times ago, this is necessary to make sure we use the library the way it should be used.

package.json
```diff
@@ -8,7 +8,7 @@
     "analyze:bundle": "lerna run analyze:bundle --stream",
     "bootstrap": "lerna bootstrap",
     "build-storybook": "lerna run build-storybook --stream",
-    "build": "lerna run build --stream",
+    "build": "lerna run build --stream --scope=@strapi/icons && lerna run build --stream --scope=@strapi/design-system",
```
mfrachet, line 11: ℹ️ order matters

packages/strapi-design-system/.storybook/main.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
     // Fix that allows to fix storybook issues with core-js
     // https://github.com/storybookjs/storybook/issues/11255
     config.resolve.alias['core-js'] = path.dirname(require.resolve('core-js'));
+    config.resolve.alias['@strapi/icons'] = path.dirname(require.resolve('../../strapi-icons/dist'));
```
soupette, line 10: I don't think you need the `/dist` if the `main` key of the strapi-icons package is already the dist one.

packages/strapi-design-system/src/Alert/__tests__/Alert.spec.js
```diff
@@ -5,12 +5,29 @@ import { Alert } from '../Alert';
 import { ThemeProvider } from '../../ThemeProvider';
 import { lightTheme } from '../../themes';
 
-jest.mock('@strapi/icons', () => ({
-  AlertInfoIcon: () => <span>AlertInfoIcon</span>,
-  AlertSucessIcon: () => <span>AlertSucessIcon</span>,
-  AlertWarningIcon: () => <span>AlertWarningIcon</span>,
-  CloseAlertIcon: () => <span>CloseAlertIcon</span>,
-  ExternalLink: () => <span>ExternalLink</span>,
+jest.mock('@strapi/icons/AlertInfoIcon', () => ({
```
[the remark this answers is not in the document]
soupette: Why do we need to mock it?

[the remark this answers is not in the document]
mfrachet: Because of peer dependencies. react is not defined in the strapi icons since the host application should resolve react by itself. Right now, I don't know why it's a problem for jest to resolve react as a peer dependency, but that's a problem I'll check later since we used to mock icons anyway before this PR

HichamELBSI (approved):
LGTM good job!!

PR #101, opened 2021-05-25
Fix textarea

mfrachet:
# Fix textarea

## Description

Fix textarea comment from @maevalienard on https://github.com/strapi/design-system-experiments/pull/84#issuecomment-847101591

>    Will the placeholder style be implemented later? cc @mfrachet

It's supposed to be done here

>    Does the description/error line changes?

Not sure to understand?

>    I think there's a bug with the area opener pointer, when I moved it (probably on the left), it stayed there

Yep, good catch, fixed in this one 👍🏻 

## Demo

### DIFF
--- a/packages/strapi-design-system/src/Textarea/TextareaInput.js
+++ b/packages/strapi-design-system/src/Textarea/TextareaInput.js
@@ -1,6 +1,7 @@
 import React, { forwardRef } from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
+import { useField } from '../Field';
 
 const Textarea = styled.textarea`
   display: block;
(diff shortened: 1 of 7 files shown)


maevalienard:
Capture d’écran 2021-05-25 à 09 01 40
Seems like we have a double focus box when the input is active. 

Also, not sure the text style is the right one (for the placeholder as much as for the actual text -> should be SmallTextButton style) :)

mfrachet:
Can you hard refresh please? cc @maevalienard

packages/strapi-design-system/src/Textarea/TextareaInput.js
```diff
@@ -0,0 +1,72 @@
...
+export const TextareaInput = forwardRef(({ disabled, ...props }, ref) => {
+  const { id, error, hint, name } = useField();
+
+  let ariaDescription;
+
+  if (error) {
+    ariaDescription = `field-error-${id}`;
+  } else if (hint) {
+    ariaDescription = `field-hint-${id}`;
+  }
```
HichamELBSI, line 45: What's the point of those aria desc `field-error-` and `field-hint-` ?

HichamELBSI (changes requested):
Good job !

packages/strapi-design-system/src/Textarea/TextareaInput.js
mfrachet: just in case the id is used by another field. "email" might be on used in multiple places, maybe 😊

maevalienard (approved):
LGTM! Thanks for all the changes!

PR #102, opened 2021-05-25
Make the focus trap aware of dynamic content

mfrachet:
# Fix focus trap bug

## Description

Before this PR, when dynamically showing content in the three (using a boolean state for instance), the focus tree was never updated because the listener was registered only at "mount time".

## Demo

## Record

Focus trap rotating correctly


soupette (commented):
LGTM

HichamELBSI (approved):
Awesome! Thanks @mfrachet

PR #103, opened 2021-05-25
Improve table focus

mfrachet:
# Improve Table focus

## Description

This PR improves the Td focus when they own focusable elements such as anchor and button. instead of focusing the cell and then having to tab to focus the anchor/button, the anchor/button is directly focused.

### DIFF
--- a/packages/strapi-design-system/src/Table/Table.js
+++ b/packages/strapi-design-system/src/Table/Table.js
@@ -2,9 +2,11 @@ import React, { useEffect, useRef, useState } from 'react';
 import PropTypes from 'prop-types';
 import { TableContext } from './TableContext';
 import { KeyboardKeys } from '../helpers/keyboardKeys';
+import { focusFocusable } from './focusFocusable';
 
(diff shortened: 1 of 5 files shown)


packages/strapi-design-system/src/Table/Table.js
```diff
@@ -2,96 +2,103 @@ import React, { useEffect, useRef, useState } from 'react';
...
-      switch (e.key) {
-        case KeyboardKeys.ARROW_RIGHT: {
-          e.preventDefault();
-          setColIndex((prevColIndex) => (prevColIndex < colCount - 1 ? prevColIndex + 1 : prevColIndex));
+    if (mountedRef.current) {
+      focusFocusable(tableRef.current);
+    }
+
+    if (!mountedRef.current) {
+      mountedRef.current = true;
```
soupette, line 22: Don't you need the cleanup?

`return () => (mountedRef.current = false)`

mfrachet: Do you think I need it? 🤔

HichamELBSI (approved):
LGTM good job!

PR #106, opened 2021-05-26
Add Tabs

mfrachet:
# Add Tabs

## Description

Add a tabs panel:

- when one of the tab is focused, pressing arrow right moves to the next one (or the first one if the focused item is the last)
- when one of the tab is focused, pressing arrow left moves to the previous one (or the last one if the focused item is the first)
- when one of the tab is focused, pressing home moves to the first tab
- when one of the tab is focused, pressing last moves to the last tab
- pressing tab outside the list sends the focus to the lastly focused tab (or the first one if it has never been selected)
- pressing tab inside the list when a tab is already focused sends the focus to the according tab panel for reading purpose

## Demo

⚠️ make sure to hard refresh

Example on : https://design-system-git-add-tabs-strapijs.vercel.app/?path=/story/tabs--base

## API

```jsx
<TabGroup label="Some stuff for the label" id="tabs">
  <Tabs>
    <Tab>First</Tab>
    <Tab>Second</Tab>
    <Tab>Third</Tab>
  </Tabs>
  <TabPanels>
    <TabPanel>First panel</TabPanel>
    <TabPanel>Second panel</TabPanel>
    <TabPanel>Third panel</TabPanel>
  </TabPanels>
</TabGroup>
```

## Voiceover

The tabs component used with VoiceOver

### DIFF
--- a/packages/strapi-design-system/src/Tabs/TabGroup.js
+++ b/packages/strapi-design-system/src/Tabs/TabGroup.js
@@ -1,0 +1,24 @@
+import React, { useRef, useState } from 'react';
+import PropTypes from 'prop-types';
+import { TabsContext } from './TabsContext';
+import { genId } from '../helpers/genId';
+
+export const TabGroup = ({ id, label, ...props }) => {
+  const tabsId = useRef(id || genId());
+  const [selectedTabIndex, setSelectedTabIndex] = useState(0);
+
+  return (
+    <TabsContext.Provider value={{ id: tabsId.current, selectedTabIndex, selectTabIndex: setSelectedTabIndex, label }}>
+      <div {...props} />
+    </TabsContext.Provider>
+  );
+};
+
+TabGroup.defaultProps = {
+  id: undefined,
+};
+
+TabGroup.propTypes = {
+  id: PropTypes.string,
+  label: PropTypes.string.isRequired,
+};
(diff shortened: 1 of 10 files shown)


maevalienard:
😍 😍 😍 Looks good! Great job!

We just miss the 4px border-radius at these two places 
Capture d’écran 2021-05-26 à 09 21 12

HichamELBSI:
I wonder if there are any standards for keyboard navigation for that kind of pattern. How is a user with SR supposed to know which keys are available for navigation?

maevalienard:
It's perfect now, well done and thanks for the changes 🚀

maevalienard (approved):

HichamELBSI:
> > I wonder if there are any standards for keyboard navigation for that kind of pattern. How is a user with SR supposed to know which keys are available for navigation?
> 
> [The WAI-aria spec provides some common patterns](https://www.w3.org/TR/wai-aria-practices-1.1/#aria_ex) and ways to solve the according accessibility issues :). Also, there are a lot of examples across the web on how one thing could improve people lives.
> 
> I'm also part of the ["A11y" slack group](web-a11y.slack.com) where you can ask questions when you're in doubt.
> 
> For example, to create this component, I've taken what looked useful in https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel and I've checked https://reach.tech/tabs 😊
> 
> One thing to keep in mind is that it's about user experience. If we think something improves user experience (in a not too much expensive way) it should hopefully be implemented

Thanks Marvin :)

HichamELBSI (approved):
I love the API thanks Marvin and Good job!

PR #107, opened 2021-05-26
Allows popover to receive a spacing for positioning from the top

mfrachet:
## Description

Allows popover to receive a spacing for positioning from the top

### DIFF
--- a/packages/strapi-design-system/src/Popover/Popover.js
+++ b/packages/strapi-design-system/src/Popover/Popover.js
@@ -22,7 +22,7 @@ const PopoverWrapper = styled(Box)`
   position: absolute;
   border: 1px solid ${({ theme }) => theme.colors.neutral150};
   background: ${({ theme }) => theme.colors.neutral0};
-  margin-top: ${({ theme, spacing }) => theme.spaces[spacing]};
+  margin-top: ${({ theme, spacingTop }) => theme.spaces[spacingTop]};
 `;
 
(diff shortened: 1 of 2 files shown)


packages/strapi-design-system/src/Popover/Popover.js
```diff
@@ -44,7 +45,7 @@ const PopoverScrollable = styled(Box)`
   }
 `;
 
-export const Popover = ({ source, visible, children, ...props }) => {
+export const Popover = ({ source, visible, children, spacing, ...props }) => {
```
HichamELBSI, line 48: That spacing prop can be a bit confusing. As a developer, I expect to have spacing on every side, not just the top.

HichamELBSI (changes requested):

packages/strapi-design-system/src/Popover/Popover.js
mfrachet: Sounds good 😊 what could be another valid name?

HichamELBSI: I don't really know maybe something like `spacingTop`, `verticalSpacing` or `marginTop`.

mfrachet: I've taken spacingTop

HichamELBSI (approved):
LGTM

PR #112, opened 2021-05-28
Main nav round 2

mfrachet:
Add a NavUser component and the NavCondense button

Main navigation with the NavUser and NavCondense component

### DIFF
--- a/packages/strapi-design-system/src/Avatar/Avatar.js
+++ b/packages/strapi-design-system/src/Avatar/Avatar.js
@@ -1,0 +1,8 @@
+import styled from 'styled-components';
+
+export const Avatar = styled.img`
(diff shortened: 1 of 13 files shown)


packages/strapi-design-system/src/Avatar/Avatar.js
```diff
@@ -0,0 +1,8 @@
+import styled from 'styled-components';
+
+export const Avatar = styled.img`
```
soupette, line 3: Should we add a PropTypes to make sure the alt is always defined?

mfrachet: Yep, sure

HichamELBSI (approved):
LGTM

PR #116, opened 2021-06-01
Fix text button

mfrachet:
## Description

The Icons for disabled were not taking their colors because the "disabled" prop has not been spread to the wrapper


HichamELBSI (approved):
LGTM

PR #119, opened 2021-06-04
Simple menu

HichamELBSI:
# Simple menu

## Description
Add the simple menu dropdown.

## Demo
Example on : https://design-system-git-simple-menu-strapijs.vercel.app/?path=/story/design-system-atoms-simplemenu--base

## API

```jsx
<SimpleMenu label="January">
   <MenuItem id="menuItem-January" onClick={() => setValue('January')}>
     January
   </MenuItem>
   <MenuItem id="menuItem-February" onClick={() => setValue('February')}>
     February
   </MenuItem>
   <MenuItem id="menuItem-Strapi" href="https://strapi.io">
     Strapi website
   </MenuItem>
 </SimpleMenu>
```

## Voiceover

simplemenu

### DIFF
--- a/packages/strapi-design-system/src/Popover/Popover.js
+++ b/packages/strapi-design-system/src/Popover/Popover.js
@@ -29,11 +29,6 @@ const PopoverScrollable = styled(Box)`
   max-height: ${3 * 5}rem;
   overflow-y: scroll;
 
-  &::-webkit-scrollbar {
-    position: absolute;
-    width: ${({ theme }) => theme.spaces[1]};
-  }
-
   &::-webkit-scrollbar-track {
     background: ${({ theme }) => theme.colors.neutral0};
   }
--- a/packages/strapi-design-system/src/Popover/Popover.js
+++ b/packages/strapi-design-system/src/Popover/Popover.js
@@ -30,6 +30,7 @@ const PopoverScrollable = styled(Box)`
   overflow-y: scroll;
 
   &::-webkit-scrollbar {
+    position: absolute;
     width: ${({ theme }) => theme.spaces[1]};
   }
 
@@ -55,9 +56,7 @@ export const Popover = ({ source, children, spacingTop, ...props }) => {
   return (
     <Portal>
       <PopoverWrapper style={style} hasRadius background="neutral0" padding={1} spacingTop={spacingTop}>
-        <PopoverScrollable paddingRight={1} {...props}>
-          {children}
-        </PopoverScrollable>
+        <PopoverScrollable {...props}>{children}</PopoverScrollable>
       </PopoverWrapper>
     </Portal>
   );
--- a/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
+++ b/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
@@ -1,0 +1,51 @@
+import React, { useRef, useState } from 'react';
+import styled from 'styled-components';
+import FilterDropdown from '@strapi/icons/FilterDropdown';
+import { Text } from '../Text';
+import { Row } from '../Row';
+import { Box } from '../Box';
+import { Popover } from '../Popover';
+
+const MenuButton = styled.button`
+  border: none;
+  background: transparent;
+  display: flex;
+  align-items: center;
+  font-size: ${12 / 16}rem;
+  svg {
+    height: 4px;
+  }
+`;
+
+export const SimpleMenu = () => {
+  const menuRef = useRef();
+  const [visible, setVisible] = useState(false);
+
+  return (
+    <div>
+      <MenuButton onClick={() => setVisible((s) => !s)} ref={menuRef}>
+        <Box paddingRight={1}>
+          <Text>January</Text>
+        </Box>
+        <FilterDropdown />
+      </MenuButton>
+      {visible && (
+        <Popover source={menuRef} spacingTop={1}>
+          <ul>
+            {Array(15)
+              .fill(null)
+              .map((_, index) => (
+                <Box key={index} padding={2}>
+                  Element {index}
+                </Box>
+              ))}
+          </ul>
+        </Popover>
+      )}
+    </div>
+  );
+};
+
+SimpleMenu.displayName = 'SimpleMenu';
+
+SimpleMenu.propTypes = {};
--- a/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
+++ b/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
@@ -1,4 +1,4 @@
-import React, { useRef, useState, Children, cloneElement } from 'react';
+import React, { useRef, useState, Children, cloneElement, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import FilterDropdown from '@strapi/icons/FilterDropdown';
@@ -7,6 +7,8 @@ import { Box } from '../Box';
 import { Row } from '../Row';
 import { Popover } from '../Popover';
 import { getOptionStyle } from './utils';
+import { useId } from '../helpers/useId';
+import { KeyboardKeys } from '../helpers/keyboardKeys';
 
 const OptionButton = styled.button`
   border: none;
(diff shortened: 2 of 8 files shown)


maevalienard:
Capture d’écran 2021-06-04 à 16 08 59
From what I see, you have an extra padding on the right of the hover (see picture)

Also, the icon should be neutral500 :D 

Thank you!

soupette (commented):
LGTM!


Good job

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,111 @@
...
+  svg {
+    height: 4px;
+  }
+`;
+
+export const MenuItem = ({ children, onClick, href, ...props }) => {
+  if (href) {
+    return (
+      <Row as="li" justyfiContent="center">
+        <OptionLink href={href} {...props}>
```
mfrachet, line 37: I think we could improve the experience by making this one a `role="menuitem"`

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,111 @@
...
+    <div>
+      <MenuButton {...props} onClick={() => setVisible((s) => !s)} ref={menuButtonRef}>
+        <Box paddingRight={1}>
+          <TextButton>{label}</TextButton>
+        </Box>
+        <FilterDropdown aria-hidden />
+      </MenuButton>
+      {visible && (
+        <Popover tabindex="0" onBlur={handleBlur} source={menuButtonRef} spacingTop={1}>
+          <Box as="ul" padding={1}>
```
mfrachet, line 95: I think we could improve the UX by adding a `role="menu"` to this element

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,111 @@
...
+      onClick: () => {
+        setVisible(false);
+        child.props.onClick();
+      },
+    }),
+  );
+
+  return (
+    <div>
+      <MenuButton {...props} onClick={() => setVisible((s) => !s)} ref={menuButtonRef}>
```
mfrachet, line 87: You can also add `aria-haspopup={true}` to inform the user that this button aims to open a popup 😊 

Also, you can use `aria-expanded={visible}` so that the user knows in which state is the menu

mfrachet: And you can also add `aria-controls={menuId}` so that assistive technologies can create a relation between the element that aims to open the popup and the popup itself

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,111 @@
...
+  background: transparent;
+  display: flex;
+  align-items: center;
+  font-size: ${12 / 16}rem;
+  svg {
+    height: 4px;
+  }
+`;
+
+export const MenuItem = ({ children, onClick, href, ...props }) => {
```
mfrachet, line 33: I think that the MenuItems should have a `tabIndex={-1}`. 

**Why?**

`tabIndex={-1}` aims to make an element focusable programmatically. While it's not possible to target them using the `tab` key, it's still possible to focus them with `element.focus()`.

**Why do we want that?**

Generally, when providing an experience to people using the keyboard, the `tab` key aims to get to the next "widget" (or component, or "compound group of elements"). In the actual state of the component, using tab gets you to the next children of the menu (e.g the MenuItem). I do think we can have a better experience by preventing tabs from getting to the next menuitem, but instead to get the focus to the next focusable widget.

For instance, imagine that we have 20 elements in the menu list: do we want the user to tab 21 times to get out of the box and focus on the next element? 😊 

**How to access the next menuitem then?**

This is where keys enter the game: using arrow up, arrow down and eventually addition keys (home + arrow left etc...)

Let me know If this is not clear enough 😊

packages/strapi-design-system/src/SimpleMenu/__tests__/SimpleMenu.spec.js
```diff
@@ -0,0 +1,126 @@
...
+
+    const button = await waitFor(() => screen.getByText('January'));
+    fireEvent.click(button);
+
+    await waitFor(() => {
+      expect(screen.getByText('February')).toBeInTheDocument();
+    });
+  });
+
+  it('display the menu and click on a menu item', async () => {
```
mfrachet, line 105: Great test 😊 🚀

mfrachet (changes requested):
The component looks really nice 👍🏻 , nice job! I'm having a bunch of issue concerning accessibility that we can address together if you want. Feel free to let me know and challenge the suggestion if you think they are worthless 😊

packages/strapi-design-system/src/Popover/Popover.js
```diff
@@ -30,6 +30,7 @@ const PopoverScrollable = styled(Box)`
   overflow-y: scroll;
 
   &::-webkit-scrollbar {
+    position: absolute;
```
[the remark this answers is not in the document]
HichamELBSI: just forgot it, the position absolute didn't work at all 😄

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
HichamELBSI: It is super clear thank you. I was wondering that also but I wasn't sure about it.

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,168 @@
...
+  const [focusedItemIndex, setFocusItem] = useState(0);
+  const childrenArray = Children.toArray(children);
+
+  const handleWrapperKeyDown = (e) => {
+    if (visible) {
+      if (e.key === KeyboardKeys.ESCAPE) {
+        setVisible(false);
+        setFocusItem(0);
+      }
+      if (e.key === KeyboardKeys.TAB) return;
```
mfrachet, line 95: I'm not sure to understand why this verification is necessary? 😅

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,168 @@
...
+
+      if (e.key === KeyboardKeys.DOWN && focusedItemIndex < childrenArray.length - 1) {
+        setFocusItem((prev) => prev + 1);
```
mfrachet, line 98: WYT of looping over the menu item when the last element is focused?

Like if you're positioned on the last element and you press arrow down, you get to the first item. Same goes for the last item 😊

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,168 @@
...
+      }
+
+      if (e.key === KeyboardKeys.UP && focusedItemIndex > 0) {
+        setFocusItem((prev) => prev - 1);
+      }
+    }
+  };
+
+  const handleKeyDown = (e) => {
```
mfrachet, line 107: I'm not sure to understand why it's necessary?

By default, a "click" event on a "button" element can be triggered with Space or Enter without the need to add an extra listener on the keys. But maybe I'm missing something here 😊

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,168 @@
...
+    }),
+  );
+
+  return (
+    <div onKeyDown={handleWrapperKeyDown}>
+      <MenuButton
+        aria-haspopup
+        aria-expanded={visible}
+        aria-controls={menuId}
+        {...props}
```
mfrachet, line 140: Could you please move the spread at the end of the compo? Just so that everything can be overriden if necessary

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,168 @@
...
+  useEffect(() => {
+    if (isFocused && menuItemRef.current) {
+      menuItemRef.current.focus();
+    }
+  }, [isFocused]);
+
+  const menuItemProps = {
+    tabIndex: -1,
+    ref: menuItemRef,
+    role: 'menuItem',
```
mfrachet, line 47: `menuItem` should be `menuitem` I think

mfrachet (changes requested):
This is a great new iteration, thanks for the job made 👍🏻 

Also, I think that could be great to add the possibility to re-focus the select item when opening the menu. Like if the select element is "February", you could put the focus to that element directly when opening the menu. You could probably achieve this using a roving index (https://www.stefanjudis.com/today-i-learned/roving-tabindex/)

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
HichamELBSI: Completly useless. Just forgot it. Nice catch!

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
HichamELBSI: Nice catch, like that!

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
HichamELBSI: My bad!

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
HichamELBSI: Well, this is the tricky part. I wasn't able to stop the propagation of the onBlur event on the popover so I replace the `click` with a `mouse down` event to open the popover. Space or Enter doesn't trigger the mouse down event. I don't really like that approach TBH.

click

packages/strapi-design-system/src/SimpleMenu/__tests__/SimpleMenu.spec.js
```diff
@@ -0,0 +1,133 @@
...
+        <button
+          aria-controls="simpleMenu-1"
+          aria-expanded="false"
+          aria-haspopup="true"
+          class="c0"
+        >
+          <div
+            class="c1"
+          >
+            <p
```
mfrachet, line 67: This is still a `<p>` for some reasons where it should be a span 😊

mfrachet (changes requested):
So cloooooose! Congratz Hicham, I still have a few comments, but the overall looks pretty nice 👍🏻

maevalienard:
Don't forget to change the carret color 🙈 

Also, when the component is open and you want to click outside of it to close it, it doesn't close :(

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -0,0 +1,169 @@
...
+      }
+
+      if (e.key === KeyboardKeys.UP) {
+        setFocusItem((prev) => (prev === 0 ? childrenArray.length - 1 : prev - 1));
+      }
+    }
+  };
+
+  const handleKeyDown = (e) => {
+    if (e.key === KeyboardKeys.ENTER) {
```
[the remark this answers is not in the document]
HichamELBSI: Sorry @mfrachet, I don't really understand, do you want me to change something here except the `Space`?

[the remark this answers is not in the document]
mfrachet: The menu should open when pressing "Enter" or "Space" 😊

mfrachet (changes requested):
When pressing "space" on the menu button, the menu opens and hides directly 😅 . Also, could you please add tests for opening the button using "Space" and "ArrowDown" please?

HichamELBSI:
> When pressing "space" on the menu button, the menu opens and hides directly 😅 . Also, could you please add tests for opening the button using "Space" and "ArrowDown" please?

Updated ! 😄

PR #121, opened 2021-06-09
Adjust popover for fullwidth

mfrachet:
The popover now receives a `fullWidth` boolean props that allows to match the size of the source element


HichamELBSI (approved):
LGTM

PR #122, opened 2021-06-14
Renaming Table to RawTable

mfrachet:
Since we now have a `Table` component in the design, I'm feeling that we should rename the existing table component to "RawTable" since it doesn't own any styles and only helps for keyboard nav 😊


HichamELBSI:
I think we should also rename BaseButton, BaseRadio, and BaseCheckbox to RawButton, RawRadio and RawCheckbox to keep consistency.
Or maybe just rename RawTable to BaseTable. 
I prefer Raw TBH 😄 .

mfrachet:
Yeah, we'll probably have to 😅

soupette (approved):
LGTM

PR #124, opened 2021-06-15
Rework select

mfrachet:
Reworking the Select component using the `aria-activedescendant` property  https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html

It has a more appropriate semantic, and it will be easier to manager the multi select thing

API stays the same

https://design-system-git-rework-select-strapijs.vercel.app/?path=/story/select--base


soupette (approved):
LGTM!!!!

PR #130, opened 2021-06-21
Add simple menu on datepicker

HichamELBSI:

# DatePicker

## Description
Add month and year menu on DatePicker

## Demo

## API
No additional API

## Voiceover
datesr


packages/strapi-design-system/src/DatePicker/DatePicker.js
```diff
@@ -68,49 +96,69 @@ export const DatePicker = ({ initialDate, selectedDate, onChange, label, selecte
...
+                <Box paddingBottom={4}>
+                  <Row>
+                    <SimpleMenu id="year" label={months[date.getMonth()]}>
+                      {months.map((month) => (
+                        <MenuItem key={month} onClick={() => handleMonthChange(month)}>
+                          {month}
+                        </MenuItem>
+                      ))}
+                    </SimpleMenu>
+                    <SimpleMenu id="months" label={date.getFullYear()}>
```
mfrachet, line 109: There's a proptypes error on this I think (number and string)

mfrachet (approved):
Expect the PropTypes error, very nice job 😊

mfrachet:
Also, I've noted that when selecting a date and opening the picker again, it does select the current year and not the previously selected one 🤔 . is that wanted?

HichamELBSI:
> Also, I've noted that when selecting a date and opening the picker again, it does select the current year and not the previously selected one 🤔 . is that wanted?

Not sure I understand. When you select a date, the displayed month and year should be those of the selected date.

mfrachet:
Yes, but when I open the menu again, the date of the current year (2021) is highlighted in blue, even if the year in the menu is "2014"

packages/strapi-design-system/src/DatePicker/DatePicker.js
HichamELBSI: Nice catch ! thanks

PR #131, opened 2021-06-21
Add onReachEnd on popover

mfrachet:
# Add onReachEnd on popover

## Description

It should trigger only when the container is scrollable

### DIFF
--- a/packages/strapi-design-system/src/helpers/useIntersection.js
+++ b/packages/strapi-design-system/src/helpers/useIntersection.js
@@ -28,5 +28,5 @@ export const useIntersection = (scrollableAreaRef, callback, { selectorToWatch, 
     return () => {
       observer.disconnect();
     };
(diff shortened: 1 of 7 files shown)


packages/strapi-design-system/src/helpers/useIntersection.js
```diff
@@ -0,0 +1,32 @@
...
+    const observer = new IntersectionObserver(onEnterZone, options);
+    const target = scrollableAreaRef.current.querySelector(selectorToWatch);
+
+    observer.observe(target);
+
+    return () => {
+      observer.disconnect();
+    };
+  }, [skipWhen, callback, selectorToWatch]);
+};
```
HichamELBSI, line 32: Very nice !!

HichamELBSI (approved):
That is so cool! good job @mfrachet

PR #134, opened 2021-06-23
Time picker

mfrachet:
# TimePicker

## Description

Add a TimePicker component which is basically a "Select" with a bunch of computation on dates.

```jsx
 <TimePicker
  label={'Choose a time'}
  disabled={false}
  error={undefined}
  hint={'Description line'}
  id={'tp-1'}
  onClear={() => setValue(undefined)}
  onChange={setValue}
  value={value}
  clearLabel={'Clear the selected time picker value'}
  step={15} // number of minutes to count, default is 15 
/>
```

## VO

Selecting 00:45 in the TimePicker component using the keyboard

### DIFF
--- a/packages/strapi-design-system/src/Select/Select.js
+++ b/packages/strapi-design-system/src/Select/Select.js
@@ -8,6 +8,7 @@ import { Popover } from '../Popover';
 import { Stack } from '../Stack';
 import { Text } from '../Text';
 import { Row } from '../Row';
+import { Box } from '../Box';
 import { genId } from '../helpers/genId';
 import { SelectList } from './SelectList';
 import { SelectButtonWrapper, IconBox, CaretBox } from './components';
@@ -31,6 +32,7 @@ export const Select = ({
   onClear,
   onReachEnd,
   multi,
+  startIcon,
   ...props
 }) => {
   const idRef = useRef(id || genId());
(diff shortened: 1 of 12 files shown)


packages/strapi-design-system/src/Select/Select.js
```diff
@@ -83,7 +84,7 @@ export const Select = ({
     }
 
     return cloneElement(node, {
-      id: optionId,
+      id: escapeSelector(optionId),
```
mfrachet, line 87: ℹ️ Necessary since a selector like `00:00` can't work because of the `:`

packages/strapi-design-system/src/TimePicker/TimePicker.js
```diff
@@ -0,0 +1,99 @@
...
+      error={error}
+      value={value}
+      onChange={onChange}
+      disabled={disabled}
+      startIcon={
+        <TimeIconWrapper>
+          <TimeIcon />
+        </TimeIconWrapper>
+      }
+      {...props}
```
HichamELBSI, line 63: why do we need to explicitly pass the props `hint, label, error, value, disabled, etc` since it is implicitly passed in the props object?

HichamELBSI (changes requested):
Nice job!! I just put a comment about the props

packages/strapi-design-system/src/TimePicker/TimePicker.js
mfrachet: Exactly for what you mentioned: being explicit.

I like when it's easy to scan quickly which props is required in order for a component to work 😊 . Not having to jump between files and try to understand what is going under the hood.

I can spread the props object if you want, not against this, but this is the reason 😄

PR #136, opened 2021-06-24
Select multi tags

mfrachet:
# Select multi tags

## Description

Add a `withTags` props to the component

## Demo

Example on : https://design-system-git-select-multi-tags-strapijs.vercel.app/?path=/story/design-system-atoms-select--multi-with-tags

## API

```diff
<Select
    id="select1"
    label="Choose your meal"
    placeholder="Your example"
    onClear={() => setValues([])}
    hint="Description line"
    clearLabel="Clear the meal"
    error={error}
    value={values}
    onChange={onValueChange}
    disabled={disabled}
    customizeContent={(values) => `${values.length} currently selected`}
    multi
+  withTags
  >
    <Option value={'pizza'}>Pizza</Option>
    <Option value={'hamburger'}>Hamburger</Option>
    <Option value={'bagel'}>Bagel</Option>
</Select>
```

## Voiceover

Selecting pizza and hamburger in in listbox

### DIFF
--- a/packages/strapi-design-system/src/Select/Select.stories.mdx
+++ b/packages/strapi-design-system/src/Select/Select.stories.mdx
@@ -91,3 +91,52 @@ Description...
     }}
   </Story>
 </Canvas>
+
+## Select multi with tags
+
+Description...
+
+<Canvas>
+  <Story name="multi-withTags">
+    {() => {
+      const [values, setValues] = useState([]);
+      const [error, toggleError] = useState();
+      const [disabled, toggleDisabled] = useState();
+      const onValueChange = (newValue) =>
+        setValues((s) => (s.includes(newValue) ? s.filter((x) => x !== newValue) : [...s, newValue]));
+      return (
+        <Stack size={11}>
+          <h2>Current value is {values.join(', ')}</h2>
+          <Select
+            id="select1"
+            label="Choose your meal"
+            placeholder="Your example"
+            onClear={() => setValues([])}
+            hint="Description line"
+            clearLabel="Clear the meal"
+            error={error}
+            value={values}
+            onChange={onValueChange}
+            disabled={disabled}
+            customizeContent={(values) => `${values.length} currently selected`}
+            multi
+            withTags
+          >
+            <Option value={'pizza'}>Pizza</Option>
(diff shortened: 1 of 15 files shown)


soupette (approved):
LGTM

packages/strapi-design-system/src/Select/SelectTags.js
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <SelectTagsWrapper>
+      <Row wrap="wrap">
+        {tags.map((tag) => (
+          <SelectTag
+            icon={<CloseIcon />}
+            aria-disabled={disabled}
+            onClick={() => handleClick(tag.value)}
+            tabIndex={-1}
+            key={`tag-${tag.value}`}
```
HichamELBSI, line 35: this will work only if the tags are all different right?

HichamELBSI (changes requested):
Good job @mfrachet! I just put two comments

packages/strapi-design-system/src/Select/SelectTags.js
soupette: At some point they should be, if it happen to be problematic we can always make an update later to generate a unique id.

packages/strapi-design-system/src/Select/Select.stories.mdx
```diff
@@ -91,3 +91,52 @@ Description...
...
+            label="Choose your meal"
+            placeholder="Your example"
+            onClear={() => setValues([])}
+            hint="Description line"
+            clearLabel="Clear the meal"
+            error={error}
+            value={values}
+            onChange={onValueChange}
+            disabled={disabled}
+            customizeContent={(values) => `${values.length} currently selected`}
```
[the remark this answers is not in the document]
mfrachet: yes

packages/strapi-design-system/src/Select/SelectTags.js
mfrachet: What would be the value of having multiple time the same identifier in a list of options?  What is different from selecting an option "pizza" and an option "pizza"?

I personnally think they should be unique 😊

HichamELBSI (approved):
LGTM good job !!

PR #137, opened 2021-06-24
Homogeneize height of inputs

mfrachet:
This PR removes the padding top / bottom of the field input and make it be `40/16 rem` high. The aim is to make the alignment of the select component and field input possible.

Another iteration will come for the select component


soupette (approved):
LGTM

PR #140, opened 2021-06-25
Add prebuild test

soupette:

# Add prebuild test

## Description

The test makes sure all the components are correctly export before building the library.


packages/strapi-design-system/package.json
```diff
@@ -48,6 +48,7 @@
     "uuid": "^8.3.2"
   },
   "scripts": {
+    "prebuild": "jest --testRegex tests/check-exports.spec.js",
```
mfrachet, line 51: I think you can remove it from here since it's part of the "test" command

mfrachet (approved):

PR #141, opened 2021-06-28
ResizeObserver on Popover

mfrachet:
## Description

Adds a ResizeObserver on Popover so that it can repositioned itself when the source container grows such as the Input in the Select multi tags example

## Demo
Example on : https://design-system-git-resize-popover-strapijs.vercel.app/?path=/story/design-system-atoms-select--multi-with-tags

## In action

Filling multiple value in the select component


soupette (approved):
LGTM

PR #145, opened 2021-06-30
Prepare the Table component (UI)

mfrachet:
# First iteration on the Table component

## Description

Implement the UI part of the Table component

## Demo

Example on : https://design-system-git-table-strapijs.vercel.app/?path=/story/design-system-organisms-table--base

## API

```jsx
<Table colCount={COL_COUNT} rowCount={ROW_COUNT}>
  <Thead>
    <Tr>
      <Th>
        <BaseCheckbox aria-label="Select all entries" />
      </Th>
      <Th>
        <TableLabel as="span">ID</TableLabel>
      </Th>
      <Th>
        <TableLabel as="span">Cover</TableLabel>
      </Th>
      <Th>
        <TableLabel as="span">Description</TableLabel>
      </Th>
      <Th>
        <TableLabel as="span">Categories</TableLabel>
      </Th>
      <Th>
        <TableLabel as="span">Contact</TableLabel>
      </Th>
      <Th>
        <VisuallyHidden>Actions</VisuallyHidden>
      </Th>
    </Tr>
  </Thead>
  <Tbody>
    {entries.map((entry) => (
      <Tr key={entry.id}>
        <Td>
          <BaseCheckbox aria-label={`Select ${entry.name}`} />
        </Td>
        <Td>
          <Text textColor="neutral600" as="span">
            {entry.id}
          </Text>
        </Td>
        <Td>
          <Avatar src={entry.cover} alt={entry.contact} />
        </Td>
        <Td>
          <Text textColor="neutral600" as="span">
            {entry.description}
          </Text>
        </Td>
        <Td>
          <Text textColor="neutral600" as="span">
            {entry.category}
          </Text>
        </Td>
        <Td>
          <Text textColor="neutral600" as="span">
            {entry.contact}
          </Text>
        </Td>
        <Td>
          <IconButton onClick={() => console.log('edit')} title="Edit" noBorder>
            <EditIcon />
          </IconButton>
        </Td>
      </Tr>
    ))}
  </Tbody>
</Table>
```

## Voiceover

A user navigating the Table component using a keyboard and VoiceOver on Safari

### DIFF
--- a/packages/strapi-design-system/src/Table/Table.stories.mdx
+++ b/packages/strapi-design-system/src/Table/Table.stories.mdx
@@ -31,7 +31,7 @@ Description...
       const COL_COUNT = 10;
       const entry = {
         cover: 'https://avatars.githubusercontent.com/u/3874873?v=4',
-        description: 'Chez Léon is a human sized Parisian...',
+        description: 'Chez Léon is a human sized Parisian',
         category: 'French cuisine',
         contact: 'Leon Lafrite',
       };
--- a/packages/strapi-design-system/src/Table/Table.stories.mdx
+++ b/packages/strapi-design-system/src/Table/Table.stories.mdx
@@ -1,0 +1,112 @@
+<!--- Table.stories.mdx --->
+
+import { Meta, Story, Canvas } from '@storybook/addon-docs/blocks';
+import EditIcon from '@strapi/icons/EditIcon';
+import { Box } from '../Box';
+import { VisuallyHidden } from '../VisuallyHidden';
+import { BaseCheckbox } from '../BaseCheckbox';
+import { Table } from './Table';
+import { Thead } from './Thead';
+import { Tbody } from './Tbody';
+import { Tr } from './Tr';
+import { Td, Th } from './Cell';
+import { Text, TableLabel } from '../Text';
+import { Avatar } from '../Avatar';
+import { IconButton } from '../IconButton';
+
+<Meta title="Design System/Organisms/Table" component={Table} />
+
+# Table
+
+This is the doc of the `Table` component
+
+## Table
+
+Description...
+
+<Canvas>
+  <Story name="base">
+    {() => {
+      const ROW_COUNT = 6;
+      const COL_COUNT = 7;
+      const entry = {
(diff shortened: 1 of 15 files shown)


packages/strapi-design-system/src/Table/Table.stories.mdx
```diff
@@ -0,0 +1,112 @@
...
+                      {entry.category}
+                    </Text>
+                  </Td>
+                  <Td>
+                    <Text textColor="neutral600" as="span">
+                      {entry.contact}
+                    </Text>
+                  </Td>
+                  <Td>
+                    <IconButton onClick={() => console.log('edit')} title="Edit" noBorder>
```
gustavpursche, line 100: @mfrachet Our editors would really appreciate it, if this element could become a `<a>` instead of a `<button>`. In case they need to edit multiple entries, it's really helpful if they can go through the list and open each entry in a separate tab ("Open in new tab" of the browser context menu), which is not possible with buttons.

mfrachet: Hello there and thank you for the comment 😊 

Yes sure, thanks for the feedback, this is a great suggestion and I'll open a PR in order to make this IconXXX (IconToggle / IconPressable, do you have other suggestions for a more generic name?) possible 👍🏻 . I won't make the modification in this specific PR since it's not its scope but I'm taking a note on the side to implement it 😄 

Happy to see you there Gustav 😊

gustavpursche: I'd personally always go for `IconButton` which renders a `<button>` without a `href` attribute and a `<a>` with a `href` attribute. `IconLink` sounds legit to me too, but I'm not sure whether this is necessary ...

HichamELBSI:
The focus outline of the latest element in the table is sticking out. I wonder if we need to fix that kind of bugs. 

Screenshot 2021-07-05 at 10 04 51

mfrachet:
> The focus outline of the latest element in the table is sticking out. I wonder if we need to fix that kind of bugs.

I might be able to fix this with a negative offset, eventually 😊

HichamELBSI (approved):
LGTM good job! I don't have any comments except the focus border 😄

PR #147, opened 2021-06-30
Setting height and width on Avatar

mfrachet:
Add height and width HTML attribute on Avatar in order to reduce CLS

https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/


HichamELBSI (approved):
LGTM

PR #149, opened 2021-06-30
Add a Grid implementation using flex

mfrachet:
## Description

Why that?

When dealing with CSS grid, it's the parent that decides where to position its children (cols, rows, areas etc...) Using the flexbox approach, each child is responsible for handling its size on a scale from 0 to 12 (customizable)

## Demo

Example on : https://design-system-git-a-better-grid-strapijs.vercel.app/?path=/story/design-system-molecules-grid--complex-grid

## API

```jsx
 <Grid gap={4} padding={4}>
  <Box col={6} color="neutral0">
    First
  </Box>
  <Box col={6}>Second</Box>
  <Box col={9}>Third</Box>
  <Box col={3} background="secondary500">
    Last
  </Box>
</Grid>
```

### DIFF
--- a/example/src/App.test.js
+++ b/example/src/App.test.js
@@ -3,7 +3,7 @@ import { ThemeProvider } from "@strapi/design-system/ThemeProvider";
 import { lightTheme } from "@strapi/design-system/themes";
 import App from "./App";
 
-describe("App", () => {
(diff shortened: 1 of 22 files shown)


HichamELBSI (approved):
That's very nice! Good job @mfrachet

PR #157, opened 2021-07-05
Asset card

mfrachet:
# Card component

## Description

Add a composable card component

## Demo

Example on : https://design-system-git-asset-card-strapijs.vercel.app/?path=/story/design-system-molecules-card--base

## API

```jsx
<Cards>
  <Card id="1">
    <CardHeader>
      <CardCheckbox value={true} />
      <CardAction position="end">
        <span>Hello world</span>
      </CardAction>
      <CardAsset src={'./somewhere-it-belongs'} />
      <CardTimer>05:39</CardTimer>
    </CardHeader>
    <CardBody>
      <CardContent>
        <CardTitle>File name</CardTitle>
        <CardSubtitle>PNG - 400x400</CardSubtitle>
      </CardContent>
      <CardBadge>Doc</CardBadge>
    </CardBody>
  </Card>
</Cards>
```

Navigating between card using the keyboard


packages/strapi-design-system/src/Card/Cards.js
```diff
@@ -0,0 +1,76 @@
...
+          allCards.item(allCards.length - 1).focus();
+        }
+
+        break;
+      }
+
+      default:
+        break;
+    }
+  };
```
HichamELBSI, line 73: Good job !!

HichamELBSI (approved):
LGTM good job!

maevalienard:
This is terrifying, how do you develop a component THAT FAST? 😂 I just specified it yesterday

- Would you mind adding the tile background? :)
- This isn't the right "×", make sure you use this symbol please :D
Capture d’écran 2021-07-06 à 15 00 29
- Is it the right background color for the badge component? I feel like it's very light 🤔 
- Can we pass this background to a neutral800? Looks too dark to me atm
Capture d’écran 2021-07-06 à 15 05 48
- Globally, is it the right height for this part, looks a bit tiny doesn't it?
Capture d’écran 2021-07-06 à 15 06 47

mfrachet:
cc @maevalienard I'm opening a new PR for adjusting this. 

The badge component is the same as http://localhost:6006/?path=/story/design-system-atoms-badge--base . Changing its color will make it change everywhere

I've modified the timer background to be 800

>  - Globally, is it the right height for this part, looks a bit tiny doesn't it? 

From Figma, it's 88px height, it's 87px height (I'm moving it to 88px right now) but it won't make a huge difference 😓 .
I think the problem is more the ratio of the image.

Concerning the background tiles, I'm on it for another PR

PR #163, opened 2021-07-06
Header layout

mfrachet:
# HeaderLayout

## Description
[add description]

## Demo

## API

```jsx
 <HeaderLayout
  sticky
  navigationAction={
    <Link startIcon={<BackIcon />} to="/">
      <VisuallyHidden>Go back</VisuallyHidden>
    </Link>
  }
  primaryAction={<Button startIcon={<AddIcon />}>Add an entry</Button>}
  secondaryAction={
    <Button variant="tertiary" startIcon={<EditIcon />}>
      Edit
    </Button>
  }
  title={<H2>Restaurants</H2>}
  subtitle={
    <Text small={true} textColor="neutral600">
      36 entries found
    </Text>
  }
/>
```

## Voiceover

Screenshot 2021-07-06 at 09 32 39
Screenshot 2021-07-06 at 09 32 48

### DIFF
--- a/packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
+++ b/packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
@@ -1,0 +1,38 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Box } from '../Box';
+import { Row } from '../Row';
+import { Stack } from '../Stack';
+
+export const HeaderLayout = ({ navigationAction, primaryAction, secondaryAction, subtitle, title, sticky }) => {
+  return (
+    <Box paddingLeft={10} paddingRight={10} paddingBottom={10} paddingTop={navigationAction ? 6 : 10}>
+      {navigationAction ? <Box paddingBottom={3}>{navigationAction}</Box> : null}
+      <Row justifyContent="space-between">
+        <Row>
+          {title}
+          {secondaryAction ? <Box paddingLeft={4}>{secondaryAction}</Box> : null}
+        </Row>
+        {primaryAction}
+      </Row>
+      <Box paddingTop={1}>{subtitle}</Box>
+    </Box>
+  );
+};
+
+HeaderLayout.defaultProps = {
+  navigationAction: undefined,
+  primaryAction: undefined,
+  secondaryAction: undefined,
+  subtitle: undefined,
+  sticky: undefined,
+};
+
+HeaderLayout.propTypes = {
+  navigationAction: PropTypes.node,
+  primaryAction: PropTypes.node,
+  secondaryAction: PropTypes.node,
+  sticky: PropTypes.bool,
+  subtitle: PropTypes.node,
+  title: PropTypes.node.isRequired,
+};
--- a/packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
+++ b/packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
@@ -34,7 +34,7 @@ export const HeaderLayout = ({ navigationAction, primaryAction, secondaryAction,
         </Row>
         {primaryAction}
       </Row>
-      <Box paddingTop={1}>{subtitle}</Box>
+      {subtitle}
     </Box>
   );
 };
--- a/packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
+++ b/packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
@@ -44,7 +44,7 @@ HeaderLayout.defaultProps = {
   primaryAction: undefined,
   secondaryAction: undefined,
   subtitle: undefined,
-  sticky: undefined,
+  sticky: false,
 };
 
 HeaderLayout.propTypes = {
--- a/packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
+++ b/packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
@@ -2,9 +2,28 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Box } from '../Box';
 import { Row } from '../Row';
-import { Stack } from '../Stack';
 
 export const HeaderLayout = ({ navigationAction, primaryAction, secondaryAction, subtitle, title, sticky }) => {
+  if (sticky) {
+    return (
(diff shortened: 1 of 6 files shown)


HichamELBSI (approved):
LGTM good job !!

maevalienard:
Ciao bambino,

First of all, great job :D 

- Not sure if your buttons have a SmallButton text style. I feel like it uses more like a SmallText style actually. That would explain the "strange" balance between the icon size and the text :) 
- Why did you put the Edit button on the right side of the sticky header? :) 
- On the fixed header, the text button should say "back" and not "go back" but I'm okay with both wording. We just need to make a decision and I'd edit the design accordingly :) 
- Is it the right text style font for the description of the fixed header? I find it slightly too bold but it's probably just my eyes. 
- Would it be possible to remove the 4px spacing between the H1 and the description in the fixed header, please? I feel like the balance is better without :) 
- I feel like there's a slight misalignment for that component. The arrow seems upper to me than the label. 
Capture d’écran 2021-07-06 à 14 41 40

Many thanks in advance 🙏

maevalienard (changes requested):
See my previous comment (still not very familiar with GitHub uh sorry for that)

mfrachet:
> Not sure if your buttons have a SmallButton text style. I feel like it uses more like a SmallText style actually. That would explain the "strange" balance between the icon size and the text :)

The buttons are the exact ones defined in https://design-system-alpha.vercel.app/?path=/story/design-system-atoms-button--variants

> Why did you put the Edit button on the right side of the sticky header? :)

It's just a button, we can change the label on the fly, it's not an "edit" button, but a button with a "edit" label

> On the fixed header, the text button should say "back" and not "go back" but I'm okay with both wording. We just need to make a decision and I'd edit the design accordingly :)

Same as above, this is not fixed content. We can change it from Strapi core side

> Is it the right text style font for the description of the fixed header? I find it slightly too bold but it's probably just my eyes.

I'm using Text body small from https://design-system-alpha.vercel.app/?path=/story/design-system-molecules-text--base

> Would it be possible to remove the 4px spacing between the H1 and the description in the fixed header, please? I feel like the balance is better without :)

Yes sure 😊 Can we also remove it from the documentation in Figma to stay consistent 🙏🏻 ?

> I feel like there's a slight misalignment for that component. The arrow seems upper to me than the label.

This is the link component in https://design-system-alpha.vercel.app/?path=/story/design-system-atoms-link--base . Changing it will make it change everywhere

maevalienard:
> The buttons are the exact ones defined in https://design-system-alpha.vercel.app/?path=/story/design-system-atoms-button--variants

Might not be the right text-style in the buttons then :/ I struggle to verify :( 

> It's just a button, we can change the label on the fly, it's not an "edit" button, but a button with a "edit" label

I understood, the thing is would that be a possibility to have the Edit button right next to the maintitle & description then? Whether it's an "Edit" or not since you mentioned the label doesn't really matter, and whether it's optional. It's just because it's the CTB the Edit button remains next to that part :) 

> Same as above, this is not fixed content. We can change it from Strapi core side

Which one do you like best then? :) 

> I'm using Text body small from https://design-system-alpha.vercel.app/?path=/story/design-system-molecules-text--base

I think you mean you used that on the sticky header, I meant on the default one. Sorry I wasn't clear :) 
 

> Yes sure 😊 Can we also remove it from the documentation in Figma to stay consistent 🙏🏻 ?

Absolutely 🌻 


> This is the link component in https://design-system-alpha.vercel.app/?path=/story/design-system-atoms-link--base . Changing it will make it change everywhere

No problemo, the misalignment is global :)

packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
```diff
@@ -0,0 +1,57 @@
...
+          <Row>
+            <Box paddingRight={3}>{navigationAction}</Box>
+            <Box>
+              {title}
+              {subtitle}
+            </Box>
+          </Row>
+          <Row>
+            {secondaryAction}
+            {primaryAction ? <Box paddingLeft={2}>{primaryAction}</Box> : undefined}
```
soupette, line 20: You can do `primaryAction && <Box>...</Box>`

packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
```diff
@@ -0,0 +1,57 @@
...
+          </Row>
+        </Row>
+      </Box>
+    );
+  }
+
+  return (
+    <Box paddingLeft={10} paddingRight={10} paddingBottom={10} paddingTop={navigationAction ? 6 : 10}>
+      {navigationAction ? <Box paddingBottom={3}>{navigationAction}</Box> : null}
```
soupette, line 29: Same here you can remove the ternary

packages/strapi-design-system/src/HeaderLayout/HeaderLayout.js
```diff
@@ -0,0 +1,57 @@
...
+    </Box>
+  );
+};
+
+HeaderLayout.defaultProps = {
+  navigationAction: undefined,
+  primaryAction: undefined,
+  secondaryAction: undefined,
+  subtitle: undefined,
+  sticky: undefined,
```
soupette, line 47: Why not setting it to false?

soupette (approved):
LGTM

maevalienard:
Did you push the changes on vercel yet?

mfrachet:
> Did you push the changes on vercel yet?

Yes they are

maevalienard:
Even for the boldness of the buttons? 🤔

mfrachet:
I do think so, it uses the same as the one from this story: https://design-system-git-fix-link-button-strapijs.vercel.app/?path=/story/design-system-atoms-button--variants . Something looks not okay?

maevalienard:
I hard refresh and still see a font-weight: 400 for the button labels. Also it seems like the Text button component still have the old alignment (not sure about that one but pretty sure about the buttons) :)

mfrachet:
@maevalienard is it still the same on https://design-system-alpha.vercel.app/?path=/story/design-system-layouts-headerlayouts--base ? Maybe I didn't rebase before merging and so, you see an older version 😓

maevalienard:
Looks goooood 🚀 Thank you!

PR #171, opened 2021-07-07
Add Carousel component

mfrachet:
# Carousel

## Description

Add a Carousel component

## Demo

Example on : https://design-system-git-carousel-strapijs.vercel.app/?path=/story/design-system-molecules-carousel--base

## API

```jsx
<Carousel
  label={`Carousel of numbers (2/3)`}
  selectedSlide={1}
  previousLabel="Previous slide"
  nextLabel="Next slide"
  onNext={() => {}}
  onPrevious={() => {}}
  actions={
    <CarouselActions>
      <IconButton onClick={() => console.log('edit')} title="Edit">
        <span>First</span>
      </IconButton>
      <IconButton onClick={() => console.log('Create')} title="Create">
        <span>Second</span>
      </IconButton>
      <IconButton onClick={() => console.log('Delete')} title="Delete">
        <span>Third</span>
      </IconButton>
      <IconButton onClick={() => console.log('Publish')} title="Publish">
        <span>Fourth</span>
      </IconButton>
    </CarouselActions>
  }
>
  <CarouselSlide label="1 of 3 slides">
    <CarouselImage src={'firstPathImg'} alt="First" />
  </CarouselSlide>
  <CarouselSlide label="2 of 3 slides">
    <CarouselImage src={'firstPathImg'} alt="second" />
  </CarouselSlide>
  <CarouselSlide label="3 of 3 slides">
    <CarouselImage src={'firstPathImg'} alt="third" />
  </CarouselSlide>
</Carousel>
```

## Voiceover

Navigating the carousel component using arrow right and arrow left and voiceover


maevalienard:
hola 👋 

From what I see:
- Arrows color should be neutral600. Can you add a hover state on them where the color would pass to a neutral900 please? :) 
- These are not the right icons in the IconOnly buttons, no big deal? Asking becomes I notice the tooltip wording wouldn't be the right ones then :) 
- No big deal to add a description line if it's necessary either? :) 

Great job as usual! Many thanks! 🔥

mfrachet:
> Arrows color should be neutral600. Can you add a hover state on them where the color would pass to a neutral900 please? :)

I've pushed a fix :)

> These are not the right icons in the IconOnly buttons, no big deal? Asking becomes I notice the tooltip wording wouldn't be the right ones then :)

It's custom, people can add whatever they want 😊 

> No big deal to add a description line if it's necessary either? :)

They can do stuff inside the slides but they will have to still it themselves. Do you have something in particular in mind?

> When on focus state and reach the third picture, an overlay appears on the image 

Not sure to understand what you mean 🤔

maevalienard:
When I click on an arrow, now the hover color remains on the arrow :D 

> Not sure to understand what you mean 🤔

It doesn't do it anymore, so forget it 😄 


> They can do stuff inside the slides but they will have to still it themselves. Do you have something in particular in mind?

I meant below the carousel, normally you should have a description line or at least an error message :)

packages/strapi-design-system/src/Carousel/Carousel.js
```diff
@@ -0,0 +1,121 @@
...
+import BackFilterIcon from '@strapi/icons/BackFilter';
+import { Box } from '../Box';
+import { Text } from '../Text';
+import { KeyboardKeys } from '../helpers/keyboardKeys';
+
+const CarouselGrid = styled.div`
+  display: grid;
+  grid-template-columns: auto 1fr auto;
+  grid-template-areas: 'startAction slides endAction';
+  position: relative;
```
HichamELBSI, line 14: Can't we use the Grid component ?

HichamELBSI (commented):
Very good job on this Marvin! I just have one comment here 👍🏽

mfrachet:
> I meant below the carousel, normally you should have a description line or at least an error message :)

Could you reflect it on the documentation please? It's strongly bound to the carousel itself?

packages/strapi-design-system/src/Carousel/Carousel.js
mfrachet: Not in this case 😊 . I wanted the focus order to remain the following: Left arow, right arrow, then actions.

Problem is that, if I wanted to keep this order, I need the two buttons to be defined in the DOM before the list of actions.

Our grid component is for a X column based grid but we don't manage areas. With the Carousel grid, I can place the element in the order I want and so position the right arrow after the slides, without breaking the tab navigation order 😊

maevalienard:
Is the description line optional? :) Looks good to me!

I have a last question: is this carousel responsive in width? It seems like it has been dev in a smaller width than what it will be in a form (in the CM) for instance. I can show you what I mean with screenshots directly if you'd like

mfrachet:
> Is the description line optional? :) Looks good to me!

Yes, it's optional :) 

> I have a last question: is this carousel responsive in width? It seems like it has been dev in a smaller width than what it will be in a form (in the CM) for instance. I can show you what I mean with screenshots directly if you'd like

I forced the carousel to have a specific width to match what we have in the documentation pane **only for the demo**.

**However**, the height of the image is fixed to `124px` as in the documentation. If we want to use other constraints (100% of the image height etc...), I would need to know that before merging this PR 😊

maevalienard:
The height doesn't change (either for the component itself than for the asset), only the width :) 

-> We would like to control the height of the grid as much as possible

soupette (approved):
LGTM great work!

PR #172, opened 2021-07-09
Modify header layout for title and subtitle props

mfrachet:
- title and subtitle props are now string to enforce H1 tag on this header
- add a shadow on the button to create the effect wanted in figma for the tertiary button

UI doesn't change that much except for the shadow


HichamELBSI (approved):
LGTM

PR #175, opened 2021-07-09
Text and paragraph

mfrachet:
Add a `Paragraph` component to prevent from doing `<Text as="span" />` everywhere (since `Text` was a `p` tag before)

Now:

- `Text` => `span`
- `Paragraph` => `p`

### DIFF
--- a/packages/strapi-design-system/src/Text/index.js
+++ b/packages/strapi-design-system/src/Text/index.js
@@ -1,3 +1,4 @@
+import React from 'react';
 import styled from 'styled-components';
 
 const handleColor = ({ theme, textColor }) => theme.colors[textColor];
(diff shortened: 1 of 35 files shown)


packages/strapi-design-system/src/Text/index.js
```diff
@@ -34,13 +35,15 @@ const textStyles = {
...
 
-export const Text = styled.p`
+export const Text = styled.span`
   font-weight: ${({ highlighted }) => (highlighted ? 500 : 400)};
   font-size: ${({ small }) => textStyles[small ? 'S' : 'M'].fontSize};
   line-height: ${({ small }) => textStyles[small ? 'S' : 'M'].lineHeight};
   color: ${handleColor};
 `;
 
+export const Paragraph = (props) => <Text as="p" {...props} />;
```
HichamELBSI, line 45: I'm afraid it's going to get long to write `Paragraph` every time. WDYT of renaming that one to `<P />`?

HichamELBSI (changes requested):

packages/strapi-design-system/src/Text/index.js
mfrachet: Yes sure

PR #193, opened 2021-07-21
Replace all HTML links by react-router-dom links

mfrachet:
Basically moves from the `<a>` tag to the `<NavLink>` react-router-dom component (and adjust the according tests)


soupette (approved):
LGTM!!!

PR #198, opened 2021-07-22
Add zoom to the Avatar component

mfrachet:
# Avatar

## Description

Add the possibility to display a "bigger" picture when hovering an avatar

## Demo

Example on : https://design-system-git-avatar-zoom-strapijs.vercel.app/?path=/story/design-system-atoms-avatar--base

## API

```diff
<Avatar
  src="https://avatars.githubusercontent.com/u/3874873?v=4"
  alt="marvin frachet"
+  preview="https://some-unknown-photo/x.png"
/>
```

or the following when the image is the same

```diff
<Avatar
  src="https://avatars.githubusercontent.com/u/3874873?v=4"
  alt="marvin frachet"
+  preview
/>
```

### DIFF
--- a/packages/strapi-design-system/src/Avatar/Avatar.js
+++ b/packages/strapi-design-system/src/Avatar/Avatar.js
@@ -1,21 +1,54 @@
-import React from 'react';
+import React, { useState } from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
 
-const avatarSize = '26px';
+const avatarSize = 26;
+const previewSize = 64;
 
 const AvatarImg = styled.img`
   border-radius: 50%;
   display: block;
-  height: ${avatarSize};
-  width: ${avatarSize};
 `;
 
-export const Avatar = ({ src, alt }) => {
-  return <AvatarImg src={src} alt={alt} width={avatarSize} height={avatarSize} />;
+const PreviewContainer = styled.img`
+  border-radius: 50%;
+  position: absolute;
+  transform: translate(-${(previewSize - avatarSize) / 2}px, -100%);
+  margin-top: -${({ theme }) => theme.spaces[1]};
+`;
+
+export const Avatar = ({ src, alt, preview }) => {
+  const [previewVisible, setPreviewVisible] = useState(false);
+
+  return (
+    <span>
+      {preview && previewVisible && (
+        <PreviewContainer
+          aria-hidden
+          alt=""
+          width={`${previewSize}px`}
+          height={`${previewSize}px`}
+          src={preview === true ? src : preview}
+        />
+      )}
+      <AvatarImg
+        src={src}
+        alt={alt}
+        width={`${avatarSize}px`}
+        height={`${avatarSize}px`}
+        onMouseEnter={() => setPreviewVisible(true)}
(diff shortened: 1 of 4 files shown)


maevalienard (approved):
LGTM, great job! Thank you :D

maevalienard (changes requested):
We should have a slight overlay on the avatar (not the zoom preview but the avatar itself), "The specific avatar gets a 60% neutral0 overlay on top of it" :)

maevalienard (approved):
Awesome, thank you!

packages/strapi-design-system/src/Avatar/Avatar.js
```diff
@@ -1,21 +1,66 @@
...
+        <PreviewContainer
+          aria-hidden
+          alt=""
+          width={`${previewSize}px`}
+          height={`${previewSize}px`}
+          src={preview === true ? src : preview}
+        />
+      ) : null}
+
+      <AvatarImgWrapper onMouseEnter={() => setPreviewVisible(true)} onMouseLeave={() => setPreviewVisible(false)}>
```
HichamELBSI, line 50: Should we change this wrapper to an inline element so that the mouse over will be triggered only on the AvatarImg and not on the div row 🤔  ?

avatar

HichamELBSI: Also, should we handle the focus on that component so we can trigger the zoom with the keyboard?

packages/strapi-design-system/src/Avatar/Avatar.js
```diff
@@ -1,21 +1,66 @@
...
-  return <AvatarImg src={src} alt={alt} width={avatarSize} height={avatarSize} />;
+const AvatarImgWrapper = styled.div`
+  position: relative;
+`;
+
+const PreviewContainer = styled.img`
+  border-radius: 50%;
+  position: absolute;
+  transform: translate(-${(previewSize - avatarSize) / 2}px, -100%);
+  margin-top: -${({ theme }) => theme.spaces[1]};
```
HichamELBSI, line 22: Maybe we will need to add some zIndex here wdyt ? 🤔

HichamELBSI (changes requested):

packages/strapi-design-system/src/Avatar/Avatar.js
mfrachet: do we need a z-index? 🤔

packages/strapi-design-system/src/Avatar/Avatar.js
mfrachet: I'm wondering about the impact of this? The image is already so small 🤔 . Do you think we could benefit from this?

HichamELBSI: We can keep it like that and see if there is any feedback on the use 👍🏽

packages/strapi-design-system/src/Avatar/Avatar.js
HichamELBSI: Probably not, we'll see on the use 👍🏽

HichamELBSI (approved):
LGTM

PR #200, opened 2021-07-22
Keyboard navigator

mfrachet:
# Keyboard navigator

## Description

The idea is to create a component that can wrap a DOM tree and allows to navigate using the keyboard.

- ArrowUp / ArrowLeft moves the focus to the next element
- ArrowDown / ArrowRight moves the focus to the previous element
- End / Home moves the focus to the first/last element

This is practical for grouping accordions for example, or a set of cards (like in the MediaLibrary). Instead of duplicating the same logic in both places, I've gathered it inside this single component

## Demo

## API

```jsx
<KeyboardNavigable tagName="" attributeName="">
  <DomTree />
</KeyboardNavigable>
```

## Demo

Navigating using the keyboard in the media library


HichamELBSI (approved):
LGTM ! That's awesome thanks 😄

PR #202, opened 2021-07-22
Fix modal layout

mfrachet:
The ModalLayout was showing below the overlay in the CRA app because the modal content was set as a sibling of the modal background

Before:

A modal visually not correctly positioned

After:

A modal visually well positioned


HichamELBSI (approved):
LGTM

PR #204, opened 2021-07-26
Fix DatePicker QA / Refactoring

mfrachet:
- Take into consideration the QA from @maevalienard 
- Split the file in multiple for clarity
- Added some E2E tests to keep the basic behaviour

### DIFF
--- a/packages/strapi-design-system/jest.e2e.js
+++ b/packages/strapi-design-system/jest.e2e.js
@@ -9,6 +9,7 @@ module.exports = {
         // headless: false,
         // devtools: true,
       },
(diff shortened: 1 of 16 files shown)


packages/strapi-design-system/jest.e2e.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
         // headless: false,
         // devtools: true,
       },
+      contextOptions: { locale: 'en-EN' },
```
mfrachet, line 12: ℹ️ Allows the date to be formatted in en/EN when using the INTL API

PR #214, opened 2021-07-30
Fix broken UI

mfrachet:
- Fix the main nav position sticky issue
- Make sure the table is scrollable in the different layouts
- Fixing the card asset that was overflowing outside the card in the grid layout

To test, make sure to start the example app with the local version of parts and navigate in the different pages (settings and media libs)

### DIFF
--- a/packages/strapi-parts/src/MainNav/MainNav.js
+++ b/packages/strapi-parts/src/MainNav/MainNav.js
@@ -7,8 +7,10 @@ import { MainNavContext } from './MainNavContext';
 const MainNavWrapper = styled(Grid)`
   width: ${({ condensed }) => (condensed ? 'max-content' : `${224 / 16}rem`)};
   background: ${({ theme }) => theme.colors.neutral0};
-  height: 100%;
-  position: relative;
+  position: sticky;
+  top: 0;
+  height: 100vh;
+  z-index: 2;
 `;
 
 export const MainNav = ({ condensed, ...props }) => {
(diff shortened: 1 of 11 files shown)


packages/strapi-parts/src/MainNav/MainNav.js
```diff
@@ -7,8 +7,10 @@ import { MainNavContext } from './MainNavContext';
 const MainNavWrapper = styled(Grid)`
   width: ${({ condensed }) => (condensed ? 'max-content' : `${224 / 16}rem`)};
   background: ${({ theme }) => theme.colors.neutral0};
-  height: 100%;
-  position: relative;
+  position: sticky;
+  top: 0;
+  height: 100vh;
+  z-index: 2;
```
mfrachet, line 13: ℹ️ This is necessary, else, the nav condense button goes under the sub nav bar that has a position sticky and that is positionned after in the DOM (meaning the subnavbar would be above)

packages/strapi-parts/src/SubNav/SubNav.js
```diff
@@ -8,12 +8,12 @@ export const subNavWidth = `${232 / 16}rem`;
 const SubNavWrapper = styled(Grid)`
   width: ${subNavWidth};
   background: ${({ theme }) => theme.colors.neutral100};
-  height: 100%;
   position: sticky;
   top: 0;
   height: 100vh;
   overflow-y: auto;
   border-right: 1px solid ${({ theme }) => theme.colors.neutral200};
+  z-index: 1;
```
mfrachet, line 16: ℹ️ Making sure this one is lower than the main nav one

HichamELBSI (approved):
LGTM good job @mfrachet !!

PR #218, opened 2021-08-02
Add scoped css reset alternative in the `ThemeProvider` ​

fdel-car:
## Description
​
Add the possibility to pass `scopedCss` props to the `ThemeProvider` (the final name is up to you).
If `scopedCss` is true, then a styled div will wrap the `children` making the css reset only applied to them, this div will have all the global css reset properties tweaked a bit in order to be scoped.
​
I chose not to split the style reset and the `ThemeProvider` since the style reset uses the theme at some point. It would probably lead to mistakes if they were splitted 🤔

~I did not properly test this PR, since the tests were already failing on `develop` unfortunately.~
Edit: forget the line above, I just didn't understand that I needed to build the icons, and didn't notice that export checker was complaining only on this branch 🙃

## API

```jsx
<ThemeProvider theme={lightTheme} scopedCss>
  <Page />
</ThemeProvider>
```


mfrachet:
Let's try this 👍🏻 Thanks for the pr!

PR #219, opened 2021-08-03
Fix the focus when closing the select with a mouse

mfrachet:
## Description

Ensure to restore the focus on the select button only when it's closed using a keyboard

## Screenshots

Opening the select with a mouse and closing it to see that the focus is not sent to the select button. Then, opening the select using the keyboard and seeing that the focus is well sent to the select button

### DIFF
--- a/packages/strapi-parts/src/Select/hooks/useButtonRef.js
+++ b/packages/strapi-parts/src/Select/hooks/useButtonRef.js
@@ -6,7 +6,7 @@ export const useButtonRef = (expanded) => {
   const mountedRef = useRef(null);
   /**
    * Allows to make sure to re-send the focus only when the last action was
-   * a triggered by a keyboard event
+   * triggered by a keyboard event
    */
   const previousState = useRef();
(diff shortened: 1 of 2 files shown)


packages/strapi-parts/src/Select/hooks/useButtonRef.js
```diff
@@ -1,14 +1,26 @@
 import { useEffect, useRef } from 'react';
+import { DownState, UpState } from '../constants';
 
 export const useButtonRef = (expanded) => {
   const buttonRef = useRef(null);
   const mountedRef = useRef(null);
+  /**
+   * Allows to make sure to re-send the focus only when the last action was
+   * a triggered by a keyboard event
```
mfrachet, line 9:
```suggestion
   * triggered by a keyboard event
```

ronronscelestes (commented):
Nice!

HichamELBSI (approved):
LGTM 👍🏽

PR #223, opened 2021-08-04
Fix id generation for multiple bundles

mfrachet:
## Description

Before this PR, we were generating IDs by incrementing a number. Problems came when we splitted the components in different bundles owning their own ways to generate IDs. Each of these bundle was starting the ID counter at 0 and could create a collision between modules.

This PR aims to provide a prefix for components needing to generate unique IDs. Since we split the different bundles at the component level, the prefix + increment should be enough to ensure unique IDs across the app.

## A new way to generate IDs

The following hook allows to generate an automated id with a prefix OR to use the passed ID:

```jsx
// When predefinedId exists, it's used instead of the generated id and the prefix
const predefinedId = 'abcd'
const generatedId = useId('my-prefix', predefinedId)
```


HichamELBSI (approved):
Good job ! Thanks for that

PR #224, opened 2021-08-04
IconButton state style && disabled state onClick handled (+Button compo)

ronronscelestes:
## What does it do?
- Style for IconButtonGroup updated
- Style for IconButton disabled created
- onClick event does nothing if disabled state (for IconButton && Button compo)

## Why is it needed?
- DS documentation synchro
- To prevent onClick event to fire when using keyboard

## How to test it?
StoryBook!

### DIFF
--- a/packages/strapi-parts/src/Button/Button.js
+++ b/packages/strapi-parts/src/Button/Button.js
@@ -35,31 +35,39 @@ export const ButtonWrapper = styled(BaseButton)`
   ${getVariantStyle}
 `;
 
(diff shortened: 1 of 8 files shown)


mfrachet (approved):
Very nice job! LGTM 😊 

You will need to update your snapshots since some values have changed in the test :). Apart from that, everything is okay :)

maevalienard:
For the disable state, the icon color should be neutral600 :D

PR #225, opened 2021-08-05
NumberInput

mfrachet:
## Description

A NumberField component, using the mechanisms and tools used by react-spectrum on their own.

This component uses https://github.com/adobe/react-spectrum/tree/main/packages/%40internationalized/number to format and parse the number coming and leaving the component.

Note: the formatting step is only triggered when blurring the component

## Demo

Example on : https://design-system-git-number-input-strapijs.vercel.app/?path=/story/design-system-molecules-numberinput--base

## API

```jsx
<NumberInput
  placeholder="This is a content placeholder"
  label="Content"
  name="content"
  hint="Description line"
  error={undefined}
  onValueChange={(value) => setContent(value)}
  value={content}
  labelAction={
    <Tooltip description="Content of the tooltip">
      <button
        aria-label="Information about the email"
        style={{ border: 'none', padding: 0, background: 'transparent' }}
      >
        <HelpIcon aria-hidden={true} />
      </button>
    </Tooltip>
  }
  step={1} /* this is the default value */
/>
```

## Voiceover

[voiceover gif]


gu-stav:
@mfrachet Sorry, for randomly commenting on a PR again :) I believe the decimal trigger (".") should be defined through the `Intl` API, because different countries use different notations (in Germany it's ",") for example. It's often a source of confusion ...

mfrachet:
@gu-stav no worries, I really appreciate your comments and intervention 😊 . I agree with you, totally :). My biggest trouble si to pass from a string to a number and then being able to use the Intl API. In my scenario, use `parseFloat` only works when using a  dot 😭 . Do you have any ideas how can I make this a bit more practical for intl users?

gu-stav:
Oh, you are right. I wasn't aware of that there is no `Intl.NumberFormat.parse()` 😢 I've checked a couple of stackoverflow discussion and and came up with this function:

```js
const parseLocaleNumber = (stringNumber, locale) => {
    const formatter = Intl.NumberFormat(locale);
    const [, { value: thousandSeparator }] = formatter.formatToParts(11111);
    const [, { value: decimalSeparator }] = formatter.formatToParts(1.1);

    return parseFloat(stringNumber
        .replace(new RegExp('\\' + thousandSeparator, 'g'), '')
        .replace(new RegExp('\\' + decimalSeparator), '.')
    );
}

parseLocaleNumber('3.500,59', 'de-DE') 
// 3500.59

parseLocaleNumber('3500,59', 'de-DE') 
// 3500.59

parseLocaleNumber('3500', 'de-DE') 
// 3500
```

This would always give you the english representation of the input, which you could then check and manipulate. What do you think - would that solve the problem?

mfrachet:
That looks promising! I'll investiguate your suggestion 😊 . Thank you for the hint! 🙏🏻

packages/strapi-parts/src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,155 @@
...
+    const [inputValue, setInputValue] = useState(value || '');
+    const generatedId = useId('numberinput', id);
+    const numberParserRef = useRef(new NumberParser(getDefaultLocale()));
+    const numberFormaterRef = useRef(new NumberFormatter(getDefaultLocale()));
+
+    useEffect(() => {
+      if (value !== undefined) {
+        setInputValue(numberFormaterRef.current.format(value));
+      }
+    }, [value]);
```
mfrachet, line 38: ℹ️ Everytime, (except when value is undefined), when the outside value changes, we format the content of the input with the localized value formated

packages/strapi-parts/src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,155 @@
...
+
+    const handleChange = (e) => {
+      if (numberParserRef.current.isValidPartialNumber(e.target.value)) {
+        setInputValue(e.target.value);
+      }
+    };
```
mfrachet, line 44: ℹ️ We only modify the internal value on change. The external one will be modified on blur and then formatted back thanks to the useEffect call

packages/strapi-parts/yarn.lock
```diff
@@ -1283,6 +1290,13 @@
   dependencies:
     "@hapi/hoek" "^9.0.0"
 
+"@internationalized/number@^3.0.2":
```
mfrachet, line 1293: ℹ️ this is an adobe package

packages/strapi-parts/src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,155 @@
...
+          break;
+      }
+    };
+
+    const handleBlur = () => {
+      const parsedValue = numberParserRef.current.parse(inputValue);
+
+      if (isNaN(parsedValue)) {
+        onValueChange(undefined);
+        setInputValue(undefined);
```
mfrachet, line 88: ℹ️ SInce the useEffect will not format the internal input and the value is modified, I reset the internal state here to undefined when blurring on a NaN value

packages/strapi-parts/src/NumberInput/NumberInput.js
```diff
@@ -0,0 +1,158 @@
...
+      if (isNaN(parsedValue)) {
+        onValueChange(undefined);
+        setInputValue(INITIAL_VALUE);
+      } else {
+        onValueChange(numberParserRef.current.parse(inputValue));
+      }
+    };
+
+    return (
+      <div>
```
HichamELBSI, line 98: Why do we need that div ?

HichamELBSI (commented):
I tested the component and it looks pretty good to me! Very good work on this topic @mfrachet. I just put a comment about the `div` that wraps the field.

PR #226, opened 2021-08-05
Handle viewport overflow top && left for tooltip

ronronscelestes:
## What does it do?
Update tooltip left, right and top position when they overflow out of then screen + tests

## View
Capture d’écran 2021-08-05 à 14 42 01
Capture d’écran 2021-08-05 à 14 42 13
Capture d’écran 2021-08-06 à 16 17 21


mfrachet (changes requested):
That looks amazing 😊 , thank you for this :D . Could you, please, add tests for verifying these behaviors? 😊

mfrachet:
And what happens if it's positioned on the right side of the screen. 😛

mfrachet (changes requested):
A last tiny comment, and then, you can merge 😊

PR #227, opened 2021-08-05
Prevent popover from overflowing on the right side of the screen

mfrachet:
## Description
Prevents the popover from overflowing when it's out of the screen on the right side

## Demo
Example on : https://design-system-git-popover-overflow-strapijs.vercel.app/?path=/story/design-system-molecules-popover--overflow-right

Popover not overflowing on the screen


HichamELBSI (approved):
LGTM good job !!

PR #230, opened 2021-08-06
Removing aria-hidden from NavBrand

mfrachet:
aria-hidden on the NavBrand is not necessary since passing an empty alt in the image would make it being ignored.


HichamELBSI (approved):
LGTM

PR #231, opened 2021-08-06
Attempt to fix Intl based date

mfrachet:
Make sure to run the tests in en (even when the computer locales is in another language)


ronronscelestes (approved):
Nice!

PR #232, opened 2021-08-09
Refactoring the layouts

mfrachet:
## Description

When building the different layouts component, I'm feeling that I have taken a wrong pass by augmenting the number of props of a layout and reducing their composable nature.

To give a bit more context, I've taken an approach that I like Nextjs. When using such a tool, you have a /pages folder. Every file in this folder becomes a page at some point in the flow. To create consistency between pages (there are several approaches, I'm existing one that I like), we can create layouts, like the OneBlockLayout that we have:

```jsx
// let's say it's index.js
<OneBlockLayout
  header={<Header />
  startActions={<StartActions />}
  endActions={<EndActions />}
  sideNav={<SideNav />}
>
  Content appearing in a big white box padded and styles
</OneBlockLayout>
```

As I mentioned, this works because every single file in the /pages folder becomes a page.

The problem in our scenario, is that we don't have an FS-based routing system. Instead of that, we manually define a router in a specific file.

```jsx
 <Switch>
  <Route path="/" component={IndexPage} />
</Switch>
```
Let's now try to imagine a scenario where we want to use the OneBlockLayout and its heading prop using such a router.

```jsx
<OneBlockLayout
  header={/* What should I put here? */}>
  startActions={<StartActions />}
  endActions={<EndActions />}
  sideNav={<SideNav />}
>
   <Switch>
     <Route path="/" component={IndexPage} />
  </Switch>
</OneBlockLayout>
```

As you may see, we can't (directly) use the content of the current page in the header since the current page will be defined deeper in the tree (and is eventually not even rendered when we want to display the header component)

We have multiple solutions in this case:
- Shift down the layout to every single page defined in the router, meaning that every page will be wrapped with a OneBlockLayout
- We can connect the Header to the redux store and update it when rendering the pages
- Or we can cut down the layout component into smaller chunks and compose the page accordingly

I've decided to take the last approach since it's more composable and it won't break the actual way Strapi is built. In order to do so, I've created smaller components doing things independently:

```jsx
<Layout sideNav={<SideNav />}>
   <HeaderLayout {...headerLayoutProps} />
   <ActionLayout startActions={<StartActions />} endActions={<EndActions />}/>
   <ContentLayout
      <p>Put content here</p>
   </ContentLayout>
</Layout>
```

## Demo

Example on : https://design-system-git-refactoring-layouts-strapijs.vercel.app/?path=/story/design-system-layouts-layout--sidenav


HichamELBSI (approved):
Very nice !! Good job @mfrachet

PR #233, opened 2021-08-09
Wysiwyg nav layout && Preview mode

ronronscelestes:
## What
- New nav layout with popover (will be fixed when popover PR is merged)
- Preview mode : 
=> When clicked all other action buttons are disabled except for preview button
=> Editor/Codemirror is read only (cannot be edited)
=> Preview (markdown-it reader) appear 
=> Preview should come on top of Editor/Codemirror but still need to test that it's not accessible first
- Tests for actions in popover and disabled actions updated/created

## Demo 
Capture d’écran 2021-08-10 à 16 54 35

Capture d’écran 2021-08-10 à 16 53 39


maevalienard:
If this was just me, I would answer "yes" to all the questions, so I'll let the other devs answer :)

mfrachet:
> Should we change the popover's default position to be centered in reference of the source element?

I don't think we should change the default position of the Popover. However, what we could do is to pass a "centered" props to the popover and position it accordingly on the screen, only when it's passed 😊

> Should we change the popover's default position to be closer to the source element (4px)?

Nope, the popover is used in other places 😊 . However, you have a `spacingTop` that accepts a spacing (defined in the spaces scale) if you want to grow / reduce the space. The minimum value in this scale is 0 and then it's 4 :D . We will never accept a change on that scale except if it's a group decision 😊 


> Should we do something about the extra pixels right side of the popover when we don't want it to have a scrolling behavior?

It's because of an overflow-y necessary for the select component 😅 . We can try to make tests without this one

maevalienard:
> We will never accept a change on that scale except if it's a group decision 😊

The spacing between an object and the popOver in the design is always 4px :)

ronronscelestes:
> I don't think we should change the default position of the Popover. However, what we could do is to pass a "centered" props to the popover and position it accordingly on the screen, only when it's passed 😊
Sure, should I create a branch for it and start on trying to make it work this way?

> Nope, the popover is used in other places 😊 . However, you have a `spacingTop` that accepts a spacing 
Perfect I just used spacingTop={1} and it works like a charm

> It's because of an overflow-y necessary for the select component 😅 . We can try to make tests without this one
Should we have a prop for this new behaviour too so it could keep its overflow-y "default" behavior? Or do you mean that we should try to erase the scroll y effect completely?

mfrachet:
> Sure, should I create a branch for it and start on trying to make it work this way?

Yup, if you don't have a focus on something else, that would be awesome 😊 

> Should we have a prop for this new behaviour too so it could keep its overflow-y "default" behavior? Or do you mean that we should try to erase the scroll y effect completely?

I think we should try first and see if the overflow still works when used in the Select component (the infinite list one) and then take a decision :)

ronronscelestes:
> Yup, if you don't have a focus on something else, that would be awesome 😊

My focus is the Wysiwyg so, I'm on it 🤗

> I think we should try first and see if the overflow still works when used in the Select component (the infinite list one) and then take a decision :)

Okay I might do a PR only for the centering issue first then will see for the overflow issue when the first one is good for us!

example/src/shared/Wysiwyg/WysiwygNav.js
```diff
@@ -24,16 +25,26 @@ import {
   MainButtons,
   CustomIconButton,
   NavWrapper,
-  MoreButton
+  MoreButton,
+  IconButtonGroupMargin
 } from './WysiwygStyles';
 
-const WysiwygNav = ({ placeholder, onActionClick }) => {
+const WysiwygNav = ({
```
HichamELBSI, line 32: In this component, you do 13 times the condition `isPreviewMode ? true : false`. Also, you don't event need to render the popover on preview mode.

You can do the condition only one time before the render:
```jsx
const WysiwygNav = ({ ..., isPreviewMode }) => {
   const buttonMoreRef = useRef();

   if (isPreviewMode) {
       return (
           <NavWrapper padding={2} background="neutral100">
              <Row justifyContent="space-between">
                <Row>
                  <Select
                    disabled
                    id="selectTitle"
                    placeholder={placeholder}
                    size="S"
                    onChange={(value) => onActionClick(value)}
                  >
                    <Option value="h1">h1</Option>
                    <Option value="h2">h2</Option>
                    <Option value="h3">h3</Option>
                    <Option value="h4">h4</Option>
                    <Option value="h5">h5</Option>
                    <Option value="h6">h6</Option>
                  </Select>
      
                  <MainButtons>
                    <CustomIconButton disabled id="Bold" label="Bold" name="Bold" icon={<Bold />} />
                    <CustomIconButton disabled id="Italic" label="Italic" name="Italic" icon={<Italic />} />
                    <CustomIconButton disabled id="Underline" label="Underline" name="Underline" icon={<Underline />} />
                  </MainButtons>
      
                  <MoreButton disabled id="more" label="more" icon={<More />} />
                </Row>
      
                <Button onClick={() => setIsPreviewMode((prev) => !prev)} variant="tertiary" size="L" id="preview">
                  Markdown mode
                </Button>
              </Row>
            </NavWrapper>
       );
   }

   return (
      // render the editable mode :)
   )
}
```

HichamELBSI (changes requested):

example/src/shared/PreviewWysiwyg/Wrapper.js
```diff
@@ -3,15 +3,20 @@ import styled from 'styled-components';
 /* eslint-disable */
 
 const Wrapper = styled.div`
-  max-height: 555px;
-  min-height: 294px;
+  /* position: absolute;
+  top: 0;
+  width: 100%;
+  height: 100%; */
```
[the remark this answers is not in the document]
ronronscelestes: This one I would like to uncomment it but I left there it to let @mfrachet check that the editor isn't tableable when on preview mode!

example/src/shared/PreviewWysiwyg/Wrapper.js
```diff
@@ -3,15 +3,20 @@ import styled from 'styled-components';
...
   overflow: auto;
-  padding: 20px 20px 0 20px;
+  padding: 12px 17px 0 17px;
```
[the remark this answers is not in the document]
ronronscelestes: Oh right, on it!

example/src/shared/Wysiwyg/Editor.js
```diff
@@ -3,29 +3,46 @@ import PropTypes from 'prop-types';
 import CodeMirror from 'codemirror';
 
 import EditorWrapper from './EditorWrapper';
+import { EditorAndPreviewWrapper } from './WysiwygStyles';
+import PreviewWysiwyg from './../PreviewWysiwyg/index'
```
HichamELBSI, line 7: You don't need the `/index` (node.js will search and load (by default) an index.js file from the folder).
You don't need the `./` (When you import a module from a higher directory)

```js
import PreviewWysiwyg from '../PreviewWysiwyg';
```

example/src/shared/PreviewWysiwyg/Wrapper.js
```diff
@@ -3,15 +3,18 @@ import styled from 'styled-components';
 /* eslint-disable */
```
soupette, line 3: I think you can remove this override.

example/src/shared/Wysiwyg/WysiwygNav.js
```diff
@@ -24,55 +25,107 @@ import {
...
-              <CustomIconButton onClick={() => onActionClick("alt")} id="Image" label="Image" name="Image" icon={<Image />} />
-              <CustomIconButton onClick={() => onActionClick("Link")} id="Link" label="Link" name="Link" icon={<Link />} />
-              <CustomIconButton onClick={() => onActionClick("Quote")} id="Quote" label="Quote" name="Quote" icon={<Quote />} />
-            </IconButtonGroup>
+              <MoreButton disabled ref={buttonMoreRef} id="more" label="more" icon={<More />} />
+            </Row>
 
-            <MoreButton onClick={() => console.log('more')} id="more" label="more" icon={<More />} />
+            <Button onClick={() => setIsPreviewMode(prev => !prev)} variant='tertiary' size='L' id='preview'>
+              Markdown mode
```
soupette, line 66: Can you add a comment so we remember to put formatMessage when migrating to code in strapi?

example/src/shared/Wysiwyg/WysiwygNav.js
```diff
@@ -24,55 +25,107 @@ import {
...
+            </Popover>
+          )}
 
-          <Button variant='tertiary' size='L'>Preview mode</Button>
         </Row>
-      </NavWrapper>
-    )
+
+        <Button onClick={() => setIsPreviewMode(prev => !prev)} variant='tertiary' size='L' id='preview'>
+          Preview mode
```
soupette, line 115: Same here

example/src/shared/Wysiwyg/_test_/Wysiwyg.spec.js
```diff
@@ -14,7 +14,7 @@ document.createRange = () => {
 };
 window.focus = jest.fn();
 
-describe("Wysiwyg", () => {
+describe("Wysiwyg render and actions buttons", () => {
```
soupette, line 17: Good job!

soupette (changes requested):
I have tested it and it works great well done!

In another PR can you insert the `code block` and `quote` markups on a new line instead of the current one?

soupette (approved):
LGTM!

PR #234, opened 2021-08-09
Popover centered

ronronscelestes:
## What/Why
- Added centered props to popover
- Added centered behavior when centered props
- Added test to popover with centered props
- Added fix if popover overflow left

## Demo
Capture d’écran 2021-08-09 à 15 05 08


mfrachet (approved):
LGTM, thank you 🙏🏻

PR #236, opened 2021-08-09
Fixed centering calculations && popover ref moved

ronronscelestes:
## What/why

- Centering calculations fixed (based on popover width and not source element width)
- Popover ref moved to PopoverWrapper, if not padding and border couldn't be part of positioning calculations


mfrachet (approved):
LGTM

PR #246, opened 2021-08-11
Add outline none to main component

mfrachet:
When navigating to the a specific page using the keyboard, instead of leaving the keyboard on the selected link, it would great to send the focus to the main element of the new page to facilitate people navigation.

While the Main component has already a tabindex="-1" (meaning focusable programmatically), it has an outline and it's weird to see it on the screen. 

So as server-side rendered page, I would like to try, on page transition, to send the focus on the Main element of the page so that it's announced 😊 . There's no outline for server-rendered page, so I suggest that there's no outline neither on client side navigation 😊

### DIFF
--- a/packages/strapi-parts/src/Main/Main.js
+++ b/packages/strapi-parts/src/Main/Main.js
@@ -1,8 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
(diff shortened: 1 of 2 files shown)


ronronscelestes (approved):
👌🔥

HichamELBSI (approved):

PR #248, opened 2021-08-11
Replaced OneBlockLayout with Layout component

ronronscelestes:
## What/why

Replaced OneBlockLayout with Layout component in parts/example

Capture d’écran 2021-08-11 à 16 27 19


mfrachet (approved):
you can remove every props of the `<Layout>` component :)

mfrachet (approved):

PR #250, opened 2021-08-12
Fix SVG size for TH and not TD

mfrachet:
The SVG size applied for the filters on the TH elements was also applied to the TD elements... Actions were a bit soo small in tables 😓


ronronscelestes (approved):
Nice, I was planning to talk about it with you during our pairprog session 🤓

PR #251, opened 2021-08-12
Doc Accordion

maevalienard:
Signed-off-by: Maeva Lienard <maeva.lienard@strapi.io>

## Description
Hello, I added some doc for the Accordion compo. 

## Demo

### DIFF
--- a/packages/strapi-parts/src/Accordion/Accordion.stories.mdx
+++ b/packages/strapi-parts/src/Accordion/Accordion.stories.mdx
@@ -2,6 +2,7 @@
 
 import { useState } from 'react';
 import { Meta, Story, Canvas } from '@storybook/addon-docs/blocks';
+import { ArgsTable } from '@storybook/addon-docs';
 import EditIcon from '@strapi/icons/EditIcon';
 import { Accordion } from './Accordion';
 import { AccordionToggle } from './AccordionToggle';
@@ -19,7 +20,17 @@ This is the doc of the `Accordion` component
 
 ## Accordion
 
-Description...
+Accordions are collapsable blocks that vertically hide a stack of content. Clicking on the block reveals the associated content.
+
+**Best practices**
+
+- Use accordion when you need big and/or related chunks of content to be organized in a same place.
+- Use accordion for lower priority information.
+- Use to shorten pages and increase readability.
+- The chevron icon pointing down indicates a collapsed state and pointing up indicates an expanded state.
+- Hover and active states should be used anytime it's possible.
+- Content should not scroll within an accordion.
+- Use chevron icon on the left side if call-to-actions are available on the right side.
 
 <Canvas>
   <Story name="base">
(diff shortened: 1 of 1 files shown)


oiorain:
We are still missing examples of code with the various props so we will add them. 
For instance a version with `expanded` and `toggle` at `true` / `false`. 
Do you see any other case?

mfrachet:
I have a story with the Accordion in an expanded variant coming. Also, you can play with the accordion itself in the story provided 😊 . I will try to reorganize the stories when I have time to get something consistent with more examples :D

packages/strapi-parts/src/Accordion/Accordion.stories.mdx
```diff
@@ -19,7 +20,17 @@ This is the doc of the `Accordion` component
...
-Description...
+Accordions are collapsable blocks that vertically hide a stack of content. Clicking on the block reveals the associated content. 
+
+**Best practices**
+
+- Use accordion when you need big and/or related chunks of content to be organized in a same place.
+- Use accordion for lower priority information.
+- Use to shorten pages and increase readability.
+- The chevron icon pointing down indicates a collapsed state and pointing up indicates an expanded state.
+- Hover and active states should be used anytime it's possible.
```
mfrachet, line 31: this is handled under the hood, people won't have to manage this

packages/strapi-parts/src/Accordion/Accordion.stories.mdx
```diff
@@ -19,7 +20,17 @@ This is the doc of the `Accordion` component
...
+- Content should not scroll within an accordion.
```
mfrachet, line 32: We don't allow this in the accordion. People can however tweak the content themselve (using an itermediate HTML element with a fixed height and an overflow). But by default, no scrollbar. Should we mention this line?

packages/strapi-parts/src/Accordion/Accordion.stories.mdx
```diff
@@ -83,7 +94,7 @@ Description...
 
 ## Accordion Group
 
-Description...
+The accordion component has two main states: collapsed and expanded. By default it has a collapsed state. If Call-to-actions are required, use them on the right side and  use chevron icon on the left one.
```
mfrachet, line 97: The accordion group story aims to show that we can wrap a set of accordion with a KeyboardNavigable component to make them keyboard navigable :)

packages/strapi-parts/src/Accordion/Accordion.stories.mdx
maevalienard: Don't you think it's preferable to add this anyway so users can understand the good practices?

packages/strapi-parts/src/Accordion/Accordion.stories.mdx
maevalienard: No because we don't want users to add a scrollbar, hence this line :D

packages/strapi-parts/src/Accordion/Accordion.stories.mdx
maevalienard: Is that something you'd like me to add, or a correction?

packages/strapi-parts/src/Accordion/Accordion.stories.mdx
mfrachet: We can leave it, but the user won't have any freedom in dealing with it, since it's part of the component 😊

HichamELBSI (approved):
LGTM

PR #256, opened 2021-08-12
Carousel documentation

maevalienard:
## What
Carousel documentation (usage + props). 

## Demo

### DIFF
--- a/packages/strapi-parts/src/Carousel/Carousel.stories.mdx
+++ b/packages/strapi-parts/src/Carousel/Carousel.stories.mdx
@@ -1,6 +1,7 @@
 <!--- Carousel.stories.mdx --->
 
 import { Meta, Story, Canvas } from '@storybook/addon-docs/blocks';
+import { ArgsTable } from '@storybook/addon-docs';
 import { useState } from 'react';
 import EditIcon from '@strapi/icons/EditIcon';
 import PublishIcon from '@strapi/icons/PublishIcon';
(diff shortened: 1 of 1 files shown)


packages/strapi-parts/src/Carousel/Carousel.stories.mdx
```diff
@@ -21,7 +22,20 @@ This is the doc of the `Carousel` component
 
 ## Carousel
 
-Description...
+Carousels are slideshow components for navigating through a set of assets. 
+
+**Best practices**
+
+- Carousels should be mostly used in forms.
```
mfrachet, line 29: I'm not sure to understand why? Carousel aims to provide a visual information, it's not an input 🤔

maevalienard: It's not an input, but it's a field and Fields are made to be part of forms :) (in the content manager)

HichamELBSI: This is true for our current use case of these components, but should it be a "best practice" since this is used to display multiple/single assets?

HichamELBSI: I mean, do we want to restrict the use of this compo in a form only?

packages/strapi-parts/src/Carousel/Carousel.stories.mdx
```diff
@@ -21,7 +22,20 @@ This is the doc of the `Carousel` component
...
+- Carousels can also be used to display only one asset.
+- The end of a Carousel label should indicate the amount of included assets.
+- A set of actions is possible via Icon Buttons within a Carousel.
+- Carousels should at least give the possibility to remove an asset from it.
```
HichamELBSI, line 33: Same here, I don't really understand why we need to force a use case on the CTAs of that component?

HichamELBSI (changes requested):

packages/strapi-parts/src/Carousel/Carousel.stories.mdx
maevalienard: For now, I think so. This component has been built to only work within a form (i.e. the main layout we have almost everywhere in Strapi), and don't know how it would react elsewhere. To keep our user experience consistent and relevant, I'd personally recommend this, until we get to know what specific usages the users would require :)

packages/strapi-parts/src/Carousel/Carousel.stories.mdx
maevalienard: The Carousel component is the one we call "Media field" within the CTB, it's different from the Card component that we use within the Media Library, hence the proposal. However, this statement is not always true and I agree with that (permissions, ...) :)

HichamELBSI (approved):
LGTM

PR #257, opened 2021-08-13
Improving stories and adding more axe tests to variants

mfrachet:
- Moving "default" stories to "base" story everywhere
- Adding e2e test about accessibility on variants of the component (and not only the main one)
- Splitting some stories by usage instead of putting dynamic interactions in it


HichamELBSI (approved):
LGTM

PR #258, opened 2021-08-13
Add dynamic header layout

bulby97:
# Dynamic header layout

## Description
Adds a container to display the headerlayout in sticky / non-sticky mode depending on the scroll position of the document

## Demo
https://user-images.githubusercontent.com/3509412/129327747-83fec94d-6897-4fe4-ac12-0334901ba54a.mp4

### DIFF
--- a/packages/strapi-parts/src/Layout/HeaderLayout.js
+++ b/packages/strapi-parts/src/Layout/HeaderLayout.js
@@ -79,6 +79,7 @@ export const BaseHeaderLayout = React.forwardRef(
           paddingBottom={3}
           background="neutral0"
           width={width}
+          data-strapi-header-sticky
         >
           <Row justifyContent="space-between">
             <Row>
@@ -107,6 +108,7 @@ export const BaseHeaderLayout = React.forwardRef(
         paddingBottom={10}
         paddingTop={navigationAction ? 6 : 10}
         background="neutral100"
+        data-strapi-header
       >
         {navigationAction ? <Box paddingBottom={3}>{navigationAction}</Box> : null}
         <Row justifyContent="space-between">
--- a/packages/strapi-parts/src/Layout/HeaderLayout.js
+++ b/packages/strapi-parts/src/Layout/HeaderLayout.js
@@ -1,60 +1,104 @@
-import React from 'react';
+import React, { useState, useRef, useEffect } from 'react';
+import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { H1, Subtitle, P, H2 } from '../Text';
 import { Box } from '../Box';
 import { Row } from '../Row';
 
-export const HeaderLayout = ({
-  navigationAction,
-  primaryAction,
-  secondaryAction,
-  subtitle,
-  title,
-  sticky,
-  ...props
-}) => {
-  if (sticky) {
-    return (
-      <Box paddingLeft={6} paddingRight={6} paddingTop={3} paddingBottom={3}>
-        <Row justifyContent="space-between">
-          <Row>
-            <Box paddingRight={3}>{navigationAction}</Box>
-            <Box>
-              <H2 as="h1" {...props}>
-                {title}
-              </H2>
-              <P small={true} textColor="neutral600">
-                {subtitle}
-              </P>
-            </Box>
-          </Row>
-          <Row>
-            {secondaryAction}
-            {primaryAction ? <Box paddingLeft={2}>{primaryAction}</Box> : undefined}
-          </Row>
-        </Row>
-      </Box>
-    );
-  }
+const useElementOnScreen = (options) => {
+  const containerRef = useRef(null);
+  const [isVisible, setIsVisible] = useState(true);
+
+  const cb = ([entry]) => {
+    setIsVisible(entry.isIntersecting);
+  };
+
+  useEffect(() => {
+    const observer = new IntersectionObserver(cb, options);
+    if (containerRef.current) observer.observe(containerRef.current);
+
+    return () => {
+      if (containerRef.current) observer.unobserve(containerRef.current);
+    };
+  }, [containerRef, options]);
+
+  return [containerRef, isVisible];
+};
+
+export const HeaderLayout = (props) => {
+  const [containerRef, isVisible] = useElementOnScreen({
+    root: null,
+    rootMargin: '0px',
+    threshold: 0,
+  });
 
   return (
-    <Box paddingLeft={10} paddingRight={10} paddingBottom={10} paddingTop={navigationAction ? 6 : 10}>
-      {navigationAction ? <Box paddingBottom={3}>{navigationAction}</Box> : null}
-      <Row justifyContent="space-between">
-        <Row>
-          <H1 {...props}>{title}</H1>
-          {secondaryAction ? <Box paddingLeft={4}>{secondaryAction}</Box> : null}
-        </Row>
-        {primaryAction}
-      </Row>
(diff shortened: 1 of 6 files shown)


packages/strapi-parts/src/Layout/HeaderLayout.js
```diff
@@ -1,72 +1,131 @@
...
+  };
+
+  useEffect(() => {
+    const observer = new IntersectionObserver(cb, options);
+    if (containerRef.current) observer.observe(containerRef.current);
+
+    return () => {
+      if (containerRef.current) observer.unobserve(containerRef.current);
+    };
+  }, [containerRef, options]);
```
mfrachet, line 23: We already have a `useIntersection` hook in `./helpers`, do you think it can be adjusted for this use case?

bulby97:
@mfrachet I just noticed the issue with the `Go back` button. What do you think about simply letting the API exposes the `to` prop of the Link comp ? So that we can handle in the component the `Go Back` not visible in sticky mode ?

Also maybe I'm too used to libs like chakra-ui, but I have the urge to do stuff like 
`<Box position='sticky' top={0} paddingLeft={6} paddingRight={6} paddingTop={3} paddingBottom={3} background="neutral0">` ^^'

packages/strapi-parts/src/Layout/HeaderLayout.js
```diff
@@ -1,72 +1,131 @@
...
+  const [containerRef, isVisible] = useElementOnScreen({
+    root: null,
+    rootMargin: '0px',
+    threshold: 0,
+  });
+
+  return (
+    <>
+      <BaseHeaderLayout ref={containerRef} {...props} />
+      {!isVisible && <BaseHeaderLayout {...props} sticky />}
```
mfrachet, line 38: Do you think it would be possible to hide the `BaseHeaderLayout` that is not sticky, somehow to avoid the following scenario when reverse tabbing?





Because it also duplicate the entries for assistive technologies 😅

mfrachet: Also, what if the HeaderLayout is an h1? I think it would duplicate it right? It's not possible to do that in semantic HTML :(

mfrachet (changes requested):
That looks amazing, thank you for this 🙏🏻 . I'm having one concern about assistive technologies and the presence of both the header in the document when scrolling in it.

And I tiny suggestion one using our existing `useIntersection` hook (if possible 😄 )

Also, could you please write some E2E tests to check the behaviour on webkit, firefox and chrome?

And lastly, thank you again, this looks very promising and amazing 🙏🏻

mfrachet:
Concerning

> @mfrachet I just noticed the issue with the Go back button. What do you think about simply letting the API exposes the to prop of the Link comp ? So that we can handle in the component the Go Back not visible in sticky mode ?

The main constraint that I see is that we need to inject I18N labels and it's easier from the outside by composing (instead of adding new props). Personal opinion but this is the reason why it has been built that way 😊

mfrachet:
Concerning

> Also maybe I'm too used to libs like chakra-ui, but I have the urge to do stuff like
<Box position='sticky' top={0} paddingLeft={6} paddingRight={6} paddingTop={3} paddingBottom={3} background="neutral0"> ^^'

I don't have a strong opinion, if you feel the need for something like this, we can totally enhance the Box component 😊

packages/strapi-parts/src/Layout/HeaderLayout.js
mfrachet: Up on this one 😊

packages/strapi-parts/src/Layout/HeaderLayout.js
```diff
@@ -1,72 +1,163 @@
...
+  top: 0;
+  left: 0;
+  width: ${(props) => props.width}px;
+`;
+
+export const BaseHeaderLayout = React.forwardRef(
+  ({ navigationAction, primaryAction, secondaryAction, subtitle, title, sticky, width, ...props }, ref) => {
+    if (sticky) {
+      return (
+        <StickyBox
```
mfrachet, line 75: Since it's a forwarded ref, maybe you could pass the ref to this box also? 😊

packages/strapi-parts/src/Layout/__tests__/HeaderLayout.e2e.js
```diff
@@ -3,7 +3,7 @@ import { injectAxe, checkA11y } from 'axe-playwright';
 describe('HeaderLayout', () => {
   describe('base', () => {
     it('triggers axe on the document', async () => {
-      await page.goto('http://localhost:6006/iframe.html?id=design-system-layouts-gridlayout--base&viewMode=story');
+      await page.goto('http://localhost:6006/iframe.html?id=design-system-layouts-headerlayout--base&viewMode=story');
```
mfrachet, line 6: Nice catch 😅

packages/strapi-parts/src/Layout/__tests__/HeaderLayout.e2e.js
```diff
@@ -22,10 +22,36 @@ describe('HeaderLayout', () => {
...
+  describe('combined w/ scroll', () => {
+    it('triggers axe on the document', async () => {
+      await page.goto(
+        'http://localhost:6006/iframe.html?id=design-system-layouts-headerlayout--combined-w-scroll&args=&viewMode=story',
+      );
+      await injectAxe(page);
+      await checkA11y(page);
+    });
+
+    it('displays the sticky header when scrolling', async () => {
```
mfrachet, line 41: I think it could be interesting to create the opposite of this test: make sure that the "base" header layout is well displayed when scrolling in the other sense. For now, we're only checking that the sticky is shown when scrolling down, but what happens when scrolling up? 😄

mfrachet (commented):
A bunch of few comments, but I like the way this component is going to plaaaaay 🚀

packages/strapi-parts/src/Layout/HeaderLayout.js
bulby97: The observer needs to check for the static component only; not the sticky one. If needed, we could update the component to support custom refs at some point.

packages/strapi-parts/src/Layout/HeaderLayout.js
```diff
@@ -1,72 +1,163 @@
...
-  title,
-  sticky,
-  ...props
-}) => {
-  if (sticky) {
+const useElementOnScreen = (options) => {
+  const containerRef = useRef(null);
+  const [isVisible, setIsVisible] = useState(true);
+
+  const cb = ([entry]) => {
```
HichamELBSI, line 12: Can you rename the `cb` function, to something more meaningful please 😅

HichamELBSI (changes requested):

mfrachet (approved):
LGTM, amazing job 🚀

maevalienard:
@mfrachet why did we decide to keep the "go back" label instead of just having the arrow already? :)

mfrachet:
@maevalienard it's a lack of wrapping the "Go back" with a VisuallyHidden component. Not important :)

maevalienard:
So that would be possible to change it? :)

HichamELBSI:
> So that would be possible to change it? :)

It is changed : https://design-system-git-feature-dynamic-header-layout-strapijs.vercel.app/?path=/story/design-system-layouts-headerlayout--sticky 😅

HichamELBSI (approved):
LGTM Good job !!

PR #259, opened 2021-08-13
Documentation Avatar 

maevalienard:
## Description
Documentation Avatar & props

## Demo

### DIFF
--- a/packages/strapi-parts/src/Avatar/Avatar.stories.mdx
+++ b/packages/strapi-parts/src/Avatar/Avatar.stories.mdx
@@ -1,6 +1,7 @@
 <!--- Avatar.stories.mdx --->
 
 import { Meta, Story, Canvas } from '@storybook/addon-docs/blocks';
+import { ArgsTable } from '@storybook/addon-docs';
 import { Avatar } from './Avatar';
 import { Box } from '../Box';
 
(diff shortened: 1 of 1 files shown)


packages/strapi-parts/src/Avatar/Avatar.stories.mdx
```diff
@@ -12,7 +13,17 @@ This is the doc of the `Avatar` component
...
 ## Avatar
 
-Description...
+Avatars are used to show a thumbnail of an asset or an individual. 
+
+**Best practices**
+
+- Use the thumbnail preview feature only when the avatar might not be enough.
+- Use avatar in small areas such as tables.
+- In a form we should rather use the Card component.
```
mfrachet, line 22: I'm not sure this is a valid call, a form is semantic. And a card has nothing to do with a form 😬

mfrachet: I'm not sure to understand the final point 😛 . Does it mean that a Card should be inside a form, like a form input of some kind?

maevalienard: Actually I think I made a mistake between Card and Carousel components 😅

HichamELBSI (approved):
LGTM

PR #261, opened 2021-08-13
DatePicker documentation

maevalienard:
## Description
DatePicker doc + props

## Demo


packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -12,7 +13,21 @@ This is the doc of the `DatePicker` component
...
 ## DatePicker
 
-Description...
+Date pickers help users selecting a specific date from a calendar. They can be associated to a Time picker component.  
+
+**Best practices**
+
+- When the date picker is clicked, it takes today's date.
+- It is not possible to write the date from the keyboard.
+- It is not possible to change the date format.
```
mfrachet, line 22: It's not possible to change the date format but the date format will be based on the browser locale. So it may differ if you use a french or an english setting

maevalienard: That's very good to know! Thanks

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -12,7 +13,21 @@ This is the doc of the `DatePicker` component
 
 ## DatePicker
 
-Description...
+Date pickers help users selecting a specific date from a calendar. They can be associated to a Time picker component.
```
HichamELBSI, line 16: Has a decision been made on this use case? (date + time picker)

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -12,7 +13,21 @@ This is the doc of the `DatePicker` component
 
 ## DatePicker
 
-Description...
+Date pickers help users selecting a specific date from a calendar. They can be associated to a Time picker component.  
+
+**Best practices**
+
+- When the date picker is clicked, it takes today's date.
```
HichamELBSI, line 20: Not sure to understand that first point. What do you mean by "it takes" ?

HichamELBSI: Is it also the case when another value has already been selected?

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -12,7 +13,21 @@ This is the doc of the `DatePicker` component
 
 ## DatePicker
 
-Description...
+Date pickers help users selecting a specific date from a calendar. They can be associated to a Time picker component.  
+
+**Best practices**
```
HichamELBSI, line 18: The "best practice" section mainly describes the behavior of the DatePicker. For me, a "best practice" section should mostly tell me in which case I should use a component and why. WDYT?

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -12,7 +13,21 @@ This is the doc of the `DatePicker` component
...
+
+- When the date picker is clicked, it takes today's date.
+- It is not possible to write the date from the keyboard.
+- It is not possible to change the date format. However, the date format will be based on the browser locale. 
+- A Date picker can used simultaneously with a Time picker in a form.
```
HichamELBSI, line 23: I'm not sure this is available ? @mfrachet

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -12,7 +13,21 @@ This is the doc of the `DatePicker` component
...
+- The PopOver closes after selecting a date.
+- One single format of date is currently available.
+- The Date picker is built on top on a Field component.
+
+## Usage
+
+A date picker is used to select an unique date. It's not possible to select a date range. A date picker can have a null value via the Clear feature. A navigation via the keyboard is possible within the calendar PopOver, however it's impossible to write a date from the keyboard.
```
HichamELBSI, line 30: I think we will need to dissociate the way we handle the data (range selection & clear data etc) and the Accessibility part (keyboard navigation). I also feel like it duplicates the best practice section because it also describes the behavior of the component.

HichamELBSI (changes requested):

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
maevalienard: About what did we have to make a decision? :)

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
maevalienard: - I probably should have said something like "it selects" or "it shows" or "it displays"? Does it sound better to you?
- Indeed, it's not the case. Once a date has been selected, this one is displayed and not _today's_ date.

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
maevalienard: Both works to me, at least for a first iteration. If you have any suggestions to add anything to this section, do not hesitate to let me know :)

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
maevalienard: You are right :) We can take a lot once the Accessibility section is made, what do you think?

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
mfrachet: This has not been spec-ed, and thus, not been implemented. For now, if you want to use both, you have to use a DatePicker and a TimePicker

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -12,7 +13,21 @@ This is the doc of the `DatePicker` component
...
+
+**Best practices**
+
+- When the date picker is clicked, it takes today's date.
+- It is not possible to write the date from the keyboard.
+- It is not possible to change the date format. However, the date format will be based on the browser locale. 
+- A Date picker can used simultaneously with a Time picker in a form.
+- The PopOver closes after selecting a date.
+- One single format of date is currently available.
+- The Date picker is built on top on a Field component.
```
mfrachet, line 26: It provides no direct value to the user

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -12,7 +13,21 @@ This is the doc of the `DatePicker` component
...
+Date pickers help users selecting a specific date from a calendar. They can be associated to a Time picker component.  
+
+**Best practices**
+
+- When the date picker is clicked, it takes today's date.
+- It is not possible to write the date from the keyboard.
+- It is not possible to change the date format. However, the date format will be based on the browser locale. 
+- A Date picker can used simultaneously with a Time picker in a form.
+- The PopOver closes after selecting a date.
+- One single format of date is currently available.
```
mfrachet, line 25: Technically, it's not true: an american person and a french person will see different formatting on the screen

mfrachet (changes requested):
Some tiny comments here and there :)

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
maevalienard: We'll have to then, find a way to achieve that as it's part of the fields possibilities in the CTB :) Happy to discuss it

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
maevalienard: Does it depend on the browser locale?

mfrachet: yup

maevalienard: Okay, when I commit the changes, I'll add the detail then :D Thanks!

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
JabStrapi: Would it be possible to activate the keyboard?

mfrachet: With a bigger effort, yes

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
mfrachet: WYT of saying "activated" instead of "clicked"?

maevalienard: Sounds good to me!

packages/strapi-parts/src/DatePicker/DatePicker.stories.mdx
JabStrapi: Thanks for your answer Marvin. How bigger should the effort be? The current working in V3 enables only the date update through the keyboard, not the initial setup. Sorry for that but I think that this will be a requirement (at least for the time & datetime compos, hence my suggestion to make it possible for the date too) since we can't prevent the Content Manager users to select a time that would be different from 00, 15, 30, 45. Could we seize the "datetime" compo refacto slot to do that?

mfrachet: Multiple days to make something consistent, working, accessible and tested I would say

JabStrapi: Thanks for your answer Marvin. We'll schedule those days then.

PR #262, opened 2021-08-13
A better release command

mfrachet:
Release parts in one command

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "test:e2e:ci": "lerna run test:e2e:ci",
     "test:e2e": "lerna run test:e2e --stream",
     "test": "lerna run test --stream",
-    "release": "lerna version --no-push --no-git-tag-version && yarn setup && lerna publish --contents dist"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "test:e2e:ci": "lerna run test:e2e:ci",
     "test:e2e": "lerna run test:e2e --stream",
     "test": "lerna run test --stream",
-    "release": "lerna publish --no-git-tag-version --contents dist"
+    "release": "lerna version --no-push --no-git-tag-version && yarn setup && git commit -am \"Releasing\" && lerna publish from-package --contents dist"
```
mfrachet, line 26: ℹ️ Lerna refuses to publish when there are uncomitted changes. By default, when `--no-git-tag-version` is not provided, the command commits AND adds the tag to the repo. Since we don't want this for now, I'm manually committing a "Releasing" message. Later on, we can remove this commit and defer to lerna version 😄 .

I've run this specific to deploy alpha 10, 5mns ago :)

PR #265, opened 2021-08-16
Docs field

maevalienard:
## Description
Docs field component + props.

## Demo


mfrachet (changes requested):
I don't think this one should be documented. It's an internal component and I don't really think it would be useful for people to know how it works 🤔 .

maevalienard:
As discussed, let's keep this one simple and make a better documentation on the TextInput one :)

mfrachet (approved):

PR #266, opened 2021-08-16
Docs link & loader

maevalienard:
## Description
Docs link & loader + Props for each

## Demo


packages/strapi-parts/src/Link/Link.stories.mdx
```diff
@@ -10,11 +11,24 @@ import { Stack } from '../Stack';
...
 
-This is the doc of the `Link` component
+Links are used to send users to another page or an anchor. They are used for external navigation. 
 
-## Link
+**Best practices**
 
-Description...
+- Use TextButton component if the action is related to internal navigation.
+- Clicking on a Link opens a new tab.
```
mfrachet, line 19: This is not always true. The "go back button" is a Link and it does not open a new tab 😊

packages/strapi-parts/src/Link/Link.stories.mdx
```diff
@@ -10,11 +11,24 @@ import { Stack } from '../Stack';
...
 # Link
 
-This is the doc of the `Link` component
+Links are used to send users to another page or an anchor. They are used for external navigation. 
 
-## Link
+**Best practices**
 
-Description...
+- Use TextButton component if the action is related to internal navigation.
```
mfrachet, line 18: Nope, a button is not a link

packages/strapi-parts/src/Link/Link.stories.mdx
```diff
@@ -10,11 +11,24 @@ import { Stack } from '../Stack';
...
+- Clicking on a Link opens a new tab.
+- Links should always be used with an icon. If it is an external link, use the ExternalLink icon accordingly.
+- Links text should describe an action and start with a verb.
```
mfrachet, line 21: they should describe the destination of the link and not start with a verb actually.

For instance, saying "Go back" does not provide detail about the navigation. What if the person was accessing the page using the URL? Does go back bring them to google?

My suggestion would be "Homepage", which is meaningful in more situations

packages/strapi-parts/src/Link/Link.stories.mdx
```diff
@@ -10,11 +11,24 @@ import { Stack } from '../Stack';
...
+- Links text should be concise and describe their purpose. Users need to know exactly what they are about to do.
+- Use capital letters.
```
mfrachet, line 23: Nope, if we want capital letters, we'll do it on the styling side, not in the code

packages/strapi-parts/src/Link/Link.stories.mdx
```diff
@@ -10,11 +11,24 @@ import { Stack } from '../Stack';
...
+
+## Usage
+
+Links are used for external navigation. Each link should send to another page in a new tab. Moreover, they should be associated with an icon to indicate its external navigation purpose.
```
mfrachet, line 27: See my first comment :)

packages/strapi-parts/src/Loader/Loader.stories.mdx
```diff
@@ -10,14 +11,27 @@ import { Loader } from './Loader';
...
-This is the doc of the `Loader` component
+Loaders indicate some data is currently loading where content usually takes place. They help to notify users that loading is ongoing.
 
-## Loader
+**Best** **practices**
 
-Description...
+- Once the content appears, the loader disappears.
+- If the trigger only concerns a portion of a page, the loader should be placed in that part.
+- Loaders should be scaled down if they aimed to be placed in a small portion of a page.
```
mfrachet, line 20: We don't have a spec for it nor a side to adjust. We can't "scale it on demand" :p

packages/strapi-parts/src/Loader/Loader.stories.mdx
```diff
@@ -10,14 +11,27 @@ import { Loader } from './Loader';
...
+- Use a loader whenever the wait time is to be longer than 1 second.
```
mfrachet, line 21: This will be tough until React 18 is out. Meaning replacing every single way strapi is getting data from the server (and that's a lot)

packages/strapi-parts/src/Loader/Loader.stories.mdx
```diff
@@ -10,14 +11,27 @@ import { Loader } from './Loader';
...
+
+
+## Usage
+
+Even if loaders give an indication that an action has been processed, they do not provide details about what is happening on the back-end side.
```
mfrachet, line 26: I personally think this is not a valid rule. As a user, I want to know that my locales are resolving (and not my permissions somehow). Loading should be contextualized, to me

mfrachet (changes requested):
A link is not a button 😝

packages/strapi-parts/src/Link/Link.stories.mdx
HichamELBSI: In the new admin views, sometimes, there is no capital letters 

Screenshot 2021-08-17 at 09 55 12

HichamELBSI (changes requested):

packages/strapi-parts/src/Link/Link.stories.mdx
maevalienard: Well this is a mistake coming from the design side then. In an ideal world, this should be like mentioned in the documentation

packages/strapi-parts/src/Link/Link.stories.mdx
maevalienard: You're right. Maybe I could rework the sentence by saying something like "Use TextButton component if an action needs to be triggered within the admin panel"?

packages/strapi-parts/src/Link/Link.stories.mdx
maevalienard: That's another topic we could discuss later on 😅 I'll remove this sentence

packages/strapi-parts/src/Link/Link.stories.mdx
maevalienard: Hicham: The idea is to have consistency, so if it's not capital letter here, this is a mistake coming from the design side 😅 
Marvin: We don't do it yet on the styling side?

packages/strapi-parts/src/Loader/Loader.stories.mdx
maevalienard: Alright, I thought something "easy" to do, let's remove it then

packages/strapi-parts/src/Loader/Loader.stories.mdx
maevalienard: Let's remove it for now then :)

packages/strapi-parts/src/Loader/Loader.stories.mdx
maevalienard: But it's not the case, right?

packages/strapi-parts/src/Link/Link.stories.mdx
mfrachet: I'm not sure to understand what you mean

packages/strapi-parts/src/Link/Link.stories.mdx
mfrachet: Yes, you can take the example of opening a modal I think 😊

packages/strapi-parts/src/Link/Link.stories.mdx
mfrachet: What I mean is that if we want to transform the text to be in uppercase, it won't be the job of the developer to type things in uppercase. Imagine a blind user navigating and be yelled on the face saying "LOG IN VIA STRAPI YOU COWARD" 😛 .

When done on the design side (I mean, on the CSS part of the thing, automatically, without a developer action), the text will be announced less agressivel :).

TL;DR => people just write "Log in via Strapi" and we transform the text for them on the according component

packages/strapi-parts/src/Loader/Loader.stories.mdx
mfrachet: And I won't bring the consistency argument with you 😝

packages/strapi-parts/src/Loader/Loader.stories.mdx
mfrachet: It's the case, but it's hidden. In this scenario, people using assistive technologies have more feedbacks than people using a screen monitor 😊 . But as mentioned, that's a personal take. But still, a label should be provided, even if hidden, for people using ATs

packages/strapi-parts/src/Link/Link.stories.mdx
maevalienard: @mfrachet If you look at the UI documentation, you'll see that the Link component is supposed to have a Capital letter style. I think Hicham is not talking about the button, but the "Forgot your password" below. I mentioned that if it's not capital letter, then it's a design mistake :) (if the semantic is fine, which is think so? 🤔 )

packages/strapi-parts/src/Loader/Loader.stories.mdx
maevalienard: @mfrachet Let's talk about this directly, so we can make a decision quickly :)

HichamELBSI: @maevalienard Has a decision been made?

maevalienard: Not yet 😂 Let me bother you instead :p

HichamELBSI (approved):
LGTM thanks

PR #269, opened 2021-08-17
Enhance row and stack components

bulby97:
# Enhance row and stack components

## Description
- Row component now inherits from Box
- Stack component is now flex column by default, to avoid wrapping inline child component with Box
- Fix flex horizontal css prop


packages/strapi-parts/src/Stack/Stack.js
```diff
@@ -26,7 +28,7 @@ const StackH = styled.div.withConfig({
   shouldForwardProp: (prop, defPropValFN) => !transientProps[prop] && defPropValFN(prop),
 })`
   display: flex;
-  flex-direction: horizontal;
```
mfrachet, line 29: oh my... 🤦🏻

mfrachet (approved):
Thank you 🙏🏻

PR #270, opened 2021-08-17
Aspecific data attribute on FieldError and zIndex

mfrachet:
- `data-strapi-field-error` exist to easily target the form element that is in error. For now, when using a "Select" component in core, it's not possible to focus it since our implementation of the `<Form>` component relies on the "name" HTML attribute that is not set on the `<button>` element
 
- set a zIndex on ModalLayout because it goes under the left main nav and subnav. Also put a zIndex on the Tooltip so that it can go over the ModalLayout (for the closing button)


HichamELBSI (approved):
LGTM good job !

PR #273, opened 2021-08-17
Table padding left && right

ronronscelestes:
## What
Updated table padding to 24px instead of 12px

## Why
To be aligned with DS mockup

## Demo
Capture d’écran 2021-08-17 à 19 17 32


mfrachet (approved):
Thank youuu 🙏🏻

PR #274, opened 2021-08-18
Fix leak on observer not disconnecting

mfrachet:
The observer was not disconnecting properly because the reference to the `.current` value does not exist anymore when cleaning the effect. The fix consists of capturing the `.current` variable

Before:

After:


bulby97 (approved):
Nice catch !

PR #275, opened 2021-08-18
Add a loading state to the button

mfrachet:
## Description

Add a loading state to the `<Button>` component

## Demo
Example on : https://design-system-git-loading-button-strapijs.vercel.app/?path=/story/design-system-molecules-button--icons

## API

Adding a `loading` prop that take precedence over the `startIcon`

```diff
<Button
+ loading
>
</Button>
```

### DIFF
--- a/packages/strapi-parts/src/Button/Button.js
+++ b/packages/strapi-parts/src/Button/Button.js
@@ -50,14 +50,23 @@ export const ButtonWrapper = styled(BaseButton)`
 
 export const Button = React.forwardRef(
   ({ variant, startIcon, endIcon, disabled, children, onClick, size, loading, ...props }, ref) => {
+    const isDisabled = disabled || loading;
+
     const handleClick = (e) => {
-      if (!disabled && onClick) {
+      if (!isDisabled && onClick) {
         onClick(e);
       }
     };
 
     return (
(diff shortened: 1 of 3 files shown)


maevalienard (changes requested):
It seems like there's a small misalignment between the loader and the button label

Capture d’écran 2021-08-18 à 10 41 01

mfrachet:
cc @maevalienard I just pushed a fix, let me know if it looks okay

packages/strapi-parts/src/Button/Button.js
```diff
@@ -45,9 +58,15 @@ export const Button = React.forwardRef(
 
     return (
       <ButtonWrapper ref={ref} aria-disabled={disabled} size={size} variant={variant} onClick={handleClick} {...props}>
-        {startIcon && (
+        {(startIcon || loading) && (
```
HichamELBSI, line 61: Shouldn't it be disabled during the loading state?

mfrachet: I agree, I think @maevalienard will be helpful since on the design, the loading state is shown on a primary variant (which is not the disabled one)

mfrachet: (I've pushed the disabled thing 😊 )

HichamELBSI (approved):
LGTM

maevalienard:
Looks good, thanks for the fix!

PR #277, opened 2021-08-18
Add a semantic link that looks like a button

mfrachet:
## Demo
Example on : https://design-system-git-link-button-strapijs.vercel.app/?path=/story/design-system-molecules-linkbutton--base

## API

Same as the Button one except that it accepts a `href` or a `to` prop. `href` is for external usages


ronronscelestes (approved):
Nice, LGTM!

HichamELBSI (approved):
Good job !!

PR #279, opened 2021-08-18
Th's icon size only forced in ActionWrapper

ronronscelestes:
## What

height: ${4 / 16}rem => only for ActionWrapper 

## Why

To prevent every icon in Th to be forced to this size

## Demo
Capture d’écran 2021-08-18 à 15 16 24

### DIFF
--- a/packages/strapi-parts/src/Table/Cell.js
+++ b/packages/strapi-parts/src/Table/Cell.js
@@ -18,7 +18,7 @@ const CellWrapper = styled(RawTd)`
   }
 `;
 
-const ActionWrapper = styled(CellWrapper)`
+const ActionWrapper = styled.span`
   svg {
     height: ${4 / 16}rem;
   }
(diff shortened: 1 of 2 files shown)


packages/strapi-parts/src/Table/Cell.js
```diff
@@ -18,20 +18,20 @@ const CellWrapper = styled(RawTd)`
   }
 `;
 
-const ThWrapper = styled(CellWrapper)`
+const ActionWrapper = styled(CellWrapper)`
```
mfrachet, line 21: `ActionWrapper` should not be a `Cellwrapper` but a `span` 😊

mfrachet (changes requested):
Except my comment, looks good 😊

mfrachet (approved):

HichamELBSI (approved):
LGTM

PR #283, opened 2021-08-20
AccordionToggle is of type button

bulby97:
Add type="button" to avoid form submit


HichamELBSI (approved):
Thanks for that 👍🏽  LGTM

PR #284, opened 2021-08-23
Docs - Radio & SearchBar

maevalienard:
## Description
Docs for the Radio button and the SearchBar + props for each

## Demo

### DIFF
--- a/packages/strapi-parts/src/Radio/Radio.stories.mdx
+++ b/packages/strapi-parts/src/Radio/Radio.stories.mdx
@@ -14,15 +14,15 @@ Radio buttons allow users to choose a single value from a range of options.
 **Best practices**
 
 - When a user picks an option, the previous one is automatically deselected.
-- Radio buttons should be used when there is a least two options.
-- Content of the Radio button should be concise and straightforward.
-- Use the Checkbox component if the user could select one or several value.
+- Radio buttons should be used when there is at least two options.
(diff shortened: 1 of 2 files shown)


packages/strapi-parts/src/Radio/Radio.stories.mdx
```diff
@@ -35,3 +49,8 @@ Description...
     }}
   </Story>
 </Canvas>
+
+## Props
+
+<ArgsTable of={Radio} />
```
soupette, line 55: We might need to specify the propTypes for the Radio component like the `name` and the `onChange`.

packages/strapi-parts/src/Searchbar/Searchbar.stories.mdx
```diff
@@ -64,3 +81,9 @@ The Searchbar disabled
     }}
   </Story>
 </Canvas>
+
+
+## Props
+
+<ArgsTable of={Searchbar} />
```
soupette, line 88: Same here about the propTypes.

packages/strapi-parts/src/Radio/Radio.stories.mdx
soupette: Also it could be great if we could show how we import the Radio component:

```
import { Radio } from '@strapi/parts/Radio';
```

PR #286, opened 2021-08-23
Docs - Status & SubNav

maevalienard:
## Description
Docs - Status & SubNav and props for each

## Demo


soupette (approved):
LGTM!

Can you add the imports so we can merge it ?

PR #287, opened 2021-08-23
Docs - MainNavigation & Switch

maevalienard:
## Description
MainNavigation & Switch docs and props for each. 

## Demo


soupette (approved):
LGTM!

Can you add the imports as well?

PR #290, opened 2021-08-23
ConfirmationDialog component

ronronscelestes:


packages/strapi-parts/src/Dialog/Dialog.js
```diff
@@ -0,0 +1,65 @@
...
+import { H2 } from '../Text';
+import { FocusTrap } from '../FocusTrap';
+import { Portal } from '../Portal';
+
+const DialogWrapper = styled.div`
+  position: absolute;
+  z-index: 3;
+  inset: 0;
+  // theme.colors.neutral800 with opacity
+  background: rgba(50, 50, 77, 0.2);
```
bulby97, line 14: We can have 8 bit hexa colors with opacity.

You can use a small helper to add opacity to the Hexadecimal code.
`const setOpacity = (hex, alpha) => ``${hex}${Math.floor(alpha * 255).toString(16).padStart(2, 0)}``;`

packages/strapi-parts/src/Dialog/Dialog.js
```diff
@@ -0,0 +1,65 @@
...
+`;
+
+const DialogContainer = styled(Box)`
+  max-width: ${412 / 16}rem;
+  margin: 0 auto;
+  overflow: hidden;
+  margin-top: 10%;
+`;
+
+const DialogHeader = styled(Box)`
```
bulby97, line 25: Maybe we can use Row instead ? It has the display:flex

packages/strapi-parts/src/Dialog/Dialog.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+This is the doc of the `Dialog` component
+
+## Dialog
+
+Description...
+
+<Canvas>
+  <Story name="base">
+    <Dialog onClose={() => {}} title="Confirmation" labelledBy="confirmation">
```
bulby97, line 27: It would be better if we could trigger the modal from a button in the stories. Right now the modal prevents you from accessing the story

bulby97 (changes requested):
Good job !

bulby97 (approved):

HichamELBSI (changes requested):
Good job @ronronscelestes

PR #295, opened 2021-08-26
Docs - Tooltip & Timepicker

oiorain:
# Tooltip & Time picker docs

## Description
Docs for the Tooltip and Time Picker components with props and imports

note: the tooltip `canvas` demo doesn't seems to work correctly though, there's a bunch of text. 

Screenshot 2021-08-26 at 16 40 29

Edit: here's my proposition for documenting the different properties.
I'm not sure the Smaller example actually displays the tooltip as smaller.
I'm propose to use real buttons to make it more obvious that you can have a cta.

Screenshot 2021-08-27 at 19 03 49

## Demo
Example on :  See message below 👇


maevalienard:
Very nice proposal!

soupette (approved):
LGTM!

PR #302, opened 2021-08-30
Dialog isOpen props

ronronscelestes:
## What/Why

Let Dialog component deal with being open or not with a prop
Updated useEffect overflow and tests

## Demo

Capture d’écran 2021-08-30 à 14 21 40

### DIFF
--- a/packages/strapi-parts/src/Dialog/Dialog.js
+++ b/packages/strapi-parts/src/Dialog/Dialog.js
@@ -27,15 +27,22 @@ const DialogHeader = styled(Row)`
   border-bottom: 1px solid ${({ theme }) => theme.colors.neutral150};
 `;
 
-export const Dialog = ({ onClose, labelledBy, title, describedBy, ...props }) => {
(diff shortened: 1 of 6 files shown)


packages/strapi-parts/src/Dialog/Dialog.js
```diff
@@ -27,15 +27,22 @@ const DialogHeader = styled(Row)`
   border-bottom: 1px solid ${({ theme }) => theme.colors.neutral150};
 `;
 
-export const Dialog = ({ onClose, labelledBy, title, describedBy, ...props }) => {
+export const Dialog = ({ onClose, labelledBy, title, describedBy, isOpen, ...props }) => {
   //FIX ME (find a way to do it globally)
   useEffect(() => {
-    const body = document.body;
-    body.classList.add('modal-open');
+    const body = document?.body;
```
bulby97, line 33: In what case body does not exist ?

bulby97 (approved):

PR #311, opened 2021-09-02
Fix and improve components

HichamELBSI:
Fix all feedback on the in progress column : https://www.notion.so/strapi/f15d5f0f7cfb4c34ab802408c0f41ea7?v=a611848256ca4488ab4d9a02a03de5fe

### DIFF
--- a/packages/strapi-parts/src/Text/index.js
+++ b/packages/strapi-parts/src/Text/index.js
@@ -2,12 +2,21 @@ import React from 'react';
 import styled from 'styled-components';
 
 const handleColor = ({ theme, textColor }) => theme.colors[textColor || 'neutral800'];
+const ellipsisStyle = ({ ellipsis }) =>
+  ellipsis &&
+  `
+    display: inline-block;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  `;
 
 export const H1 = styled.h1`
   font-weight: 600;
   font-size: ${32 / 16}rem;
   line-height: 1.25;
   color: ${handleColor};
+  ${ellipsisStyle}
 `;
 
 export const H2 = styled.h2`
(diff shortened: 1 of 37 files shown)


soupette (approved):
LGTM

ronronscelestes (approved):
LGTM 🔥

packages/strapi-parts/src/Text/index.js
```diff
@@ -49,14 +49,21 @@ export const Subtitle = styled(Text)`
...
 
-export const TableLabel = styled(TextButton)`
+export const TableLabel = styled(ButtonText)`
   font-weight: 600;
   font-size: ${11 / 16}rem;
   line-height: 1.45;
   text-transform: uppercase;
 `;
+
+export const EllipsisText = styled(Text)`
```
bulby97, line 64: WDYT about having ellipsis as a component and not a prop ?
Some might use a TableLabel, ButtonText, or any other Text component with an ellipsis prop on it

bulby97 (changes requested):
Good job ;)

packages/strapi-parts/src/Text/index.js
HichamELBSI: I understand that but the problem is that the Text component is a span. To use the ellipsis style, we need a `block` element not an `inline`. I'm afraid that moving the text component to a `block` will break a lot of things in the other components, layouts or migrated views. @bulby97

bulby97: Got your point, but ellipsis can also be applied to a inline-block. I don't think the inline block can break inline stuff, but maybe I'm wrong

HichamELBSI: You are right, I can add the ellipsis to the inline-block my bad.

PR #313, opened 2021-09-06
Wysiwyg fix TextButton export

ronronscelestes:
## What / Why

App crashed with the Wysiwyg export =>
Replace ButtonText component with TextButton

### DIFF
--- a/example/src/shared/Wysiwyg/index.js
+++ b/example/src/shared/Wysiwyg/index.js
@@ -5,7 +5,7 @@ import WysiwygFooter from './WysiwygFooter';
 import WysiwygExpand from './WysiwygExpand';
 import MediaLibrary from './MediaLibrary';
 import Editor from './Editor';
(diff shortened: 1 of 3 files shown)


example/src/shared/Wysiwyg/index.js
```diff
@@ -5,7 +5,7 @@ import WysiwygFooter from './WysiwygFooter';
 import WysiwygExpand from './WysiwygExpand';
 import MediaLibrary from './MediaLibrary';
 import Editor from './Editor';
-import { ButtonText } from '@strapi/parts/Text';
+import { TextButton } from '@strapi/parts/Text';
```
mfrachet, line 8: The component that is exported from Text is ButtonText, not TextButton :). If it does not work, make sure to yarn setup at the project run, and reinstall the example dependencies. 😊

mfrachet (changes requested):
It's the other way 😛

PR #314, opened 2021-09-07
Live regions and notify

mfrachet:
# Add tools to notify for changes

## Description

This PR introduces the `useNotifyAT()` hook that aims to provide functions to notify Assistive technologies from content being loaded.

For instance, when fetching data from a server, people using ATs don't receive feedbacks when the request has finished, by default. With this new hook, we can provide them with information when the data has finished loaded so that they know they can navigate to the loaded content 😊 

## Demo
Example on : 

## API
[api details]

```jsx
const { notifyAlert, notifyStatus, notifyLog } = useNotifyAT();

// later in the component
notifyAlert('Hello moto')
notifyStatus('Hello moto')
notifyLog('Hello moto')
```

## Voiceover

Add live regions


soupette (approved):
LGTM

PR #320, opened 2021-09-12
Binding main and h1, implicitly

mfrachet:
Make main labelled by the default H1 id. main label and H1 id can both be overriden if necessary

Instead of:

```jsx
<Main labelledBy="title">
  <H1 id="title"></H1>
</Main>
```

just do

```jsx
<Main>
  <H1></H1>
</Main>
```


soupette (approved):
LGTM

PR #321, opened 2021-09-13
Minor fixes on component props and outline

mfrachet:
- Box can now have a children of string
- Outline is well offsetted on input elements
- ToggleInput receives the destructured name attribute
- ActionLayout wrap


soupette (approved):
LGTM

PR #324, opened 2021-09-14
CardAsset with a children

mfrachet:
## Description

Add a children to `CardAsset` to pass icons instead of images when necessary

```jsx
<CardAsset>
    <Icon />
</CardAsset>
```

or 

```jsx
<CardAsset src="url" />
```


soupette (approved):
LGTM

PR #325, opened 2021-09-14
Add combobox

bulby97:
From 
https://www.24a11y.com/2019/select-your-poison/
https://www.24a11y.com/2019/select-your-poison-part-2/

image

### DIFF
--- a/packages/strapi-parts/src/Combobox/Combobox.js
+++ b/packages/strapi-parts/src/Combobox/Combobox.js
@@ -1,6 +1,5 @@
-import React, { useEffect, useState, useRef, useLayoutEffect } from 'react';
+import React, { useEffect, useState, useRef, useLayoutEffect, Children, cloneElement } from 'react';
 import PropTypes from 'prop-types';
-import styled from 'styled-components';
 import { useId } from '../helpers/useId';
 import DropdownIcon from '@strapi/icons/FilterDropdownIcon';
 import {
@@ -17,76 +16,33 @@ import { CaretBox } from '../Select/components';
 import { Popover } from '../Popover';
 import { Box } from '../Box';
 import { Text } from '../Text';
-
-const MainRow = styled(Row)`
-  position: relative;
-  border: 1px solid ${({ theme, hasError }) => (hasError ? theme.colors.danger600 : theme.colors.neutral200)};
-  padding-right: ${({ theme }) => theme.spaces[3]};
-  padding-left: ${({ theme }) => theme.spaces[3]};
-  border-radius: ${({ theme }) => theme.borderRadius};
-  background: ${({ theme }) => theme.colors.neutral0};
-
-  ${({ theme, disabled }) =>
-    disabled
-      ? `
-    color: ${theme.colors.neutral600};
-    background: ${theme.colors.neutral150};
-  `
-      : undefined}
-
-  &:focus-within {
-    border: 1px solid ${({ theme }) => theme.colors.primary600};
-  }
-`;
-
-const Input = styled.input`
-  min-height: ${40 / 16}rem;
-  border: none;
-  flex: 1;
-  font-size: ${14 / 16}rem;
-  color: ${({ theme }) => theme.colors.neutral800};
-  &:focus-visible {
-    outline: none;
-    box-shadow: none;
-    outline-offset: 0;
-  }
-`;
-
-const OptionBox = styled(Box)`
-  width: 100%;
-  border: none;
-  text-align: left;
-  outline-offset: -3px;
-  ${({ isSelected, theme }) => isSelected && `background: ${theme.colors.primary100};`}
-
-  &:hover {
-    background: ${({ theme }) => theme.colors.primary100};
-  }
-`;
+import { Loader } from '../Loader/Loader';
+import { Input, MainRow, OptionBox } from './components';
 
 export const Combobox = ({
   createMessage,
   disabled,
   label,
-  options,
   value,
   onChange,
   placeholder,
-  isCreatable,
+  creatable,
+  loading,
   onCreateOption,
   onLoadMore,
   noOptionsMessage,
   hasMoreItems,
+  children: nodes,
 }) => {
   const [activeIndex, setActiveIndex] = useState(0);
   const [selectedIndex, setSelectedIndex] = useState(null);
-  const [filteredOptions, setFilteredOptions] = useState(options);
+  const [filteredNodes, setFilteredNodes] = useState(nodes);
   const [open, setOpen] = useState(false);
-  const [inputValue, setInputValue] = useState(value ? options.find((v) => v.value === value)?.name : '');
+  const [inputValue, setInputValue] = useState(value ? nodes.find((v) => v.value === value)?.name : '');
 
   useEffect(() => {
-    setFilteredOptions(filterOptions(options, inputValue));
-  }, [inputValue, options]);
+    setFilteredNodes(filterOptions(nodes, inputValue));
+  }, [inputValue, nodes]);
 
   const firstUpdate = useRef(true);
   useLayoutEffect(() => {
@@ -95,13 +51,13 @@ export const Combobox = ({
       return;
     }
 
-    const index = filteredOptions.findIndex((v) => v.value === value);
+    const index = filteredNodes.findIndex((node) => node.props.value === value);
     if (index !== -1) {
-      const selected = filteredOptions[index];
-      setInputValue(selected.name);
+      const selected = filteredNodes[index];
+      setInputValue(selected.props.children);
       setActiveIndex(0);
       setSelectedIndex(0);
-      setFilteredOptions(filterOptions(options, inputValue));
+      setFilteredNodes(filterOptions(nodes, inputValue));
     }
   }, [value]);
 
@@ -122,7 +78,7 @@ export const Combobox = ({
 
   const onInput = () => {
     const curValue = inputRef.current.value;
-    setFilteredOptions(filterOptions(options, curValue));
+    setFilteredNodes(filterOptions(nodes, curValue));
     setActiveIndex(0);
     setSelectedIndex(null);
 
@@ -137,7 +93,7 @@ export const Combobox = ({
 
   const onInputKeyDown = (event) => {
     const { key } = event;
-    const max = filteredOptions.length - 1;
+    const max = filteredNodes.length - 1;
     const action = getActionFromKey(key, open);
 
     switch (action) {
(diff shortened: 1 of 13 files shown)


mfrachet:
Let me take a look at this 😊 , thank you!

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,275 @@
...
+  text-align: left;
+  outline-offset: -3px;
+  ${({ isSelected, theme }) => isSelected && `background: ${theme.colors.primary100};`}
+
+  &:hover {
+    background: ${({ theme }) => theme.colors.primary100};
+  }
+`;
+
+export const Combobox = ({ label, options, value, onChange, placeholder }) => {
```
mfrachet, line 68: I think it misses the disabled props here. We don't manage disabled using plain HTML but instead rely on `aria-disabled` and prevent the onChange event from being triggered :)

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,275 @@
+import React, { useEffect, useState, useRef, useLayoutEffect } from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { useId } from '../helpers/useId';
+import DropdownIcon from '@strapi/icons/FilterDropdownIcon';
+import {
+  getActionFromKey,
+  getUpdatedIndex,
+  isScrollable,
+  maintainScrollVisibility,
```
mfrachet, line 10: we use a library to handle these specific scenario, see https://github.com/strapi/parts/blob/develop/packages/strapi-parts/package.json#L51

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,275 @@
...
+import { Text } from '../Text';
+
+const MainRow = styled(Row)`
+  position: relative;
+  border: 1px solid ${({ theme, hasError }) => (hasError ? theme.colors.danger600 : theme.colors.neutral200)};
+  padding-right: ${({ theme }) => theme.spaces[3]};
+  padding-left: ${({ theme }) => theme.spaces[3]};
+  border-radius: ${({ theme }) => theme.borderRadius};
+  background: ${({ theme }) => theme.colors.neutral0};
+  overflow: hidden;
```
mfrachet, line 28: Why is this one needed?

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,275 @@
...
+  }
+`;
+
+const Input = styled.input`
+  min-height: ${40 / 16}rem;
+  border: none;
+  flex: 1;
+  font-size: ${14 / 16}rem;
+  color: ${({ theme }) => theme.colors.neutral800};
+  &:focus-visible {
```
mfrachet, line 49: I think this is handled by default no? (in our global styles)

packages/strapi-parts/src/Combobox/utils.js
```diff
@@ -0,0 +1,126 @@
+export const Keys = {
```
mfrachet, line 1: There's already a list of Keyboard keys in the utils folder, could you please rely on these? 🙏🏻

mfrachet (changes requested):
The code is very, very good, thank you for this contribution :) .

I'm having one "major" concern: we try to rely on more declarative APIs using JSX for composition. Do you think it could be possible to refactor a little bit the code in order to make it more like the Select API?

Something like:

```jsx
<Combobox>
   <ComboboxOption>Hey ya</ComboboxOption>
</Combobox>
```

???

The rest is nitpicky stuff :). Very nicely handled 👏🏻

packages/strapi-parts/src/Combobox/Combobox.js
bulby97: c/c of your select component :D

packages/strapi-parts/src/Combobox/Combobox.js
bulby97: Doesn't look like it

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,274 @@
...
+      },
+      onClick: () => onOptionClick(i),
+      onMouseDown: onOptionMouseDown,
+      isSelected: isActive,
+    });
+  });
+
+  return (
+    <>
+      <MainRow ref={containerRef} disabled={disabled}>
```
mfrachet, line 169: The disabled props will spread on the underlying DOM and we don't really want that 😓

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,274 @@
...
+            ) : creatable ? (
+              <OptionBox
+                paddingLeft={4}
+                paddingRight={4}
+                paddingTop={2}
+                paddingBottom={2}
+                ref={activeOptionRef}
+                onMouseDown={onOptionMouseDown}
+                role="option"
+                onClick={() => onOptionSelect()}
```
mfrachet, line 225:
```suggestion
                onClick={onOptionSelect}
```

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,274 @@
...
+                <Text textColor="neutral800">{createMessage(inputValue)}</Text>
+              </OptionBox>
+            ) : (
+              <Box paddingLeft={4} paddingRight={4} paddingTop={2} paddingBottom={2} ref={activeOptionRef}>
+                <Text textColor="neutral800">{noOptionsMessage(inputValue)}</Text>
+              </Box>
+            )}
+            {loading && (
+              <Row justifyContent="center" alignItems="center" paddingTop={2} paddingBottom={2}>
+                <Loader small>Loading content...</Loader>
```
mfrachet, line 238: This should be formatted from the outside

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,274 @@
...
+            )}
+          </div>
+        </Popover>
+      )}
+    </>
+  );
+};
+
+Combobox.defaultProps = {
+  createMessage: (value) => `Create "${value}"`,
```
mfrachet, line 249: Aren't this props and the `createable` one redundant? WYT of only using one prop since the message should always be provided when `createable` is true

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,274 @@
...
+  disabled: false,
+  loading: false,
+  hasMoreItems: false,
+  creatable: false,
+  noOptionsMessage: () => 'No results found',
```
mfrachet, line 254: Why a function?

packages/strapi-parts/src/Combobox/Combobox.js
```diff
@@ -0,0 +1,274 @@
...
+          aria-disabled={disabled}
+          readOnly={disabled}
+          ref={inputRef}
+          role="combobox"
+          type="text"
+          value={inputValue}
+          onBlur={!disabled ? onInputBlur : undefined}
+          onClick={!disabled ? () => updateMenuState(true) : undefined}
+          onInput={!disabled ? onInput : undefined}
+          onKeyDown={!disabled ? onInputKeyDown : undefined}
```
mfrachet, line 186: Suggestion reversing the ternary meaning:

```suggestion
          onBlur={disabled ? undefined : onInputBlur}
          onClick={disabled ? undefined : () => updateMenuState(true)}
          onInput={disabled ? undefined : onInput }
          onKeyDown={disabled ? undefined : onInputKeyDown}
```

mfrachet (changes requested):
This is a very nice job 👏🏻 . I'm having some more small comments. Also, on the Figma documentation tab, the label is always displayed. And I feel that we want to have the error and the hint displayed also. Do you think you could manage these case with FieldLabel and FieldError (and Field)

🙏🏻

bulby97:
> The code is very, very good, thank you for this contribution :) .
> 
> I'm having one "major" concern: we try to rely on more declarative APIs using JSX for composition. Do you think it could be possible to refactor a little bit the code in order to make it more like the Select API?
> 
> Something like:
> 
> ```js
> <Combobox>
>    <ComboboxOption>Hey ya</ComboboxOption>
> </Combobox>
> ```
> 
> ???
> 
> The rest is nitpicky stuff :). Very nicely handled 👏🏻



> This is a very nice job 👏🏻 . I'm having some more small comments. Also, on the Figma documentation tab, the label is always displayed. And I feel that we want to have the error and the hint displayed also. Do you think you could manage these case with FieldLabel and FieldError (and Field)
> 
> 🙏🏻

I get your point but in our real use case (RBAC), there is no label, nor error. So what I might suggest is to have this component, and maybe another one with all the fields

packages/strapi-parts/src/Combobox/Combobox.js
bulby97: This fn accept a index param, We don't want to provide the event param here

packages/strapi-parts/src/Combobox/Combobox.js
bulby97: Right, but you could want the component to be creatable without providing a createMessage value.
Or maybe we could do like react-select, by having a <Creatable /> component

packages/strapi-parts/src/Combobox/Combobox.js
bulby97: Inspired by react-select
If "No results for ${value}"

packages/strapi-parts/src/Combobox/Combobox.js
mfrachet: The Creatable component makes sense for compasibility I think, do you feel you could do it that way?

mfrachet:
> @mfrachet I get your point but in one of our real use case (RBAC), there is no label, nor error. So what I might suggest is to have this component, and maybe another one with all the fields

The Figma documentation tab is the source of truth, not the other ones :). Not adding the hint and the error looks okay to me, but the label should be there

bulby97:
My bad, I got confused with the multi-select component ;)

packages/strapi-parts/src/helpers/keyboardKeys.js
```diff
@@ -12,4 +12,6 @@ export const KeyboardKeys = {
   DELETE: 'Delete',
   PAGE_UP: 'PageUp',
   PAGE_DOWN: 'PageDown',
+  BACKSPANCE: 'Backspace',
```
mfrachet, line 15: Small typo, I guess? 😛

PR #328, opened 2021-09-15
Fix Card import

oiorain:
the DS documentation import was wrong so it's fixed 
Screenshot 2021-09-15 at 15 00 11


soupette (approved):
Lgtm

PR #330, opened 2021-09-15
Fix header layout console error

mfrachet:
Biggest fix of all time preventing Box from yelling because children is not undefined


soupette (approved):
Lgtm

PR #353, opened 2021-09-23
Testing caching config on the branch that should pass all tests

ivanThePleasant:
# Dependency caching

## Description
Refactored Actions config and added yarn dependency caching step which caches yarn dependencies on a successful CI/CD actions run. Caches are available to all branches of the repo, so when actions are run next time, they will be used instead of fetching all yarn dependencies from scratch. Cache itself is dependant on the root yarn.lock file, if it changes then restoring cache step will be skipped and on successful actions run yarn dependencies will be re-cached.

## Use
Once PR is merged, actions will be updated for all new PR's


.github/workflows/playwright-ci.yml
```diff
@@ -17,8 +17,26 @@ jobs:
...
+        with:
+          path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: Install yarn dependencies
+        run: yarn global add http-server && yarn
+
+      - name: Install Lerna dependencies
```
HichamELBSI, line 35: Can you rename that step please?
Something like : `Link locale packages and install their dependencies.`

HichamELBSI (changes requested):
Thanks for the help @ivanThePleasant ! 👍🏽

.github/workflows/playwright-ci.yml
ivanThePleasant: Sure, no worries. Just to clarify, which step you would like to be renamed? Since steps need to stay separate, I could name them something like:

`Link locale yarn packages and install their dependencies`
and
`Link locale Lerna packages and install their dependencies`

HichamELBSI: I was talking about the `yarn bootstrap` command. IMO we can just rename the line 35.

HichamELBSI (approved):
LGTM

PR #357, opened 2021-09-27
Checkbox align start

mfrachet:
For Multiline checkbox, the checkmark is now stick to the top of the label instead of at the center


HichamELBSI (approved):
LGTM

PR #360, opened 2021-09-27
Action Layout skrink fix

ronronscelestes:
## What

Added flex-shrink 0 to end actions in action layout to prevent components to skrink


HichamELBSI (approved):
LGTM

PR #363, opened 2021-09-28
Fix modal position & click away to close

HichamELBSI:

### DIFF
--- a/packages/strapi-parts/src/ModalLayout/ModalLayout.js
+++ b/packages/strapi-parts/src/ModalLayout/ModalLayout.js
@@ -1,10 +1,11 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useRef } from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { Box } from '../Box';
 import { FocusTrap } from '../FocusTrap';
 import { Portal } from '../Portal';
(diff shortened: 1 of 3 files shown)


packages/strapi-parts/src/ModalLayout/ModalLayout.js
```diff
@@ -23,22 +23,36 @@ const ModalContent = styled(Box)`
 `;
 
 export const ModalLayout = ({ onClose, labelledBy, ...props }) => {
+  const modalContentRef = useRef();
+
   //FIX ME (find a way to do it globally)
   useEffect(() => {
     const body = document.body;
     body.classList.add('lock-body-scroll');
+    document.addEventListener('mousedown', handleClickAway);
```
mfrachet, line 32: why not to put a onCLick on the overlay directly instead of attaching the event to the whole document?

mfrachet (changes requested):
I put a comment about the click outside thing

packages/strapi-parts/src/ModalLayout/ModalLayout.js
HichamELBSI: Oh you are right !!

ronronscelestes:
Shouldn't we add the same changes to the Dialog component?

mfrachet (approved):
lgtm for now

PR #369, opened 2021-09-30
Popover overflow bottom fix

ronronscelestes:
## What

Fixed popover overflow bottom maths
Popover padding and border (10px) weren't returned by popoverRect.height which caused miscalculations


HichamELBSI (approved):
LGTM

PR #376, opened 2021-10-05
Table border radius

ronronscelestes:
## What

Fixed border-radius for table => table should always have 4px border-radius all around
- If footer then footer deals with bottom border-radius if not table handles it
- Added hasRadius to main table container to prevent weird borders overflowing
- Overflow hidden on TableBox to prevent shadows to overflow border-radius

## Screenshots 

Capture d’écran 2021-10-05 à 16 14 09

Capture d’écran 2021-10-05 à 16 14 16

Capture d’écran 2021-10-05 à 16 12 01

### DIFF
--- a/packages/strapi-parts/src/Table/Table.js
+++ b/packages/strapi-parts/src/Table/Table.js
@@ -11,7 +11,8 @@ const TableWrapper = styled(RawTable)`
 
 const TableBox = styled(Box)`
   position: relative;
-  border-radius: ${({ theme }) => theme.borderRadius} ${({ theme }) => theme.borderRadius} 0 0;
+  border-radius: ${({ theme, footer }) =>
+    footer ? `${theme.borderRadius} ${theme.borderRadius} 0 0` : theme.borderRadius};
 
   &:before {
     // TODO: make sure to add a token for this weird stuff
@@ -46,6 +47,7 @@ const ScrollContainer = styled(Box)`
 export const Table = ({ colCount, rowCount, footer, ...props }) => {
   const tableRef = useRef(null);
   const [overflowing, setOverflowing] = useState();
+  console.log(footer);
 
   const handleScroll = (e) => {
     const maxScrollLeft = e.target.scrollWidth - e.target.clientWidth;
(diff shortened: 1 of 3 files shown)


maevalienard (approved):
Well done Julie! Thank you :D

packages/strapi-parts/src/Table/Table.js
```diff
@@ -46,6 +48,7 @@ const ScrollContainer = styled(Box)`
 export const Table = ({ colCount, rowCount, footer, ...props }) => {
   const tableRef = useRef(null);
   const [overflowing, setOverflowing] = useState();
+  console.log(footer);
```
HichamELBSI, line 51: Can you remove the console.log please

packages/strapi-parts/src/Table/Table.js
```diff
@@ -70,8 +73,8 @@ export const Table = ({ colCount, rowCount, footer, ...props }) => {
   }, []);
 
   return (
-    <Box shadow="tableShadow">
-      <TableBox background="neutral0" overflowing={overflowing}>
+    <Box shadow="tableShadow" hasRadius>
+      <TableBox footer={footer} background="neutral0" overflowing={overflowing}>
```
HichamELBSI, line 77: Can you rename the TableBox props please:

Something like 
```js
fullBorderRadius={footer}
or
withBorderRadius={footer}
or 
hasFullBorderRadius={footer}
🤷🏽‍♂️
```

HichamELBSI (changes requested):

packages/strapi-parts/src/Table/Table.js
ronronscelestes: oh sure will do sorry!

soupette: We should add an eslint rule for this, WDYT?

packages/strapi-parts/src/Table/Table.js
soupette: I'd prefer something that starts by `with` ....

ronronscelestes: let's go for with

mfrachet (changes requested):
I'll take a look tomorrow, I'm really not fan of adding this kind of props

PR #387, opened 2021-10-14
Switch from Jest to Playwright official test runner for End-to-end tests

fdel-car:
# Summary

At first, I wanted to parallelize the execution of our end-to-end tests, but it appears that it was already the case when running tests for multiple files. But since I ended up switching to the official test runner, it could be of interest to you 🙂 

- Uses the official playwright test runner ([`@playwright/test`](https://playwright.dev/docs/intro/)) instead of [`jest-playwright`](https://github.com/playwright-community/jest-playwright). It was recommended to do so everywhere I looked.

image

- Updates every `*.e2e.js` test to be compatible with the new API.
- Removes the step using the [playwright-github-action](https://github.com/microsoft/playwright-github-action) in our GitHub Actions workflow as it's not needed anymore.
- Moves the `injectAxe` calls outside `beforeEach` since it's needed only when executing `checkA11y`. It should trim off a few milliseconds here and there.
If you find it annoying I don't mind reverting it.

## Remarks

- It's faster when running only one test file though because by default playwright is able to parallelize executions between browsers or every test when using [`.parallel`](https://playwright.dev/docs/test-parallel/#parallelize-tests-in-a-single-file) - jest was just parallelizing test files execution, but **it was faster previously** with jest when running all the tests at once.
I think it's clearly possible to improve the performance though, It's likely that when switching to their new API I did some mistakes that would slow tests down. I'm curious so I'll give that a more thorough look.
- I'm still working on making sure the browsers are cached between CI runs.
_Edit_: I'll probably do that in another PR if you are interested in this switch.

Pros:
- Supports TypeScript out of the box, better IDE experience without any setup efforts. I can switch tests to TS in this PR if you are interested.
- Their API is very intuitive, you can have a look on https://playwright.dev/docs/intro
- It's the official test runner, should offer better long-term support, and resources should be more widely available.
- Provides some additional features like visual comparisons (I'm sure they are others but I didn't find an exhaustive list to compare to `jest-playwright`).

Cons:
- For some unknown reasons it's slower (~20%) when running all tests at once.
- No test coverage out of the box.


fdel-car:
They are some merge conflicts, I'll fix them if you find this PR interesting and are interested in merging it soon.
Otherwise, I'll wait until you are ready to review it 🙂

HichamELBSI:
@fdel-car Is this PR still relevant?

fdel-car:
It does not provide much for the parallelization of the e2e tests which was the initial request.

If you can invest a bit of time to review this and you wanna stick to the recommended packages and API, those changes are - I think - headed in the right direction.
I had a chat with @mfrachet who seemed interested back then, so I would suggest going through the documentation of the official test runner I've linked above to make your own opinions - or we can talk about it when you have time.

Having said all of that it's up to you, considering that this PR is a bit old, things evolved rapidly recently and most of the files changed are tests files, it's definitely not a bad idea to close it and start working again on this subject in the future when you feel it's the right time (cause most of the changes in this PR are regex search & replace, which could be done again pretty fast). And I'll happily provide help on the subject if I can 🙂

_Side note_: considering the caching of the browsers, I was planning to make another PR once this one landed and follow the instructions you can find [here](https://playwright.dev/docs/ci#caching-browsers).

HichamELBSI:
Alright! Thanks for the followup @fdel-car. I'll convert this PR to draft for the moment until we can work on it.

HichamELBSI:
Hi @fdel-car, I finally took the time to work on it . Thanks for your initial work 🙏🏽  : https://github.com/strapi/design-system/pull/544

PR #389, opened 2021-10-15
Prevent escape from propagatin

mfrachet:
When using a select or a simple menu in a modal, pressing escape closes the modal instead of closing the dropdown


ronronscelestes (approved):
LGTM!

soupette (approved):
LGTM

PR #392, opened 2021-10-20
Add disabled state to tabs

soupette:

# [title]

## Description
[add description]

## Demo

## API
[api details]

```jsx
<YourComponent />
```

## Voiceover

[voiceover gif]

### DIFF
--- a/packages/strapi-parts/src/Tabs/Tabs.js
+++ b/packages/strapi-parts/src/Tabs/Tabs.js
@@ -1,10 +1,17 @@
 import React, { Children, cloneElement } from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 import { useTabs } from './TabsContext';
 import { ButtonText, TableLabel } from '../Text';
 import { KeyboardKeys } from '../helpers/keyboardKeys';
 import { useTabsFocus } from './useTabsFocus';
 import { DefaultTabsRow, DefaultTabButton, DefaultTabBox, SimpleTabBox } from './components';
+
+const TabButton = styled.button`
+  &[aria-disabled='true'] {
+    cursor: not-allowed;
+  }
+`;
 
 export const Tabs = ({ children, ...props }) => {
   const { id, selectedTabIndex, selectTabIndex, label, variant } = useTabs();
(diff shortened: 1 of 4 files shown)


packages/strapi-parts/src/Tabs/Tabs.js
```diff
@@ -97,25 +108,28 @@ export const Tab = ({ selected, id, children, variant, hasError, onClick, onTabC
...
-      <button
+      <TabButton
         id={tabId}
         role="tab"
         aria-controls={selected ? tabPanelId : undefined}
         tabIndex={selected ? 0 : -1}
         aria-selected={selected}
         type="button"
         onClick={handleClick}
+        aria-disabled={disabled}
```
HichamELBSI, line 126: Don't we need to pass the disabled props also ?

```js
disabled={disabled}
```

packages/strapi-parts/src/Tabs/Tabs.js
```diff
@@ -1,11 +1,18 @@
...
 import { ButtonText, TableLabel } from '../Text';
 import { KeyboardKeys } from '../helpers/keyboardKeys';
 import { useTabsFocus } from './useTabsFocus';
 import { DefaultTabsRow, DefaultTabButton, DefaultTabBox, SimpleTabBox } from './components';
 
+const TabButton = styled.button`
+  &[aria-disabled='true'] {
+    cursor: not-allowed;
+  }
+`;
```
HichamELBSI, line 14: Why do we need this modification ?

HichamELBSI: If we do this here, don't we need to standardize it in every disabled components ? @maevalienard

packages/strapi-parts/src/Tabs/Tabs.js
mfrachet: disabled=disabled is not practical for people using keyboard. When using disabled, the field is not tabbable and so not announced to the users. On the other hand, relying on aria-disabled and preventing the real click allows for the button to be tabbable and announced as "dimmed" by SRs

PR #417, opened 2021-11-02
QA/ Checkbox cursor

ronronscelestes:
## What

QA request
Added cursor pointer to BaseCheckbox and Checkbox components

### DIFF
--- a/packages/strapi-design-system/src/Checkbox/Checkbox.js
+++ b/packages/strapi-design-system/src/Checkbox/Checkbox.js
@@ -12,7 +12,7 @@ const CheckboxLabel = styled(Text)`
   display: flex;
   align-items: flex-start;
   * {
-    cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'default')};
+    cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
   }
 `;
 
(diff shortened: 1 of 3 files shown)


packages/strapi-design-system/src/Checkbox/Checkbox.js
```diff
@@ -12,7 +12,7 @@ const CheckboxLabel = styled(Text)`
   display: flex;
   align-items: flex-start;
   * {
-    cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'default')};
+    cursor: ${({ disabled }) => (disabled ? 'not-allowed' : 'pointer')};
```
HichamELBSI, line 15: About that update, the default cursor on checkboxes is the default behaviour of html checkboxes. IMO, if we do that modification here, we should standardize this modification also in other clickable components (Toggle, etc) @maevalienard

maevalienard: Apart from Checkbox and toggle, what else is impacted? :)

HichamELBSI: Checkbox, Toggle, Radio button (maybe the Select list item too)

maevalienard: That'd make sense to me to have a cursor change on all these :) @HichamELBSI

HichamELBSI: I created a task for the other components on the migration feedback board. @ronronscelestes

PR #419, opened 2021-11-02
QA/ Subnav Header alignment

ronronscelestes:
## What

Fixed alignment between Label and Search button
QA request: when Label takes 2 lines the alignment should not be centered but flex-start

## Snapshots

before
Capture d’écran 2021-11-02 à 18 09 52

after
Capture d’écran 2021-11-02 à 18 10 49


HichamELBSI (approved):
LGTM

PR #421, opened 2021-11-04
Typography

HichamELBSI:
Add variant to Typography 

## Api
```js
//  variants list => 'alpha' | 'beta' | 'delta' | 'epsilon' | 'omega' | 'pi' | 'sigma'

<Typography variant="alpha">Title</Typography>
```


ronronscelestes (approved):
Good job!! 🔥

example/src/CM/EditView/Information.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { Box } from "@strapi/design-system/Box";
 import { Stack } from "@strapi/design-system/Stack";
-import { Text } from "@strapi/design-system/Text";
+import { Typography } from "@strapi/design-system/Typography";
```
soupette, line 4: WDYT about deleting the example app completely we don't need it anymore.

packages/strapi-design-system/src/Accordion/AccordionToggle.js
```diff
@@ -109,19 +110,19 @@ export const AccordionToggle = ({ title, description, as, togglePosition, action
...
             <>
               {size === 'S' ? (
-                <Text bold={true} as={as} id={ariaLabelId} textColor={titleColor}>
+                <Typography fontWeight="bold" as={as} id={ariaLabelId} textColor={titleColor}>
                   {title}
-                </Text>
+                </Typography>
               ) : (
-                <H3 as={as} id={ariaLabelId} textColor={titleColor}>
+                <AccordionTypography variant="delta" as={as} id={ariaLabelId} textColor={titleColor}>
```
soupette, line 117: Just a thought about the naming but wdyt about `DeltaTypography` as name for this component?

example/src/CM/EditView/Information.js
HichamELBSI: Can we do that in another PR ?

packages/strapi-design-system/src/Accordion/AccordionToggle.js
HichamELBSI: This is a specific case to help the accordion to select the right css class. I name it AccordionTypography because it shouldn't be used outside the accordion. If we rename it `<DelaTypography variant="delta" />`, the name wouldn't fit if we change the variant eg `<DeltaTypography variant="beta" />` 🤔

example/src/CM/EditView/Information.js
soupette: Done in #423 can you update your PR?

PR #429, opened 2021-11-08
Accordion error state

ronronscelestes:
## What

Accordion and AccordionGroup error state
If only Accordion, it applies border and error message
If AccordionGroup, AccordionGroup applies error message and Accordion applies border

## Snapshot

Screenshot 2021-11-08 at 16 46 12

Screenshot 2021-11-08 at 16 46 21


HichamELBSI (approved):
LGTM

PR #430, opened 2021-11-08
Accordion overflow // TO MERGE AFTER ERROR STATE

ronronscelestes:
## What

Removed overflow hidden on Accordion causing issues when users add certain type of component inside AccordionContent needing to overflow the parent

Added the border radius with CSS selectors instead


HichamELBSI (approved):
LGTM

PR #432, opened 2021-11-09
Typography ellipsis fix

ronronscelestes:
## What

Fixed Typography ellipsis (needed display block)
Added ellipsis example in Typography stories

### DIFF
--- a/packages/strapi-design-system/src/Select/__tests__/Select.spec.js
+++ b/packages/strapi-design-system/src/Select/__tests__/Select.spec.js
@@ -60,7 +60,7 @@ describe('Select', () => {
 
       .c9 {
         color: #32324d;
-        display: inline-block;
+        display: block;
         white-space: nowrap;
         overflow: hidden;
         text-overflow: ellipsis;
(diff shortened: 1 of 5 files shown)


packages/strapi-design-system/src/Select/__tests__/Select.spec.js
```diff
@@ -60,7 +60,7 @@ describe('Select', () => {
 
       .c9 {
         color: #32324d;
-        display: inline-block;
+        display: block;
```
HichamELBSI, line 63: The ellipsis text shouldn't be a block element IMO

HichamELBSI (changes requested):

packages/strapi-design-system/src/Select/__tests__/Select.spec.js
ronronscelestes: nb: discussed and fixed during peer programming

HichamELBSI (approved):
LGTM

PR #443, opened 2021-11-19
Input alignment required issue

ronronscelestes:
## What

Attempt to fix inputs alignment when required field next to not required field

## Snapshot

before
Screenshot 2021-11-19 at 12 31 33
Screenshot 2021-11-19 at 12 47 22

after
Screenshot 2021-11-19 at 12 39 52


HichamELBSI (approved):
LGTM

PR #445, opened 2021-11-22
NumberInput arrow click issue

ronronscelestes:
## What

Fixed NumberInput not activating onValueChange when clicking on arrows up/down
Added disabled state 
Added tests

## Demo

Click on arrows to change the value
The value should update immediately (not on blur)

### DIFF
--- a/packages/strapi-design-system/src/NumberInput/NumberInput.js
+++ b/packages/strapi-design-system/src/NumberInput/NumberInput.js
@@ -54,23 +54,39 @@ export const NumberInput = React.forwardRef(
       }
     };
 
-    const increment = () => {
+    const increment = (fromKeyBoard) => {
       const parsedValue = numberParserRef.current.parse(inputValue);
 
       if (isNaN(parsedValue)) {
-        setInputValue(numberFormaterRef.current.format(step));
+        if (fromKeyBoard) {
+          setInputValue(numberFormaterRef.current.format(step));
+        } else {
+          onValueChange(step);
+        }
       } else {
-        setInputValue(numberFormaterRef.current.format(parsedValue + step));
+        if (fromKeyBoard) {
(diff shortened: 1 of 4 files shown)


packages/strapi-design-system/src/NumberInput/NumberInput.js
```diff
@@ -124,10 +146,29 @@ export const NumberInput = React.forwardRef(
...
+                <ArrowButton
+                  aria-hidden
+                  reverse
+                  onClick={() => {
+                    // increment needs an argument, so we can't remove the parenthesis
+                    increment();
+                  }}
+                  tabIndex={-1}
+                  type="button"
+                  data-testid="ArrowUp"
```
HichamELBSI, line 158: Don't we need to add the disabled props here? To fix the disabled state

packages/strapi-design-system/src/NumberInput/NumberInput.js
```diff
@@ -124,10 +146,29 @@ export const NumberInput = React.forwardRef(
...
-                <ArrowButton aria-hidden onClick={decrement} tabIndex={-1} type="button">
+                <ArrowButton
+                  aria-hidden
+                  onClick={() => {
+                    // decrement needs an argument, so we can't remove the parenthesis
+                    decrement();
+                  }}
+                  tabIndex={-1}
+                  type="button"
+                  data-testid="ArrowDown"
```
HichamELBSI, line 170: same question

packages/strapi-design-system/src/NumberInput/NumberInput.js
ronronscelestes: you're right good catch, on it!

mfrachet (approved):
go for it when tests pass 👍🏻

PR #447, opened 2021-11-22
Select multi tags / Remove a tag issue

ronronscelestes:
## What

Fixed removing a tag from select multi tags causing bug


HichamELBSI (approved):
LGTM

PR #459, opened 2021-11-29
Fix dependencies

HichamELBSI:
Fix deps


packages/strapi-design-system/package.json
```diff
@@ -9,11 +9,11 @@
     "@babel/preset-env": "^7.13.0",
     "@babel/preset-react": "^7.12.13",
     "@figspec/react": "^0.1.5",
-    "@storybook/addon-actions": "^6.1.20",
-    "@storybook/addon-essentials": "^6.1.20",
-    "@storybook/addon-links": "^6.1.20",
-    "@storybook/addon-storyshots": "^6.3.7",
-    "@storybook/react": "^6.1.20",
+    "@storybook/addon-actions": "^6.4.0",
```
soupette, line 12: What do you think about using a fixed version for each dependency?

mfrachet: I'm not for

PR #471, opened 2021-12-21
Update getYears.js

misterboe:
As it should be possible to use the Date field as "Birthday" -50 is a bit low

### DIFF
--- a/packages/strapi-design-system/src/DatePicker/utils/getYears.js
+++ b/packages/strapi-design-system/src/DatePicker/utils/getYears.js
@@ -6,5 +6,5 @@ export const getYears = () => {
 
   return Array(101)
     .fill(null)
-    .map((_, index) => currentYear - 50 + index);
+    .map((_, index) => currentYear - 125 + index);
 };


packages/strapi-design-system/src/DatePicker/utils/getYears.js
```diff
@@ -6,5 +6,5 @@ export const getYears = () => {
 
   return Array(101)
     .fill(null)
-    .map((_, index) => currentYear - 50 + index);
+    .map((_, index) => currentYear - 125 + index);
```
HichamELBSI, line 9: Hello @misterboe, thanks for your contribution, this modification is not enough to increase the number of years in the DatePicker. The length of the array should be equal to `numberOfYears * 2 + 1` to have the same number of years before and after the current year plus the current year.

Also, I think you can add a `numberOfYears` prop to the DatePicker with a default value to 50 and update all the related tests (snapshots /  e2e / utils test).

The way to do that:
 
```js
// DatePicker/utils/getYears.js

export const getYears = (numberOfYears) => {
  const currentYear = new Date().getFullYear();
  const yearsArray = numberOfYears * 2;

  return Array(yearsArray)
    .fill(null)
    .map((_, index) => currentYear - numberOfYears + index);
};
```

```js
// DatePicker/DatePickerCalendar.js

export const DatePickerCalendar = ({ ..., numberOfYears }) => {
    ...;
    const years = getYears(numberOfYears);
    ....;
};

// The proptype list should be sorted
DatePickerCalendar.propTypes = {
    ...;
    numberOfYears: PropTypes.number.isRequired,
};
```

```js
// DatePicker/DatePicker.js

export const DatePicker = ({ ..., numberOfYears, ...props }) => {
    ...;
    
    <DatePickerCalendar
        ...
        numberOfYears={numberOfYears}
     />

    ...;
}

DatePicker.defaultProps = {
  ...;
  numberOfYears: 50,
  ...;
};

DatePicker.propTypes = {
  ...;
  numberOfYears: PropTypes.number,
  ...;
};
```

To update the snapshot, you can run `yarn test -u`  in `packages/strapi-design-system`

To run the e2e tests, you can run `yarn test:e2e DatePicker` in `packages/strapi-design-system`

HichamELBSI (changes requested):

HichamELBSI:
Hi @misterboe  I'm closing this PR since there is no update. Feel free to open a new one when you are available.

PR #473, opened 2021-12-23
Hide the Autocomplete Dialog when Field is disabled

### LINKED ISSUE #469: A non-editable field should not show autocomplete options
### Describe the bug

A non-editable field should not show autocomplete options.

1. Configue a view with a non-editable field.
2. Click on the field and see the autocomplete options.

### Expected behavior

Non autocomplete options.

### Screenshots

Screenshot 2021-12-17 at 19 15 25

> This issue was previously opened on the main [Strapi repository](https://github.com/strapi/strapi/issues/11940)

nextrapi:
## Issue
# Fix #469 A non-editable field should not show autocomplete options 

image

## Solution

I think the disabled field is not actually being passed down to the input element.

image

The height adjustment also helps with the input color (specially on autocomplete) fitting the outer container:

image

image

### Update
Fixed it by turning autoComplete manually off when the form is disabled, while still allowing it to be overwritten by the user.


HichamELBSI:
Hello @nextrapi,

Thanks for your contribution. The `disabled` prop do work for the autocomplete use case. However, disabling a field makes it invisible when navigating with the keyboard (especially with the tab) and the screen reader (unless you scan the text). For each component modification, we have to think about making it readable and usable for everyone. That's why the disabled props is not passed to the components. So, the `aria-disabled` property allows us to give the information of this state. Concerning auto completion issue, this is a common issue we have here (especially with browser password extensions). I think we can try to set the props `autoComplete="off"` if the field is disabled.

Don't forget to run the test and regenerate the snapshots 😉 

To regenerate the snapshots, you can run `yarn test -u` in `packages/strapi-design-system`;

```js
const FieldInput = ({ ..., autoComplete }) => {
    ...;
    <Input
        ...
        autoComplete={disabled ? "off" : autoComplete}
};

FieldInput.defaultProps = {
  autoComplete: undefined,
  ...,
};

FieldInput.propTypes = {
  autoComplete: PropTypes.string,
  ...,
};
```

packages/strapi-design-system/src/Field/FieldInput.js
```diff
@@ -20,7 +20,7 @@ const Input = styled.input`
   font-size: ${14 / 16}rem;
   display: block;
   width: 100%;
-
+  height: 100%;
```
HichamELBSI, line 23: What was the issue with the height ?

HichamELBSI (changes requested):

packages/strapi-design-system/src/Field/FieldInput.js
nextrapi: If you look at my description and the pictures I attached, the inner input tag's height doesn't fill the outer div container, therefore when you have a field with previous autocomplete fields it gives the input a blue color, but it looks a bit off to me as it is now: 

image

This is the result with height 100%:
image

HichamELBSI: Great! Thanks for that!

HichamELBSI (approved):
Looks good to me! Great work @nextrapi ! Thanks for the help 😄

HichamELBSI:
The snapshots seems to be broken. Can you check all the tests suites please.

At the root:
Run the test : `yarn test`
Run the e2e test : `yarn test:e2e`

HichamELBSI:
I noticed that we had a problem with the test configuration. develop is now up to date, can you update your branch so we can merge it. Sorry for that.

PR #481, opened 2022-01-12
Loader : Move img to svg component (for theme colors usage)

HichamELBSI:


packages/strapi-design-system/src/Loader/LoaderSvg.js
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+
+const loaderSvg = (props) => {
```
ronronscelestes, line 3: Small comment: should we add the L in uppercase for LoaderSvg component?

ronronscelestes (approved):
LGTM!

PR #487, opened 2022-01-24
Required UI for ToggleInput, Textarea and Carousel

ronronscelestes:
## What

Added required UI for
- ToggleInput
- Textarea
- CarouselInput


packages/strapi-design-system/src/Carousel/Carousel.js
```diff
@@ -1,33 +1,8 @@
-import React, { Children, cloneElement, useRef } from 'react';
+import React from 'react';
```
HichamELBSI, line 1: IMO this component should be named CarouselInput as it uses the Field components. The other one is a simple Carousel

packages/strapi-design-system/src/Carousel/CarouselInput.js
```diff
@@ -0,0 +1,138 @@
+import React, { Children, cloneElement, useRef } from 'react';
```
HichamELBSI, line 1: This one should be the Carousel component 😅

HichamELBSI (approved):
Great job @ronronscelestes  🤩

PR #499, opened 2022-02-01
Docs: improve wording of contributing guide

gu-stav:
# Improve wording of the contributing guide

This makes the contribution part a bit more explicit about which parts shouldn't be edited.


HichamELBSI (approved):
Thanks for that @gu-stav !

PR #501, opened 2022-02-02
Ds website

Bassel17:
# Ds website content fix

## Description
- Fix content (remove v1.0)
- Add google analytics tags

## Demo
Example on : https://design-system-website-staging.vercel.app


packages/strapi-design-system/src/Field/FieldInput.js
```diff
@@ -92,6 +92,7 @@ export const FieldInput = forwardRef(({ endAction, startAction, disabled, onChan
         hasLeftAction={Boolean(startAction)}
         hasRightAction={Boolean(endAction)}
         onChange={handleChange}
+        autoComplete={disabled ? 'off' : autoComplete}
```
ronronscelestes, line 95: I would suggest creating another PR for this autoComplete proposal, I'm not sure it has something to do with the purpose of this PR, WDYT?

Out of curiosity, what brought up the need to add this now, is it coming from a sprint task or a Github issue?

website/pages/_document.js
```diff
@@ -9,8 +9,25 @@ class MyDocument extends Document {
   render() {
     return (
       <Html lang="en">
-        <Head />
+        <Head>
+          <script
```
gu-stav, line 13: When we'd like to add GTM/ GA we must also add a way for the user to consent the data collection (via a cookie banner e.g.). Without it would be a violation of the GDPR.

gu-stav (changes requested):

HichamELBSI:
Hi @Bassel17 any news on this PR ?

packages/strapi-design-system/src/Field/FieldInput.js
HichamELBSI: +1 please create another PR for this proposal

HichamELBSI (changes requested):

Bassel17:
Separated things into different pull requests, will close this PR

PR #502, opened 2022-02-03
GridItem: prevent column blowout

gu-stav:
## Description

In it's current state `GridItems` can break out the grid, if their content exceeds their width, as seen in the last column here:

Screenshot from 2022-02-03 10-31-22

Adding a `max-width: 100%` to each column ensures it's size in the grid. With that the columns falls in line with the others:

Screenshot from 2022-02-03 10-35-44


ronronscelestes (approved):
LGTM!

Issue #504, opened 2022-02-04
Layour component as a named export


ivanThePleasant:
In the Design System Storybook `Layout` component is suggested to be imported as a default export:

`import Layout from '@strapi/design-system/Layout';`

However, this component should be imported as a named export instead:

`import { Layout } from '@strapi/design-system/Layout';`

Importing it as a default export break the application. The documentation on Storybook should be changed to import `Layout` component as a named export. 

Steps to reproduce:
1. Generate new plugin in Strapi
2. Add the `Layout` component to plugin's front end importing it with default export syntax (as suggested in the Storybook)

Link to the Storybook page:

https://design-system-alpha.vercel.app/?path=/docs/design-system-components-layout--base

PR #505, opened 2022-02-04
Docs: Replace default export with named imports everywhere

gu-stav:
Some component documentation claimed that components should use default imports. All of them are using named exports and therefore should use named imports too.

Fixes: #504


ronronscelestes (approved):
LGTM!

PR #508, opened 2022-02-09
Card: Properly handle multiple children in CardAction 

gu-stav:
This allows `CardActions`s to contain more than one child. By using the `Stack` component, multiple children will be shown properly.

Refs: https://github.com/strapi/strapi/pull/12428


HichamELBSI (approved):
🔥  Thanks for this modification.

PR #511, opened 2022-02-09
Enable Specify sync

HichamELBSI:
Fix Specify conf and Synchronize tokens + Icons


ronronscelestes (approved):
LGTM!

gu-stav (commented):
Thanks @HichamELBSI . It's quite a huge changeset 👀 I've found these two instances where I'm not sure whether we want that:

Screenshot from 2022-02-10 10-14-38

Screenshot from 2022-02-10 10-14-07

HichamELBSI:
> 
Nice catch @gu-stav 
Fyi, this is a Specify sync, these modifications has been done in Figma.

@maevalienard Do we want that?

gu-stav:
@maevalienard And while you are at it: I do have a remark concerning the wheelchair icon:

Screenshot from 2022-02-10 10-27-42

Many people in wheelchairs don't feel properly represented by that icon, because it shows a very passive person. Therefore a new icon was created. If you like you can [read more about it](https://accessibleicon.org/). I don't know it is being used, but maybe it makes sense to incorporate these thoughts?

maevalienard:
How can this circle icon with the cross be broken like this! 😂 Good catch! I've updated it in Figma. 

Regarding the spacing around the chevron icon, we needed it on Figma side because otherwise everything was broken. We should keep an eye on it indeed. 

-------
@gu-stav: I didn't know about the wheelchair, it completely makes sense. I used the one I found in the icon library that we use. What I suggest is that we keep it as such for now and whenever I have some time, I'll rework that icon myself. What do you think?

ronronscelestes:
> And then, it will break everything in Figma, what am I supposed to do here 😂 I probably won't have the time to address this today, can it wait for next week?

I don't know about you @HichamELBSI, for me it can wait a little, only timing issue we could have is for when I'll have to deal with the dark mode 👀

HichamELBSI:
@maevalienard Can you rename that Icon temporarily on Figma so we can merge? I will keep the previous one manually

maevalienard:
@HichamELBSI Sounds good. I added "ONHOLD" as a prefix to that icon :)

PR #515, opened 2022-02-10
Tag: fix text color

gu-stav:
This PR fixes the text color of the `Tag` component. The proper color was already set on the container, but not passed onto the text.

Disabled
Screen Shot 2022-02-10 at 14 07 40

Enabled
Screen Shot 2022-02-10 at 14 07 30

Refs: https://strapi-inc.atlassian.net/browse/DS-2


HichamELBSI (approved):
LGTM

PR #524, opened 2022-02-15
SimpleMenu: Allow elments to be passed as prop-types

gu-stav:
In https://github.com/strapi/strapi/pull/12213/files#diff-945b0ae1436ead2b4ac56d7921a076ac866efcd77ed5985466444d07b1751d78R30-R56 I'd like to create a custom label for the `SimpleMenu`. It works already, by the prop-types don't allow elements yet. This PR adds `element`s next to `string` and `number` as valid types.

An example was added too.

Refs: https://github.com/strapi/strapi/pull/12213

### DIFF
--- a/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
+++ b/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
@@ -193,5 +193,5 @@ SimpleMenu.propTypes = {
   as: PropTypes.any,
   children: PropTypes.oneOfType([PropTypes.arrayOf(menuItemType), menuItemType]).isRequired,
   id: PropTypes.string,
-  label: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.element]).isRequired,
 };
--- a/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
+++ b/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
@@ -97,12 +97,24 @@ export const SimpleMenu = ({ label, children, id, as: asComp, ...props }) => {
   const Component = asComp || Button;
 
   useEffect(() => {
-    // Useful to focus the selected item in the list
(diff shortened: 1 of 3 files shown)


HichamELBSI:
This modification will break the way we define the default item to focus in the list.

```js
  // SimpleMenu.js 
  useEffect(() => {
    // Useful to focus the selected item in the list
    const defaultItemIndexToFocus = childrenArray.findIndex((c) => c.props.children === label);
    if (defaultItemIndexToFocus !== -1) {
      setFocusItem(defaultItemIndexToFocus);
    }
  }, [label]);
```

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -97,10 +97,18 @@ export const SimpleMenu = ({ label, children, id, as: asComp, ...props }) => {
...
-    if (defaultItemIndexToFocus !== -1) {
-      setFocusItem(defaultItemIndexToFocus);
+    if (['string', 'number'].includes(typeof label)) {
+      // Useful to focus the selected item in the list
+      const defaultItemIndexToFocus = childrenArray.findIndex((c) => c.props.children === label);
+
+      if (defaultItemIndexToFocus !== -1) {
+        setFocusItem(defaultItemIndexToFocus);
+      }
+    } else if (typeof label === 'object') {
```
gu-stav, line 107: @HichamELBSI I'm not sure about the check, but what do you think about this? I feel like this works properly in all cases.

HichamELBSI:
@gu-stav Can you just add a condition to this useEffect to check if label is not an element so we can merge this one 👍🏽

HichamELBSI (approved):
LGTM 👍🏽

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
soupette: Maybe you can check if the object as `$$typeof` key?

HichamELBSI: Maybe you can use `React.isValidElement(label)` instead?

PR #525, opened 2022-02-15
MenuItem: Disable hover styles when aria-disabled is set

gu-stav:
When `aria-disabled` is passed into `MenuItem` hover styles should be disabled. The issue was discovered in https://github.com/strapi/strapi/pull/12213#discussion_r803889633.


HichamELBSI (approved):
Thanks :)

PR #535, opened 2022-02-17
feat(combobox-oninputchange): Add onInputChange prop for Combobox

artkondakov:
# Add onInputChange prop for Combobox

## Description
I noticed that there are no way to subscribe on onChange event to Combobox' input, not all widget.
I wanted to use Combobox with calling API after search query updated, but with no events from it that's impossible.
With this prop you can request options data from endpoint:
* specify onInputChange callback
* get target value from input
* use it as param in your request
* get data from endpoint
* map ComboboxOptions with new items

## API
onInputChange accepts a function

```jsx
<Combobox onInputChange={(e) => console.log(e.target.value)} />
```

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.0.1-alpha.75",
+  "version": "0.0.1-alpha.76",
   "npmClient": "yarn"
 }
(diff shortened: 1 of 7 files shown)


HichamELBSI (approved):
Hi @artkondakov, thanks for the contribution

HichamELBSI (approved):

PR #536, opened 2022-02-17
Move labelAction into the FieldLabel component

gu-stav:
Before this PR each Field component used it's own way to render the action for a label. This PR moves it into the `FieldLabel` component, so that: 1) no component is forgotten 2) it is rendered visually consistent and 3) we are not duplicating the code in multiple places.

Screen Shot 2022-02-17 at 12 08 14

~The positioning of the icon is not great at the moment. It's not perfectly centered, because if interferes with the line-height of the label. It would help to push it down a little. What do you think about it?~

Refs: https://strapi-inc.atlassian.net/browse/CONTENT-131

### DIFF
--- a/packages/strapi-design-system/src/Field/FieldLabel.js
+++ b/packages/strapi-design-system/src/Field/FieldLabel.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
+import { Flex } from '../Flex';
 import { useField } from './FieldContext';
 import { Typography } from '../Typography';
 
@@ -8,7 +9,15 @@ const TypographyAsterisk = styled(Typography)`
   line-height: 0;
 `;
 
-export const FieldLabel = ({ children, required, ...props }) => {
+const Action = styled(Flex)`
+  line-height: 0;
+
+  svg path {
+    fill: ${({ theme }) => theme.colors.neutral500};
+  }
+`;
+
+export const FieldLabel = ({ children, required, action, ...props }) => {
   const { id } = useField();
 
   return (
@@ -21,16 +30,21 @@ export const FieldLabel = ({ children, required, ...props }) => {
       required={required}
       {...props}
     >
-      {children}
-      {required && <TypographyAsterisk textColor="danger600">*</TypographyAsterisk>}
+      <Flex alignItems="center">
+        {children}
+        {required && <TypographyAsterisk textColor="danger600">*</TypographyAsterisk>}
+        {action && <Action marginLeft={1}>{action}</Action>}
+      </Flex>
     </Typography>
   );
 };
 
 FieldLabel.defaultProps = {
   required: false,
+  action: undefined,
 };
 FieldLabel.propTypes = {
+  action: PropTypes.element,
   children: PropTypes.node.isRequired,
   required: PropTypes.bool,
 };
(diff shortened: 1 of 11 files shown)


ronronscelestes:
Nice idea @gu-stav 
I don't know if it'll help but on the UI side, if you replace Action as a Box for Action as a Flex component, it helps to center the Icon with the label / required ⬇️ WDYT?
image

packages/strapi-design-system/src/Field/FieldLabel.js
```diff
@@ -1,14 +1,22 @@
...
+import { Flex } from '../Flex';
 import { useField } from './FieldContext';
 import { Typography } from '../Typography';
 
 const TypographyAsterisk = styled(Typography)`
   line-height: 0;
 `;
 
-export const FieldLabel = ({ children, required, ...props }) => {
+const Action = styled(Box)`
```
ronronscelestes, line 13: Proposal from my comment posted before: 
`const Action = styled(Flex)`
That would fix the alignment issue

gu-stav: That's a great proposal. I keep forgetting that `styled` can be used that way. Thanks.

Added in d707436e291900c0d4aefb947f2ae7d7b3bd9fa4

gu-stav:
I've add a line-height reset in d1d67a5798140011995a29b48f3c56a08dae4cd6 aswell, to improve the following cases:

| Before  | After  |
|---|---|
| Screen Shot 2022-02-17 at 12 40 46 | Screen Shot 2022-02-17 at 12 40 28 |

packages/strapi-design-system/src/Field/FieldLabel.js
```diff
@@ -1,14 +1,23 @@
...
 import { useField } from './FieldContext';
 import { Typography } from '../Typography';
 
 const TypographyAsterisk = styled(Typography)`
   line-height: 0;
 `;
 
-export const FieldLabel = ({ children, required, ...props }) => {
+const Action = styled(Flex)`
+  line-height: 1;
```
ronronscelestes, line 13: Good catch! I think the tooltip causes this issue of misalignment. 
I tried `line-height: 0` as for TypographyAsterisk and it seems to be even better aligned, not a fan of those tricks but not sure how to handle this, WDYT? @gu-stav 

image

gu-stav: Me neither, but I think it this case it gets the job done 🤓

HichamELBSI:
> I've add a line-height reset in [d1d67a5](https://github.com/strapi/design-system/commit/d1d67a5798140011995a29b48f3c56a08dae4cd6) aswell, to improve the following cases a little:
> 
> Before	After
> Screen Shot 2022-02-17 at 12 40 46	Screen Shot 2022-02-17 at 12 40 28

Not sure to see the benefits here? 🤔

gu-stav:
@HichamELBSI It's hard to spot indeed - if you open both images side-by-side you'll see that after the change, the "i" Icon is a little (1 - 2px) more aligned. It's still sits too high (I'd say) but it's a little better with line-height: 0 ...

gu-stav:
@HichamELBSI plus I'm adding the bug label, since this fixes a bug, where the i18n icon is missing from the content manager.

ronronscelestes (approved):
LGTM!

PR #537, opened 2022-02-17
CardAsset: avoid stretching images

gu-stav:
Before this PR images in the media library were stretched, if they don't fit the proportions of `CardAsset`. That's something we should avoid and instead always show the real proportions of the image.

| Before  | After  | 
|---|---|
| Screen Shot 2022-02-17 at 13 09 45 |  Screen Shot 2022-02-17 at 13 09 37  |

Refs: https://strapi-inc.atlassian.net/browse/CONTENT-89

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.0.1-alpha.75",
+  "version": "0.0.1-alpha.76",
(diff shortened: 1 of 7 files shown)


ronronscelestes (approved):
Nice, LGTM!

HichamELBSI (approved):
LGTM

PR #544, opened 2022-02-22
Move e2e test from Jest-Playwright to official Playwright runner

HichamELBSI:
Move e2e test from jest-playwright to official Playwright test runner

- A codemod script made with jscodeshift to update all the `*.e2e.js` files:
    - Add an import statement `import { test } from '@playwright/test';`
    - Add `test.` before `describe / beforeAll / beforeEach / afterAll / afterEach`
    - Replace `it(...)` by `test(...)`
    - Add `page` object params to async functions
- Several modifications has been made to match the Playwright API
- A new (better for CI) Playwright config (retry on CI, forbid `.only` test on CI, timeout, baseUrl, etc)
- Usage of `page.locator` instead of `page.$` (Highly recommended by Playwright)
- Parallelizing tests `test.describe.parallel`
- `*.e2e.js` do not supports `import`


packages/strapi-design-system/codemod/playwright_runner_test_files.js
```diff
@@ -0,0 +1,55 @@
...
+        j(path).replaceWith(j.identifier('test.' + path.node.name));
+      }
+      if (path.node.name == 'it') {
+        j(path).replaceWith(j.identifier('test'));
+      }
+    })
+    .toSource();
+
+  return result;
+}
```
HichamELBSI, line 55: ℹ️  I think we should keep all the codmod (maybe useful for the community)

packages/strapi-design-system/src/index.js
```diff
@@ -49,7 +49,6 @@ export * from './Switch';
 export * from './Tabs';
 export * from './Table';
 export * from './Tag';
-export * from './Text';
```
gu-stav, line 52: I guess `Text` is not needed anymore?

packages/strapi-design-system/codemod/playwright_runner_test_files.js
gu-stav: Very cool. Should we rename the file or add some documentation, so that it is possible in a couple of months to understand why it is there and what it is doing? Maybe by co-locating a README?

gu-stav (approved):
Looks very good to me. Nice job! ✨

packages/strapi-design-system/src/index.js
HichamELBSI: 👍🏽

packages/strapi-design-system/src/Badge/__tests__/Badge.e2e.js
```diff
@@ -1,13 +1,15 @@
-import { injectAxe, checkA11y } from 'axe-playwright';
+const { injectAxe, checkA11y } = require('axe-playwright');
 
-describe('Badge', () => {
-  beforeEach(async () => {
+const { test } = require('@playwright/test');
+
+test.describe('Badge', () => {
```
ronronscelestes, line 5: some of the tests have `test.describe.parallel` and others only `test.describe`, is it wanted?

ronronscelestes (approved):
LGTM 🔥

.github/workflows/playwright-ci.yml
```diff
@@ -1,11 +1,11 @@
 name: Playwright tests
 on:
   pull_request:
-    branches: ["main"]
+    branches: ["main", "1.0.0"]
```
soupette, line 4: Why not running the tests on all branches instead?

packages/strapi-design-system/codemod/playwright_runner_test_files.js
```diff
@@ -0,0 +1,55 @@
+// jscodeshift can take a parser, like "babel", "babylon", "flow", "ts", or "tsx"
+// Read more: https://github.com/facebook/jscodeshift#parser
+export const parser = 'babel';
+
+// Press ctrl+space for code completion
+export default function transformer(file, api) {
+  const j = api.jscodeshift;
```
soupette, line 7: Can we avoid naming constants as `j,s,n` ? 😄

packages/strapi-design-system/playwright.config.js
```diff
@@ -0,0 +1,40 @@
+const { devices } = require('@playwright/test');
+
+const config = {
+  testMatch: '**/__tests__/*.e2e.js',
+  timeout: 10000,
+  // Forbid test.only on CI
```
soupette, line 6: Why not always forbidding it?

HichamELBSI: 🤔  I don't understand your comment here. Do you mean forbid `.only` on local too?
On local, if you add a `.only` on a test, I guess you want to run only that test 😅 . 
On the CI, you want to run the whole test suite in all cases. (forbidOnly will trigger an error if `.only`)

packages/strapi-design-system/codemod/playwright_runner_test_files.js
HichamELBSI: Nice catch, it is the default [ast-explorer](https://astexplorer.net/) convention naming (same as [facebook/jscodeshift](https://github.com/facebook/jscodeshift#the-jscodeshift-api)). But you are right, I will rename those constants.

.github/workflows/playwright-ci.yml
HichamELBSI: You are right!

HichamELBSI: We should!

packages/strapi-design-system/playwright.config.js
HichamELBSI: I mean, if we always forbid the `.only`, you'll not be able to run only one test if you want (on local).

packages/strapi-design-system/codemod/playwright_runner_test_files.js
soupette: If it's a convention then you can leave it as is. Thanks for the clarification

packages/strapi-design-system/playwright.config.js
soupette: Sure but I guess that you can run your test locally and then the precommit hook can fail if your test contains `.only`

PR #545, opened 2022-02-22
SimpleMenu: expose onOpen and onClose callbacks

gu-stav:
As a next step of https://github.com/strapi/strapi/pull/12213 I'd like to unify the behavior of all menus in the dynamic table. Relations are currently using it's own implementation, because their content is fetched dynamically. By exposing `onOpen` and `onClose` events, we know when a user opens the menu, so that we can start fetching the relation and display the according states in the menu.


HichamELBSI (approved):
LGTM

HichamELBSI:
Can you fix the conflicts please 😄

PR #552, opened 2022-02-24
SimpleMenu: Add size prop and define small size

gu-stav:
As proposed in https://github.com/strapi/strapi/pull/12213 this PR introduces a `size` prop for `SimpleMenu`, to decrease the padding of the trigger button. The default (size=M) stays the same as it is.

## Screenshots
Screenshot 2022-02-24 at 12 57 07
.


packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -206,4 +211,5 @@ SimpleMenu.propTypes = {
   children: PropTypes.oneOfType([PropTypes.arrayOf(menuItemType), menuItemType]).isRequired,
   id: PropTypes.string,
   label: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.element]).isRequired,
+  size: PropTypes.string.isRequired,
```
HichamELBSI, line 214: Can you the update proptype please -> `Proptypes.oneOf(['S', 'M']`.

Also, IMO this prop shouldn't be required. The default value should be set to M.

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -88,13 +92,13 @@ MenuItem.propTypes = {
...
 
-export const SimpleMenu = ({ label, children, id, as: asComp, ...props }) => {
+export const SimpleMenu = ({ label, children, id, as: asComp, size, ...props }) => {
   const menuButtonRef = useRef();
   const menuId = useId('simplemenu', id);
   const [visible, setVisible] = useState(false);
   const [focusedItemIndex, setFocusItem] = useState(0);
   const childrenArray = Children.toArray(children);
-  const Component = asComp || Button;
+  const Component = asComp || (size === 'S' ? StyledButtonSmall : Button);
```
HichamELBSI, line 101: Maybe we can create a new constant here, to make it a bit more readable
```js
const defaultComponent = size === 'S' ? StyledButtonSmall : Button;
const Component = asComp || defaultComponent;
```

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
HichamELBSI: Also, the `size` prop won't work is `as` is defined. Can you add some storybook description to this prop as specify that please. 

```js
/**
 * Here
*/
size: PropTypes.oneOf(['S', 'M'],
```

You can also add some description for all the props if you want 🙊

HichamELBSI (changes requested):

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
gu-stav: Done in 851b710fc230a62fe933e426ecb4cb9a89f7121d

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
gu-stav: Good idea. Implemented in 851b710fc230a62fe933e426ecb4cb9a89f7121d

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -201,9 +207,19 @@ SimpleMenu.displayName = 'SimpleMenu';
...
   as: PropTypes.any,
   children: PropTypes.oneOfType([PropTypes.arrayOf(menuItemType), menuItemType]).isRequired,
   id: PropTypes.string,
   label: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.element]).isRequired,
+
+  /**
+   * Size of the trigger button
+   */
+
+  size: Proptypes.oneOf(['S', 'M']),
```
HichamELBSI, line 224: `PropTypes`

HichamELBSI (changes requested):

gu-stav:
@HichamELBSI Thanks for the review again. I think everything you suggested makes sense. Before merging this, I'd like to talk with @maevalienard on Monday, because this change would need to be reflected in Figma too (I guess).

maevalienard (approved):

maevalienard:
Can we add to the StoryBook documentation the purpose of the S size SimpleMenu? :)

HichamELBSI (approved):
Awesome thanks

HichamELBSI:
I've updated the target branch to 1.0.0 Can you fix the conflicts please 🙏🏽

PR #554, opened 2022-02-28
Datepicker: make min/ max date configurable and extend default range

gu-stav:
This PR adds two new props the the `DatePicker` component:

- `minDate`: minimum date which is used by the year select
- `maxDate`: maximum date which is used by the year select

It also extends the default range of the `DatePicker` component to minus 200 years and plus 15 years and replaces the default placeholder with the current date instead of 1970.

💡 I didn't name the prop `minYear` or `maxYear` because these could also be used at some point to validate the input for a max/ min value.

## Demo

- https://design-system-git-content-132-min-max-date-strapijs.vercel.app/?path=/story/design-system-components-datepicker--min-max-date

Refs: https://github.com/strapi/strapi/issues/12441
Refs: https://strapi-inc.atlassian.net/browse/CONTENT-132

### DIFF
--- a/packages/strapi-design-system/src/DatePicker/DatePicker.js
+++ b/packages/strapi-design-system/src/DatePicker/DatePicker.js
@@ -20,6 +20,8 @@ export const DatePicker = ({
   onClear,
   clearLabel,
   disabled,
+  minDate,
+  maxDate,
   id,
   ...props
 }) => {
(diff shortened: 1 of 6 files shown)


HichamELBSI (commented):
Thanks !!

HichamELBSI (approved):

MarionLemaire (commented):
Tested on Getstarted using Chrome : dates can be set from 1822 to 2037

PR #556, opened 2022-02-28
Loader: add small usage to docs, use theme spaces

gu-stav:
This PR adds documentation on how to use the `small` prop and a preview how it will look like. It also removes the hard coded `25px` and replaces it with a design-system constant, which will result in `24px`.


HichamELBSI (approved):
LGTM

PR #558, opened 2022-03-01
SimpleMenu: remove invalid HTML attribute label

gu-stav:
This removes the attribute `label` from the trigger component of the `SimpleMenu`. It is not a valid HTML attribute and since props are spread, it is already possible to pass in `aria-label`.


packages/strapi-design-system/tests/__snapshots__/storyshots.spec.js.snap
```diff
@@ -11741,7 +11741,7 @@ exports[`Storyshots Design System/Components/DatePicker base 1`] = `
                 class="c8"
                 id="datepicker-45"
                 name="datepicker"
-                placeholder="2/28/2022"
+                placeholder="3/1/2022"
```
gu-stav, line 11744: @HichamELBSI Something is not right with this test. It seems to be flaky on my machine. Do you have an idea? I think jest timers are already enabled ...

HichamELBSI: The default DatePicker placeholder should be `placeholder="1/1/1970"`. Has anything changed?

gu-stav: I've updated this yesterday, because it feels weird to me to use that: https://github.com/strapi/design-system/pull/554/files#diff-2f9c2263c88e72993a7228e624ea72dd2d4a7ef4fd1831d086bd09b91ee1137eR33

Still: since jest has fixed timers, it should not use the current date, but the one defined in https://github.com/strapi/design-system/pull/554/files#diff-4696489232b498b61c8bf2dedaee9ce2d77a5cf040783c0888c8e9fc13d7d944R4-R6 No?

HichamELBSI: Are you sure that storybook storyshots addon is using any jest related code? 🤔

gu-stav: Well: the code has been there before already :D let me check.

gu-stav: I'll remove the date changes manually from this PR ...

HichamELBSI: It's confusing because it's a .spec.js file also but I think @storybook/addon-storyshots uses only the stories file to generate storyshots 👍🏽

PR #559, opened 2022-03-01
DatePicker: allow passing in a placeholder, to fix a flaky test

gu-stav:
Restores behavior from before https://github.com/strapi/design-system/pull/554, but allows to pass in a placeholder attribute manually.


HichamELBSI (approved):
Thanks! LGTM

PR #563, opened 2022-03-02
SimpleMenu: only render label in case it isn't a custom component

gu-stav:
In https://github.com/strapi/design-system/pull/558 I removed the label from the `SimpleMenu`, but I missed the case, that the prop is used by the underlying `Button` component to render a tooltip.

This PR passes the prop down to the `Button` again and adds a check: in case `label` is a custom component, it won't render it as a tooltip. In case it' anything else (string, number) it works as before.


HichamELBSI (approved):
LGTM

PR #564, opened 2022-03-03
Datepicker e2e tests

ronronscelestes:
## What

Fixed Datepicker story using text without theme that broke Datepicker e2e tests in dark mode


HichamELBSI (commented):
LGTM good job!

PR #565, opened 2022-03-03
DatePicker: fix onSelect event

gu-stav:
Fixes the `onEscape` callback of the `DatePickerCalendar` component. Before, it wasn't possible (anymore) to escape the focus trap.

https://user-images.githubusercontent.com/2244375/156546491-0136f981-bb76-41aa-94f1-966f60f1fc35.mp4

Demo: https://design-system-git-fix-datepicker-escape-strapijs.vercel.app/?path=/story/design-system-components-datepicker--base


ronronscelestes (approved):
Nice catch 🚀

PR #572, opened 2022-03-09
FolderCard: create new component

gu-stav:
## Description

This PR adds a new component called `FolderCard` which should look like the following:

Screenshot 2022-03-09 at 16 13 16

ℹ️ Why `FolderCard` instead of `Folder`? I found `Folder` ambiguous and too generic. Also, the component will most certainly appear in the context of other `Card` components, which are often called `ImageCard`, `AssetCard` ...

### Behavior

- on double click on the whole area, it should navigate to a given target
- on single click on the folder name, it should navigate to the given target

### TODO:

- [ ] Do we have a way to programmatically call `navigate()` on double click? Could we pass it in from `react-router-dom`? https://reactrouter.com/docs/en/v6/api#usenavigate
- [ ] Is there an easier way to style the `FolderIcon`?
- [ ] Does the naming work?

Figma: https://www.figma.com/file/doUJzYdMJTjpgtw6Lyaa7D/%5B%E2%9C%85Done%5D-Content-squad---Media-folders?node-id=671%3A4833

## Demo

- https://design-system-git-feat-ml-folder-strapijs.vercel.app/?path=/story/design-system-components-foldercard--base

### DIFF
--- a/packages/strapi-design-system/src/BaseLink/BaseLink.js
+++ b/packages/strapi-design-system/src/BaseLink/BaseLink.js
@@ -4,6 +4,7 @@ import styled from 'styled-components';
 
 const A = styled.a`
   cursor: pointer;
+  text-decoration: ${({ textDecoration }) => textDecoration || 'underline'};
 `;
 
 export const BaseLink = React.forwardRef(({ href, rel, target, disabled, isExternal, ...props }, ref) => {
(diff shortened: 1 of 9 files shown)


gu-stav:
It has been decided that this component will live in the helper-plugin of strapi.

PR #576, opened 2022-03-11
Fix / Primary500 

ronronscelestes:
## What

Specify sync for dark primary500
It fixes accessibility on button primary background color on hover 

before
image

after
image


HichamELBSI (approved):
LGTM

PR #578, opened 2022-03-15
Build v2 components

HichamELBSI:
Provide every links components for the v2

```js
import { Link } from '@strapi/design-system/v2/Link';
import { SimpleMenu, MenuItem } from '@strapi/design-system/v2/SimpleMenu';
import { SubNavLink } from '@strapi/design-system/v2/SubNav';
import { LinkButton } from '@strapi/design-system/v2/LinkButton';
import { NavLink } from '@strapi/design-system/v2/MainNav';
import { Pagination } from '@strapi/design-system/v2/Pagination';
```


packages/strapi-design-system/createIndexFiles.js
```diff
@@ -1,8 +1,10 @@
 const fs = require('fs-extra');
 const path = require('path');
 
-const excludedFolders = ['helpers', '.DS_Store'];
-const fileNames = fs.readdirSync(path.resolve(__dirname, 'src'));
+const excludedFolders = process.env.IS_V2 === 'false' ? ['helpers', '.DS_Store'] : ['helpers', '.DS_Store', 'v2'];
```
soupette, line 4: Can we create a constant `isV2Build` ? 

```
const isV2Build = process.env.IS_V2 === 'true'
const excludedFolders = ['helpers', '.DS_Store'];
let entryFolder = path.resolve(__dirname, 'src');
let outputPath = path.resolve(__dirname, 'dist', `${name}.js`);

if (isV2Build) {
 excludedFolders.push('v2')
 entryFolder = path.resolve(__dirname, 'src', 'v2');
outputPath = ...
}
```

Also why are you using the `=== 'false'` instead of the `'true'` one here?

packages/strapi-design-system/webpack.config.js
```diff
@@ -2,19 +2,27 @@ const fs = require('fs');
...
+const isV2Build = process.env.IS_V2 === 'true';
+const outputPath = isV2Build ? path.join(distPath, 'v2') : distPath;
+
 // Allows to create distinct bundles in the dist folder
 // for people wanting to import only specific components such as
 // import Button from '@strapi/design-system/Button
-const excludedFolders = ['helpers', '.DS_Store'];
-const fileNames = fs.readdirSync(path.resolve(__dirname, 'src'));
+const entryFolder = isV2Build ? path.join(__dirname, 'src', 'v2') : path.join(__dirname, 'src');
+const excludedFolders = isV2Build ? ['helpers', '.DS_Store'] : ['helpers', '.DS_Store', 'v2'];
```
soupette, line 13: Can you create a util here and reuse it in both the createIndexFiles.js and the webpack config?

packages/strapi-design-system/webpack.config.js
```diff
@@ -2,19 +2,27 @@ const fs = require('fs');
...
+const fileNames = fs.readdirSync(path.resolve(entryFolder));
 const entry = fileNames
   .filter((name) => !excludedFolders.includes(name))
+  // .filter((name) => (isV2Build ? true : !name.includes(path.join(__dirname, 'src', 'v2'))))
```
soupette, line 17: You can remove this part and the other comments.

packages/strapi-design-system/webpack.config.js
```diff
@@ -66,6 +74,7 @@ module.exports = {
     {
       react: 'react',
       'react-dom': 'react-dom',
+      'react-router-dom': 'react-router-dom',
```
soupette, line 77: Here you can do instead is to use the package.json peerDependencies:

```
const packageJSON = require('./package.json');

const packageJson = require('./package.json');

const externalNodeModules = [];
[
  ...Object.keys(packageJson.peerDependencies),
].forEach(module => {
  nodeModules.push(new RegExp(`^${module}(/.+)?$`));
});

module.exports = {
 // ...
 externals: externalNodeModules
};
```

soupette (changes requested):

soupette (approved):
LGTM for the build part!

packages/strapi-design-system/src/v2/LinkButton/LinkButton.stories.mdx
```diff
@@ -0,0 +1,153 @@
...
+## Usage
+
+The most common LinkButton component used in the interface (primary action). It can have different variants and meaning.
+
+Use the `to` parameter to redirect using the router's routes, `href` parameter for an external url.
+
+<Canvas>
+  <Story name="base">
+    <Box paddingBottom={2}>
+      <LinkButton variant="default" href="https://strapi.io/">
```
ronronscelestes, line 45: I think we are missing a target logic for this, we have target self even with href 🤔

image

packages/strapi-design-system/src/Link/Link.stories.mdx
```diff
@@ -1,114 +1,44 @@
...
-
-Display icons on the left or right side.
-
-<Canvas>
-  <Story name="icons">
-    <Stack spacing={5}>
       <div>
-        <Link isExternal href="https://strapi.io/" startIcon={<Strapi />}>
-          Strapi
+        <Link to="https://strapi.io/">External link</Link>
```
ronronscelestes, line 18: could we replace the to props with an href prop for the external links? those are breaking
image

ronronscelestes (changes requested):
Nice job 🔥

packages/strapi-design-system/src/v2/LinkButton/LinkButton.stories.mdx
HichamELBSI: `_self` is the default target attribute for anchor tags. For the v2 components, you need to pass `isExternal` to define that the link is an external link (`isExternal` will set the target to `_blank`. We can't use `href` to define that the link is external because with NextJS, href is used for internal links too.)

ronronscelestes: I see makes sense, thank you for adding it in the Story, will be easier to remember!

PR #579, opened 2022-03-16
FieldInput: Make input element cover the whole box

gu-stav:
Before this PR the actual `input` element didn't cover the full height of it's wrapper. This leads to the problem that you can not click everywhere since the click might end up in the wrapper.

This PR removes height from the wrapper and bumps to a height of the input elements to the same as the container by using padding. It is not possible to use `height` on input fields - therefore I had to use some magic numbers.

| Before  | After  |
|---|---|
| Screenshot 2022-03-16 at 14 18 00  | Screenshot 2022-03-16 at 14 18 13  |

## Demo

- https://design-system-git-fix-field-input-height-strapijs.vercel.app/?path=/story/design-system-components-field--base

### DIFF
--- a/packages/strapi-design-system/src/Field/FieldInput.js
+++ b/packages/strapi-design-system/src/Field/FieldInput.js
@@ -7,14 +7,19 @@ import { useField } from './FieldContext';
 import { Flex } from '../Flex';
 import { Box } from '../Box';
 
+// padding-[top|bottom] must ensure, the input matches the height of getThemeSize('input')
+const PADDING_Y = {
+  S: 6.5,
+  M: 10.5,
+};
+
 const Input = styled.input`
   border: none;
   border-radius: ${({ theme }) => theme.borderRadius};
-  // The padding-[top|bottom] must ensure, the input matches the height of getThemeSize('input')
-  padding-bottom: ${10.5 / 16}rem;
+  padding-bottom: ${({ size }) => `${PADDING_Y[size] / 16}rem`};
   padding-left: ${({ theme, hasLeftAction }) => (hasLeftAction ? 0 : theme.spaces[4])};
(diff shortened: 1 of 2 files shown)


HichamELBSI (approved):
LGTM thanks.

ronronscelestes (changes requested):
hey, on my way to check more precisely how we can handle it but not using getThemeSize also means that we lose the S size documented in the DS system 😰

gu-stav:
@ronronscelestes Ah, I didn't see how/ if the small size would be affected by that. If so (I'll double-check tomorrow), we could just reduce the padding for that size I guess. Or do you have something different in mind?

gu-stav:
@ronronscelestes I've fixed `size=S` - thanks again. Very good catch!

For now, I'd keep the padding definition local, because I don't think we should advertise it since all other components use an explicit height. It works for them because they aren't native input fields.

I'd consider this a fix and not a methodology. If we would put it into a general schema/ function, we'd need to find a good name for it and start to advertise its usage. I'd therefore keep it local to the component if it's ok for you.

ronronscelestes:
hey @gu-stav 
thanks for the fix!

tbh I'm not 100% fan we remove usage of sizes object for the inputs, that would be nice indeed to find a more global way to handle this.
maybe we could at least add a card in the DS backlog to make sure we don't forget it? WDYT @HichamELBSI?

HichamELBSI:
Can we merge this one @gu-stav ?

gu-stav:
@HichamELBSI from my side: yes. @ronronscelestes was not happy with the implementation and asked for your feedback. That's the reason why I kept it open.

HichamELBSI:
Let's merge this PR and create a card in the DS backlog to not forget how we can handle this globally.

PR #582, opened 2022-03-20
Icon Component: Add width and height

MattieBelt:
# Add width and height to Icon Component

## Description
Use theme spaces instead of just CSS for width and height. 

## Demo
Example on: 
 
## API
```jsx
<Icon width={6} height={6} color="danger500" as={Trash} />
```

### DIFF
--- a/packages/strapi-design-system/src/Icon/Icon.js
+++ b/packages/strapi-design-system/src/Icon/Icon.js
@@ -4,6 +4,9 @@ import { Box } from '../Box';
 import styled from 'styled-components';
 
 const IconWrapper = styled(Box)`
+  width: ${({ theme, width }) => theme.spaces[width]};
+  height: ${({ theme, height }) => theme.spaces[height]};
+
   path {
     fill: ${({ color, theme }) => theme.colors[color]};
   }
@@ -19,9 +22,13 @@ Icon.displayName = 'Icon';
 Icon.defaultProps = {
   color: 'neutral600',
   colors: () => undefined,
+  height: 1,
+  width: 1,
 };
 
 Icon.propTypes = {
   color: PropTypes.string,
   colors: PropTypes.func,
+  height: PropTypes.number,
+  width: PropTypes.number,
 };


gu-stav (approved):
I think it's a nice addition, but I'd like to have another opinion from @HichamELBSI or @ronronscelestes .

ronronscelestes:
Hi @MattieBelt, thank you for your addition, nice one!
I think it's a good idea too, though I would find a way to still allow custom sizing in case `theme.spaces[width/height]` don't exist. We have some components in Strapi using Icon with `${160 / 16}rem` or `${88 / 16}rem` for example.
Maybe we could check if `theme.spaces[width/height]` exists, if not, just passing the value as it is? @gu-stav @HichamELBSI

MattieBelt:
@ronronscelestes 
> Hi @MattieBelt, thank you for your addition, nice one! I think it's a good idea too, though I would find a way to still allow custom sizing in case `theme.spaces[width/height]` don't exist. We have some components in Strapi using Icon with `${160 / 16}rem` or `${88 / 16}rem` for example. Maybe we could check if `theme.spaces[width/height]` exists, if not, just passing the value as it is? @gu-stav @HichamELBSI

Yes I think that backwards compatibility could work out in such a way 👍🏼

MattieBelt:
What should be the default value for these fields? Would this maybe also affect components using `Icon` without any `height` or `width`?

@gu-stav @ronronscelestes

packages/strapi-design-system/src/Icon/Icon.js
```diff
@@ -4,8 +4,8 @@ import { Box } from '../Box';
 import styled from 'styled-components';
 
 const IconWrapper = styled(Box)`
-  width: ${({ theme, width }) => theme.spaces[width]};
-  height: ${({ theme, height }) => theme.spaces[height]};
+  width: ${({ theme, width }) => theme.spaces[width] ? theme.spaces[width] : width};
+  height: ${({ theme, height }) => theme.spaces[height] ? theme.spaces[height] : height};
```
gu-stav, line 8: I think you could shorten that:

```suggestion
  width: ${({ theme, width }) => theme.spaces[width] ?? width};
  height: ${({ theme, height }) => theme.spaces[height] ?? height};
```

gu-stav, line 8: I think you could shorten that:

```suggestion
  width: ${({ theme, width }) => theme.spaces[width] ?? width};
  height: ${({ theme, height }) => theme.spaces[height] ?? height};
```

MattieBelt: I just saw that `handleResponsiveValues` also allows string values, so what about using it here as well? 
@ronronscelestes @gu-stav

> Should these new props also allow responsive properties just like the padding props in the Box component?

ronronscelestes: mmh good question!
the utility of `handleResponsiveValues` is to allow responsive values depending on media queries we chose - is it something we want to include here 🤔

MattieBelt:
> mmh good question! the utility of `handleResponsiveValues` is to allow responsive values depending on media queries we chose - is it something we want to include here thinking

Well, as the spaces in the default theme are absolute, it might be good to allow for responsive values, e.g. mobile or desktop.
But let me know WDYT?

@ronronscelestes

packages/strapi-design-system/src/Icon/Icon.js
HichamELBSI: Since the Icon component is based on `Box` and `Box` already handle width and height. I'm wondering if we really need that modification 🤔 

https://github.com/strapi/design-system/blob/main/packages/strapi-design-system/src/Box/Box.js#L74

HichamELBSI (changes requested):

ronronscelestes:
@HichamELBSI I think you are right, except that Box's width and height props don't allow custom and as I said previously some Icon component inside Strapi uses `${160 / 16}rem` or `${88 / 16}rem` for width - height props. That wouldn't mean we couldn't override it using styled components though but I don't know if it's preferable? 

@MattieBelt really good question about handling responsive values for Icon, tbh I don't know if it's something that we'll need or want, I don't have a strong opinion about that. @HichamELBSI @gu-stav WDYT?

HichamELBSI:
> except that Box's width and height props don't allow custom

Not sure to understand, sorry. What do you mean by "custom"?

ronronscelestes:
> > except that Box's width and height props don't allow custom
> 
> Not sure to understand, sorry. What do you mean by "custom"? @ronronscelestes You can do `width={'${160 / 16}rem'}` with the box. Did I miss something?

I absolutely forgot that fact, my bad! 🤦🏻‍♀️

packages/strapi-design-system/src/Icon/Icon.js
soupette: I agree with you @HichamELBSI maybe the issue is with the documentation we need to display these props in the storybook instead.

HichamELBSI:
https://github.com/strapi/design-system/pull/582#discussion_r833112363

@MattieBelt Since it can be done with the current implementation, do you think we still need that PR?

soupette:
> [#582 (comment)](https://github.com/strapi/design-system/pull/582#discussion_r833112363)
> 
> @MattieBelt Since it can be done with the current implementation, do you think we still need that PR?

@HichamELBSI I think we need to open a PR to display the props for the components that inherit from `Flex` and `Box` it would solve the issue.

Issue #584, opened 2022-03-20
Disabled Switch doesn't show if the value is on or off


dzcpy:
![image](https://user-images.githubusercontent.com/203980/159191457-0185c165-a512-4ca3-9377-af1ed83816f3.png)
As shown in the image above, on the left side it's the disabled Switch, but actually it has a default value although it can't be changed. But since the change of the design, there's no way I can tell if it's on or off. It's quite inconvinient. Please make it visually distinguishable. Thanks!

gu-stav:
Thank you for opening the issue. I forgot about this case during the refactoring and opened another PR to fix the issue: https://github.com/strapi/design-system/pull/585 Does that work for you?

PR #585, opened 2022-03-21
ToggleCheckbox: improve disabled state styles and show default value

gu-stav:
With the latest refactoring in https://github.com/strapi/design-system/pull/538 it is not possible anymore:

- to visually distinguish the disabled `ToggleCheckbox`: this PR lightens the styles, in this case, a bit (contrast is still accessible)
- to show the default value, if the component is disabled: I didn't think about this case - this PR restores the previous behavior

---

- Fixes https://github.com/strapi/design-system/issues/584
- Example https://design-system-git-fix-toggle-disabled-strapijs.vercel.app/?path=/story/design-system-components-togglecheckbox--disabled

### DIFF
--- a/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
+++ b/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
@@ -25,10 +25,9 @@ const ToggleCheckboxWrapper = styled(Box)`
 const ValueBox = styled(Flex).attrs({
   hasRadius: true,
 })`
-  background-color: ${({ theme, disabled, checked }) => (checked && !disabled ? theme.colors.neutral0 : 'transparent')};
+  background-color: ${({ theme, checked }) => (checked ? theme.colors.neutral0 : 'transparent')};
   border: 1px solid
-    ${({ theme, disabled, checked }) =>
-      checked && checked !== null && !disabled ? theme.colors.neutral200 : theme.colors.neutral100};
+    ${({ theme, checked }) => (checked && checked !== null ? theme.colors.neutral200 : theme.colors.neutral100)};
   position: relative;
   user-select: none;
   z-index: 2;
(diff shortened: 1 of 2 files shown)


ronronscelestes (approved):
LGTM!

packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
```diff
@@ -76,35 +75,23 @@ export const ToggleCheckbox = React.forwardRef(
...
-            checked={checked === null ? false : !checked}
-            disabled={disabled}
-          >
+          <ValueBox paddingLeft={7} paddingRight={7} aria-hidden={true} checked={checked === null ? false : !checked}>
             <Typography
               variant="pi"
               fontWeight="bold"
               textTransform="uppercase"
-              textColor={disabled ? 'neutral600' : offCheckboxLabelColor}
+              textColor={disabled ? 'neutral500' : offCheckboxLabelColor}
```
HichamELBSI, line 83: We have a very low contrast errors here with the `neutral500`. 
Why do we need to change the text color value?

Screenshot 2022-03-23 at 12 14 29

packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
```diff
@@ -76,35 +75,23 @@ export const ToggleCheckbox = React.forwardRef(
...
-            checked={checked === null ? false : checked}
-            disabled={disabled}
-          >
+          <ValueBox paddingLeft={7} paddingRight={7} aria-hidden={true} checked={checked === null ? false : checked}>
             <Typography
               variant="pi"
               fontWeight="bold"
               textTransform="uppercase"
-              textColor={disabled ? 'neutral600' : onCheckboxLabelColor}
+              textColor={disabled ? 'neutral500' : onCheckboxLabelColor}
```
HichamELBSI, line 94: Same here about the color contrast issue

dzcpy:
Please make it available for Strapi, it's a bit annoying not being able to see the status of a disabled switch.

packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
gu-stav: I agree. However, we have the problem that we visually need to distinguish between disabled and active:

### With the change (neutral500)

disabled / active

Screenshot 2022-04-04 at 11 30 55

### Without the change (neutral600)

disabled / active

Screenshot 2022-04-04 at 11 31 30

@maevalienard Do you have an idea what we could do to visually distinguish between disabled and not-disabled without introducing color-contrast issues?

maevalienard: Yep, we had one similar case for the multi-select with tags - I've added the state in this frame: https://www.figma.com/file/PICiE8O4NLrHO1lhJftLdE/Strapi---UI-Kit-%F0%9F%A7%A9?node-id=10500%3A120443 :)

gu-stav: @maevalienard Amazing, thanks! It works well on light-mode I think, but I'm not sure about the dark-mode:

disabled / active

Screenshot 2022-04-04 at 11 55 52

Screenshot 2022-04-04 at 11 55 58

packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
gu-stav: Addressed in 8cbac4fbe4f47ee42c1c2bb32d0856b1ee6dced1. Discussion continues above.

HichamELBSI (approved):
Good job! Thanks for the update

PR #586, opened 2022-03-21
[Dark mode] Memorize mode

ronronscelestes:
## What

- Fixed mode chosen not memorized
- Refacto Theme component used in `preview.js`
- Added `window.matchMedia` to `test-bundler.js `  ➡️
- Needed to use `import { useDarkMode } from 'storybook-dark-mode'` in Theme component

## How to test

- Launch Storybook
- Change mode 
- Switch to a different story, reload the page
- It should memorize your mode chosen
- Shouldn't break dark mode e2e tests


packages/strapi-design-system/.storybook/preview.js
```diff
@@ -17,7 +17,6 @@ export const parameters = {
   actions: { argTypesRegex: '^on[A-Z].*' },
   darkMode: {
     // Override the default dark theme
-    current: 'light',
```
HichamELBSI, line 20: Does it mean that the default displayed theme will be the dark one on storybook?

packages/strapi-design-system/test-bundler.js
```diff
@@ -19,6 +19,20 @@ window.IntersectionObserver = () => ({
...
+    matches: false,
+    media: query,
+    onchange: null,
+    addListener: jest.fn(),
+    removeListener: jest.fn(),
+    addEventListener: jest.fn(),
+    removeEventListener: jest.fn(),
+    dispatchEvent: jest.fn(),
+  })),
+});
```
HichamELBSI, line 34: Not sure to understand why we need that?

HichamELBSI (changes requested):

packages/strapi-design-system/.storybook/preview.js
ronronscelestes: here you can find the mode choice "hierarchy" applied by the [addon](https://storybook.js.org/addons/storybook-dark-mode) 
image

the issue when we leave `current: 'light'` parameter is that everytime you go from one story to another or reload the page, it will set light as current.
with my modification:
- first time it will use your OS mode
- then the mode last used

WDYT @HichamELBSI ?

HichamELBSI: Alright thanks ! Awesome

HichamELBSI (approved):
LGTM

PR #599, opened 2022-04-14
Move vertical stack from Box to Flex with direction set to column

ronronscelestes:
## What

Move vertical stack from Box to Flex with direction set to column
Align-items set to stretch to keep previous UI behavior

## Why

See DS Jira ticket DS-14
Will allow StackV to handle flex props (align-items, justify-content)
Remove some needs to customize Stack with styled-props

## Snapshot 

image


HichamELBSI (approved):
LGTM

PR #611, opened 2022-05-16
Fix simple menu click and fix some stories props

HichamELBSI:
### What does it do?

Prevent default on mouse down event of the SimpleMenu component.

### Why is it needed?

Prevent a bug on React 18.

### How to test it?

Should work the same way

### DIFF
--- a/packages/strapi-design-system/src/Box/BoxProps.js
+++ b/packages/strapi-design-system/src/Box/BoxProps.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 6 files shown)


gu-stav (approved):
Looks good to me. Thanks!

gu-stav:
@HichamELBSI Does this need to be done on the `v2` version of the `SimpleMenu` too?

HichamELBSI:
> @HichamELBSI Does this need to be done on the `v2` version of the `SimpleMenu` too?

Yes, nice catch!

gu-stav (approved):
✨

PR #612, opened 2022-05-20
Fix click for React 18

HichamELBSI:
### What does it do?

Prevent default on click event.

### Why is it needed?

Fix Select usage on React 18 projects


soupette (approved):
LGTM

gu-stav (approved):
🚀

PR #613, opened 2022-05-26
fix(theme): use Typography to avoid color issue while switching theme

imklau:
### What does it do?

This PR replaces regular text with the Typography component in Theme stories. It fixes the dark text color on a dark background while switching between the Storybook themes.

### Why is it needed?

The text is almost impossible to read right now.

Zrzut ekranu 2022-05-26 o 14 20 44

### How to test it?

### Related issue(s)/PR(s)

-


HichamELBSI (approved):
Thanks for the contribution @imklau 😊

Issue #618, opened 2022-06-07
ReferenceError: Element is not defined when using `Combobox`


yeminkhaung-airitech:
### Required System information

- Npm version: 6.14.17
- NodeJS version: v14.19.3
- Strapi Design System version: 1.1.1
- Browser: Firefox 101.0
- React/React-DOM version: 18.1.0
- NextJS version: 10.0.6

### Describe the bug

I am using Next.js (React) and try to add starpi-design-system's `Combobox`. Error `ReferenceError: Element is not defined`  occurred when NextJS page is loaded (reference code below). `Box` component alone is working.

### Steps to reproduce the behavior

1. ComboBox is added according to documentation.
```
function HomePage {
  const [comboValue, setComboValue] = useState('');

  return (
      <>
        <Box padding={8} background="neutral100">
          <Combobox label="Food" value={comboValue} onChange={setComboValue}>
            <ComboboxOption value="hamburger">Hamburger</ComboboxOption>
            <ComboboxOption value="bagel">Bagel</ComboboxOption>
            <ComboboxOption value="tartuffo">Tartuffo</ComboboxOption>
            <ComboboxOption value="carbonara">Carbonara</ComboboxOption>
          </Combobox>
```
4. Start NextJS with `yarn dev`
5. load the page at localhost:3000
6. See error

### Full Error Log

```
yarn dev
yarn run v1.22.18
$ node server/server.js
Loaded env from /.env.local
info  - Using external babel configuration from /.babelrc
event - compiled successfully
> Ready on http://localhost:3000
event - build page: /404
wait  - compiling...
event - build page: /
event - build page: /
event - compiled successfully
ReferenceError: Element is not defined
    at eval (webpack-internal:///./src/Popover/Popover.js:180:75)
    at Module../src/Popover/Popover.js (\node_modules\@strapi\design-system\Combobox.development.js:598:1)
    at __webpack_require__ (\node_modules\@strapi\design-system\Combobox.development.js:30:30)
    at eval (webpack-internal:///./src/Popover/index.js:2:66)
    at Module../src/Popover/index.js (\node_modules\@strapi\design-system\Combobox.development.js:610:1)
    at __webpack_require__ (\node_modules\@strapi\design-system\Combobox.development.js:30:30)
    at eval (webpack-internal:///./src/Combobox/Combobox.js:24:67)
    at Module../src/Combobox/Combobox.js (\node_modules\@strapi\design-system\Combobox.development.js:370:1)
[...]

PR #622, opened 2022-06-14
Fix Element is not defined on SSR

HichamELBSI:
### What does it do?

Will fix the design system for SSR. IMO we should stop using Element because it will always be undefined on a server side rendered app.

### Related issue(s)/PR(s)

https://github.com/strapi/design-system/issues/618


gu-stav (approved):
Works for me! Thanks for fixing it and apologies for the delay.

HichamELBSI:
> Works for me! Thanks for fixing it and apologies for the delay.

No worries, thank you 😄

PR #628, opened 2022-06-20
[CONTENT-233] Card component border

ronronscelestes:
## What

- Added a border to improve the accessibility of the Card component with Dark mode
- Fixed CardAsset not applying top `border-radius` on the `repeating-conic-gradient` background (see images below)

image

## Snapshots

CardAsset background overflowing Card `border-radius` (not visible on the right side because `repeating-conic-gradient` property uses `neutral100` and transparent:
Screenshot 2022-06-20 at 20 43 13

Fix by adding `border-top-left-radius` and right (overflow hidden on Card component could also be a solution):
Screenshot 2022-06-20 at 20 48 58

### DIFF
--- a/packages/strapi-design-system/src/Card/CardAsset.js
+++ b/packages/strapi-design-system/src/Card/CardAsset.js
@@ -26,6 +26,8 @@ const CardAssetWrapper = styled.div`
   width: 100%;
   background: repeating-conic-gradient(${({ theme }) => theme.colors.neutral100} 0% 25%, transparent 0% 50%) 50% / 20px
     20px;
+  border-top-left-radius: 4px;
+  border-top-right-radius: 4px;
 `;
 
 export const CardAsset = ({ size, children, ...props }) => {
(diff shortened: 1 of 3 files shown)


packages/strapi-design-system/src/Card/CardAsset.js
```diff
@@ -26,6 +26,8 @@ const CardAssetWrapper = styled.div`
   width: 100%;
   background: repeating-conic-gradient(${({ theme }) => theme.colors.neutral100} 0% 25%, transparent 0% 50%) 50% / 20px
     20px;
+  border-top-left-radius: 4px;
+  border-top-right-radius: 4px;
```
gu-stav, line 30: What do you think about using `theme.borderRadius` instead, to keep this value consistent with all others?

ronronscelestes: nice catch - sorry I keep forgetting about this one

gu-stav (approved):
Very nice, thanks!

HichamELBSI (approved):

Issue #629, opened 2022-06-16
Wrong system font on Safari


tomtuv:
### Required System information

- Node.js version: 16.13.0
- NPM version: 8.10.0
- Strapi version: 4.2.0
- Database: MySQL
- Operating system: macOS 12.4

### Describe the bug

The content manager is not using the system font on Safari, Helvetica Neue is used instead of San Francisco. The bug is not present in other browsers.

### Steps to reproduce the behavior

Use a Safari extension like WhatFont to inspect the fonts on the page.

### Expected behavior

San Francisco should be used instead of Helvetica Neue as it is the macOS system font.

### Screenshots

<img width="615" alt="Capture d’écran 2022-06-16 à 19 11 24" src="https://user-images.githubusercontent.com/16865165/174128381-3d7a8d51-a1f2-4882-9972-8d647a9a2ac1.png">

### Code snippets

https://github.com/strapi/design-system/blob/610129978e4c1ab2018d842c33ad048e70af0ed8/packages/strapi-design-system/src/ThemeProvider/ThemeProvider.js#L156

https://github.com/strapi/strapi/blob/5cbb60ce4b652e84dd8d65ffa2713437ecb4e619/packages/core/admin/admin/src/content-manager/components/Wysiwyg/EditorStylesContainer.js#L19

### Additional context

This can be fixed by replacing `--apple-system` with `-apple-system` which is the correct font family name.

PR #630, opened 2022-06-21
DS: Fix global apple system font-family

gu-stav:
### What does it do?

Fixes https://github.com/strapi/design-system/issues/629. On MacOS we are currently using the wrong system-font notation. Github uses the same notion (`-apple-system` instead of `--apple-system`).

| Before  | After  | 
|---|---|
| Screen Shot 2022-06-21 at 17 45 19-fullpage  | Screen Shot 2022-06-21 at 17 45 13-fullpage  |
| Screen Shot 2022-06-21 at 17 46 08-fullpage  | Screen Shot 2022-06-21 at 17 46 15-fullpage  |
| Screen Shot 2022-06-21 at 17 49 20-fullpage | Screen Shot 2022-06-21 at 17 49 24-fullpage |

### Why is it needed?

To display proper fonts.

### Related issues

- Fixes https://github.com/strapi/design-system/issues/629

### DIFF
--- a/packages/strapi-design-system/src/ThemeProvider/ThemeProvider.js
+++ b/packages/strapi-design-system/src/ThemeProvider/ThemeProvider.js
@@ -153,7 +153,7 @@ const GlobalStyle = createGlobalStyle`
     height: 100%;
   }
   body {
-    font-family: --apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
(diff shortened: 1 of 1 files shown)


ronronscelestes (approved):
nice catch, thank you!

gu-stav:
@ronronscelestes @HichamELBSI Thanks for the quick reviews 🚀 I'll keep this PR open till ~ 11:00 tomorrow morning so that @maevalienard has a chance to have a look, since the rendering differs quite a bit and I'm wondering why this hasn't been noticed by looking at Figma <> CSS yet.

maevalienard (approved):
Good catch! Thank you 😄

PR #636, opened 2022-07-05
Add MIT license file

### LINKED ISSUE #619: license assignment
This work is licensed under what license? Could you add it to the project's codebase?

Thanks

gu-stav:
### What does it do?

Adds a license file. The license is already mentioned in the [`package.json`](https://github.com/strapi/design-system/blob/main/package.json), but the file was missing from the repo.

### Why is it needed?

To avoid users confusion, under which license the project is published.

### Related issue(s)/PR(s)

- Fixes #619

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,22 @@
+Copyright (c) 2015-present Strapi Solutions SAS
+
+Portions of the Strapi software are licensed as follows:
+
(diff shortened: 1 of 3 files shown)


alexandrebodin (commented):
I think we could use a plain MIT license and not the monorepo one for the time being as we don't plan to add ee features to the design system

PR #643, opened 2022-07-08
[Breadcrumbs update] HeaderLayout subtitle node support

ronronscelestes:
## What

- Extending HeaderLayout subtitle type support needed for ML folders

## Snapshot

image

### DIFF
--- a/packages/strapi-design-system/src/Layout/HeaderLayout.js
+++ b/packages/strapi-design-system/src/Layout/HeaderLayout.js
@@ -63,6 +63,8 @@ const StickyBox = styled(Box)`
 
 export const BaseHeaderLayout = React.forwardRef(
   ({ navigationAction, primaryAction, secondaryAction, subtitle, title, sticky, width, ...props }, ref) => {
+    const isSubtitleString = typeof subtitle === 'string';
+
     if (sticky) {
       return (
         <StickyBox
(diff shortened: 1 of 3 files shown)


packages/strapi-design-system/src/Layout/HeaderLayout.js
```diff
@@ -81,9 +83,13 @@ export const BaseHeaderLayout = React.forwardRef(
                 <Typography variant="beta" as="h1" {...props}>
                   {title}
```
ronronscelestes, line 84: I was wondering about allowing the title to be a node too. My only worry is that so far we make sure we have an `h1` on the page, which wouldn't be the case allowing a node, wdyt @gu-stav?

gu-stav: I'd say the design-system should be as strict as possible. If we don't need to open up that slot for nodes, I'd keep it as is.

packages/strapi-design-system/src/Layout/HeaderLayout.js
```diff
@@ -137,7 +147,7 @@ BaseHeaderLayout.propTypes = {
   primaryAction: PropTypes.node,
   secondaryAction: PropTypes.node,
   sticky: PropTypes.bool,
-  subtitle: PropTypes.string,
+  subtitle: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
gu-stav, line 150: Could you add a comment:

```
// TODO V2: Remove the string fallback
```

to make sure we remove that in the next major iteration?

gu-stav (approved):
Great, thanks!

PR #644, opened 2022-07-09
Added hook for value change and changed conditions for 'empty' value

not-vainglorious:
### What does it do?

Fixes [strapi/design-system#658](https://github.com/strapi/design-system/issues/658).

### Why is it needed?

To correctly handle changes outside of component (when "value" props changes). Display empty field when you get null values as response from strapi

### How to test it?

Only testable with some code which changes value outside of component. Can add example with buttons to storybook if requested.

### Related issue(s)/PR(s)

- Refs https://github.com/strapi/design-system/issues/658


not-vainglorious:
Also moving all "value" logic (checks and changes) to custom hook seems reasonable to me. Let me know and i can do it if you need.

packages/strapi-design-system/src/NumberInput/NumberInput.js
```diff
@@ -51,11 +51,22 @@ export const NumberInput = React.forwardRef(
...
     const numberFormaterRef = useRef(new NumberFormatter(getDefaultLocale()));
+    const inputRef = useRef();
 
     if (!label && !props['aria-label']) {
       throw new Error('The NumberInput component needs a "label" or an "aria-label" props');
     }
 
+    useEffect(() => {
+      if (inputRef.current) {
+        setInputValue(value === undefined || value === null ? INITIAL_VALUE : String(value));
```
ronronscelestes, line 62: I think we would need more value verification before updating `inputValue` 🤔
for example I used a button to update the value to a string: 

image

gu-stav, line 62: What do you think about using:

```js
const getValueWithDefaults = (value) => (value === undefined || value === null ? INITIAL_VALUE : String(value));

// ...

useEffect(() => {
  setInputValue(getValueWithDefaults(value));
}, [value]);
```

instead?

packages/strapi-design-system/src/NumberInput/NumberInput.js
```diff
@@ -149,13 +160,13 @@ export const NumberInput = React.forwardRef(
     };
 
     const handleFocus = () => {
-      if (value !== undefined) {
+      if (value !== undefined && value !== null) {
```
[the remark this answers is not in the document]
not-vainglorious: Just checked this case. Its not possible to achieve null in strapi due to [this code](https://github.com/strapi/strapi/blob/b0d73be3e776b4dfde0251b8befe5a4a63a960f8/packages/core/helper-plugin/lib/src/components/GenericInput/index.js#L52). In strapi component will always have `undefined` instead of `null`.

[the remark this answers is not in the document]
not-vainglorious: After spending some time on research i got following.

`NumberInput` is used 3 times in strapi/strapi repo:

1. [`GenericInput`](https://github.com/strapi/strapi/blob/master/packages/core/helper-plugin/lib/src/components/GenericInput/index.js) with [`const value = defaultValue ?? undefined;`](https://github.com/strapi/strapi/blob/master/packages/core/helper-plugin/lib/src/components/GenericInput/index.js#L52) so the value always be **undefined**.
2. [`Inputs`](https://github.com/strapi/strapi/blob/master/packages/core/helper-plugin/lib/src/components/FilterPopoverURLQuery/Inputs.js) which is used only in [`FilterPopoverURLQuery`](https://github.com/strapi/strapi/blob/master/packages/core/helper-plugin/lib/src/components/FilterPopoverURLQuery/index.js). `Inputs` component is not handling nullish values itself but `FilterPopoverURLQuery` default value is empty string ([default state](**url**), [value passed directly to `Inputs`](https://github.com/strapi/strapi/blob/master/packages/core/helper-plugin/lib/src/components/FilterPopoverURLQuery/index.js#L167)). Also you should notice that you got two components named `Inputs` but this one is used only in `FilterPopoverURLQuery`. You can check it by typing path like this in global project search `FilterPopoverURLQuery/Inputs` because there is no any other way to import `Inputs` component from _FilterPopoverURLQuery folder_.
3. [`CheckboxWithNumberField`](https://github.com/strapi/strapi/blob/master/packages/core/helper-plugin/lib/src/components/FilterPopoverURLQuery/Inputs.js) with [0 value by default](https://github.com/strapi/strapi/blob/master/packages/core/helper-plugin/lib/src/components/FilterPopoverURLQuery/Inputs.js#L83).


**Conclusion**.
Seems like there is no way that `NumberInput` will get `null` value passed into props. I`m not part of core team and can make a mistake and I provided full information so you can check this out by yourself fast enough.

packages/strapi-design-system/src/NumberInput/NumberInput.js
not-vainglorious: Actually its not that easy for me as for back-end dev (with angular experience). Maybe you can give me an advice. 

For validation we got `numberParserRef` but we can't use it in `useEffect(..., [])` and on initial call ref is still not initialized but we can't just check if its initialized or not (with `if (numberParserRef.current)`) and should wait for init so starting value will be checked and converted correctly. 

So i can spend time on weekend to resolve this or do not fix it now, open new issue and let core team fix it later on.

packages/strapi-design-system/src/NumberInput/NumberInput.js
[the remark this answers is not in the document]
gu-stav: Thanks for checking. I agree with your analysis and also think it will never receive `null`. If so, it is a bug.

packages/strapi-design-system/src/NumberInput/NumberInput.js
ronronscelestes: hi again 🌞 sorry for the delay!
I worked on finding a fix this morning - we have access to `numberParserRef` in the `useEffect` but since `value` is received as a number instead of a string that's why it seems like not working, the value needs `value.toString()` transformation before being passed to `numberParserRef.current.isValidPartialNumber()` function.
I think it would be great to have a function able to handle the change both for `handleChange` and the `useEffect` - I could try to push on your branch or let you try to handle it as you prefer 😊

packages/strapi-design-system/src/NumberInput/__tests__/NumberInput.e2e.js
```diff
@@ -45,6 +45,7 @@ test.describe.parallel('NumberInput', () => {
       });
 
       test('puts the step value in the input when pressing ArrowUp and that the input is empty', async ({ page }) => {
+        await page.fill('input', '');
```
not-vainglorious, line 48: @ronronscelestes

Also its a bit strange that this methods was removed from tests.
Default iframe for this tests is `iframe.html?id=design-system-components-numberinput--base&viewMode=story`
[Link to vercel](https://design-system-git-main-strapijs.vercel.app/iframe.html?id=design-system-components-numberinput--base&viewMode=story)

Don't know how is even possible to [navigate before each test to this iframe](https://github.com/strapi/design-system/blob/main/packages/strapi-design-system/src/NumberInput/__tests__/NumberInput.e2e.js#L10) with default value of `3.14159265359` and after get `1` after increment. I think it should be fixed this way. Also we can use other iframe (any other empty) for this tests.

Actually i would like to suggest you creating new 'describe' with goto to [this one](https://design-system-git-main-strapijs.vercel.app/iframe.html?id=design-system-components-numberinput--with-initial-empty&viewMode=story).

If you are OK with that just tell me and i`ll commit changes.

packages/strapi-design-system/src/NumberInput/__tests__/NumberInput.e2e.js
```diff
@@ -53,6 +53,7 @@ test.describe.parallel('NumberInput', () => {
...
```
not-vainglorious, line 56: @ronronscelestes

Also its a bit strange that this methods was removed from tests.
Default iframe for this tests is iframe.html?id=design-system-components-numberinput--base&viewMode=story
[Link to vercel](https://design-system-git-main-strapijs.vercel.app/iframe.html?id=design-system-components-numberinput--base&viewMode=story)

Don't know how is even possible to [navigate before each test to this iframe](https://github.com/strapi/design-system/blob/main/packages/strapi-design-system/src/NumberInput/__tests__/NumberInput.e2e.js#L10) with default value of 3.14159265359 and after get 1 after increment. I think it should be fixed this way. Also we can use other iframe (any other empty) for this tests.

Actually i would like to suggest you creating new 'describe' with goto to [this one](https://design-system-git-main-strapijs.vercel.app/iframe.html?id=design-system-components-numberinput--with-initial-empty&viewMode=story).

If you are OK with that just tell me and i`ll commit changes.

packages/strapi-design-system/src/NumberInput/__tests__/NumberInput.e2e.js
```diff
@@ -53,6 +53,7 @@ test.describe.parallel('NumberInput', () => {
...
         await page.focus('input');
         await page.fill('input', '1,');
```
not-vainglorious, line 58: And also according to test case input should be empty, isn't it?

packages/strapi-design-system/src/NumberInput/NumberInput.stories.mdx
```diff
@@ -90,7 +91,7 @@ NumberInputs can be used without a label.
             name="content"
             hint="Description line"
             error={undefined}
-            onValueChange={(value) => setContent(value)}
+            onValueChange={(value) => { console.log('trigger'); setContent(value); }}
```
not-vainglorious, line 94: I should remove console.log

joshuaellis:
closing in favour of #773.

Thank you for all the work you put in to this PR.

PR #646, opened 2022-07-12
[Breadcrumbs update] Breadcrumb v2 & CrumbLink

ronronscelestes:
## What

- Updated Breadcrumbs v1 style
- Breadcrumbs & Crumb v2
- CrumbLink

## Snapshots

image

image

### DIFF
--- a/packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
@@ -9,8 +9,8 @@ import { VisuallyHidden } from '../VisuallyHidden';
 
 const CrumbWrapper = styled(Flex)`
   svg {
-    height: 10px;
-    width: 10px;
+    height: ${10 / 16}rem;
+    width: ${10 / 16}rem;
     path {
       fill: ${({ theme }) => theme.colors.neutral500};
     }
@@ -30,7 +30,7 @@ export const Crumb = ({ children }) => {
       <Typography variant="pi" textColor="neutral600">
         {children}
       </Typography>
-      <Box paddingLeft={3} paddingRight={3}>
+      <Box aria-hidden paddingLeft={3} paddingRight={3}>
         <ChevronRight />
       </Box>
     </CrumbWrapper>
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
@@ -2,6 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Flex } from '../../Flex';
 import { VisuallyHidden } from '../../VisuallyHidden';
+import { Crumb } from './Crumb';
+import { CrumbLink } from './CrumbLink';
 
 export const Breadcrumbs = ({ label, children, ...props }) => {
   return (
@@ -12,8 +14,10 @@ export const Breadcrumbs = ({ label, children, ...props }) => {
   );
 };
 
+const crumbType = PropTypes.shape({ type: PropTypes.oneOf([Crumb, CrumbLink]) });
+
 Breadcrumbs.displayName = 'Breadcrumbs';
 Breadcrumbs.propTypes = {
-  children: PropTypes.node.isRequired,
+  children: PropTypes.oneOfType([PropTypes.arrayOf(crumbType), crumbType]).isRequired,
   label: PropTypes.string.isRequired,
 };
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
@@ -1,14 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Flex } from '../../Flex';
-import { VisuallyHidden } from '../../VisuallyHidden';
 import { Crumb } from './Crumb';
 import { CrumbLink } from './CrumbLink';
 
 export const Breadcrumbs = ({ label, children, ...props }) => {
   return (
-    <Flex {...props}>
-      <VisuallyHidden>{label}</VisuallyHidden>
+    <Flex aria-label={label} {...props}>
       <ol aria-hidden={true}>{children}</ol>
     </Flex>
   );
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
@@ -1,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Flex } from '../../Flex';
+import { VisuallyHidden } from '../../VisuallyHidden';
+
+export const Breadcrumbs = ({ label, children, ...props }) => {
+  return (
+    <Flex {...props}>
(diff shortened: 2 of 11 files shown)


packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,23 @@
...
+import PropTypes from 'prop-types';
+import { Flex } from '../../Flex';
+import { VisuallyHidden } from '../../VisuallyHidden';
+import { Crumb } from './Crumb';
+import { CrumbLink } from './CrumbLink';
+
+export const Breadcrumbs = ({ label, children, ...props }) => {
+  return (
+    <Flex {...props}>
+      <VisuallyHidden>{label}</VisuallyHidden>
```
gu-stav, line 11: What do you think about making this an `aria-label` or the `Flex` container? This would allow us to write:

```
<nav aria-label="Folder Structure Navigation" />
```

is is a very nice pattern for screenreader users to discover the structure of a page.

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,23 @@
...
+      <ol aria-hidden={true}>{children}</ol>
+    </Flex>
+  );
+};
+
+const crumbType = PropTypes.shape({ type: PropTypes.oneOf([Crumb, CrumbLink]) });
+
+Breadcrumbs.displayName = 'Breadcrumbs';
```
gu-stav, line 19: What do you think about defining the const first and export it later, to avoid the duplication of the name?

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.stories.mdx
```diff
@@ -0,0 +1,97 @@
...
+import { Crumb } from './Crumb';
+import { CrumbLink } from './CrumbLink';
+import { Stack } from '../../Stack';
+import CollectionType from '@strapi/icons/CollectionType';
+
+<Meta title="Design System/Components/v2/Breadcrumbs" component={Breadcrumbs} />
+
+# Breadcrumbs
+
+Breadcrumbs are a list of links that inform users of their navigation.
```
gu-stav, line 15:
```suggestion
Breadcrumbs are a list of hierarchical links that inform users about their possible navigation path upwards.
```

packages/strapi-design-system/src/v2/Breadcrumbs/Crumb.js
```diff
@@ -0,0 +1,42 @@
...
+import styled from 'styled-components';
+import { Flex } from '../../Flex';
+import { Box } from '../../Box';
+import { Typography } from '../../Typography';
+import ChevronRight from '@strapi/icons/ChevronRight';
+
+const CrumbWrapper = styled(Flex)`
+  svg {
+    height: 10px;
+    width: 10px;
```
gu-stav, line 12: 🤔 `rem`?

packages/strapi-design-system/src/v2/Breadcrumbs/Crumb.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+`;
+
+export const Crumb = ({ children }) => {
+  return (
+    <CrumbWrapper inline as="li">
+      <Typography variant="pi" textColor="neutral600">
+        {children}
+      </Typography>
+      <Box paddingLeft={3} paddingRight={3}>
```
gu-stav, line 32: Since the icon is purely decorative it should be `aria-hidden`.

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+  a {
+    text-decoration: none;
+    color: ${({ theme }) => theme.colors.neutral600};
+    font-size: ${({ theme }) => theme.fontSizes[1]};
+    line-height: ${({ theme }) => theme.lineHeights[3]};
+    padding: ${({ theme }) => theme.spaces[1]};
+    border-radius: ${({ theme }) => theme.borderRadius};
+    :hover {
```
gu-stav, line 24: `:focus` is missing.

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
```diff
@@ -0,0 +1,58 @@
...
+      background: ${({ theme }) => theme.colors.neutral200};
```
gu-stav, line 25: Let's be explicit here and use `background-color`.

gu-stav (changes requested):
Added a few comments to discuss. Thanks for working on it! I think in `@strapi/strapi` we try to order CSS attributes alphabetically, which I'd like to see here too to make it easier to understand what is being done and what isn't. What do you think?

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.stories.mdx
ronronscelestes: for the story's wording I pinged @maevalienard 
she will handle it 😊

maevalienard: Well seems like it's done already so let's make profit of the time already spent :D I'll do it next time! :)

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
gu-stav: I think it wouldn't be an issue - `div`s can reiceive an `arial-label` too. It's just less semantical.

packages/strapi-design-system/src/Breadcrumbs/Breadcrumbs.js
```diff
@@ -11,19 +11,23 @@ const CrumbWrapper = styled(Flex)`
...
+      fill: ${({ theme }) => theme.colors.neutral500};
+    }
   }
   :last-of-type ${Box} {
     display: none;
   }
+  :last-of-type ${Typography} {
+    color: ${({ theme }) => theme.colors.neutral800};
+    font-weight: ${({ theme }) => theme.fontWeights.bold};
+  }
```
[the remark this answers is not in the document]
ronronscelestes: mmh not sure I understand "`we are moving knowledge about the actual Crumb into the parent level`"
this is the v1 component and this styling is done on `Crumb` component, not on `Breadcrumbs`.
this pattern was already there before my changes when removing the last icon here:
image

about updating this logic by passing a prop to `Crumb`, why not but since it would change the API (if you don't pass isLast then you would have the last icon appearing) I don't think we should do it for Breadcrumbs v1 component.
I could implement it for the v2 one though, wdyt?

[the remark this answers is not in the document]
gu-stav: Ah, my bad then. Yeah, for the v2 version that would be great :)

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
gu-stav: Are we? Then never mind ...

ronronscelestes:
Implemented here #653

PR #649, opened 2022-07-12
Fix e2e:watch command using chokidar-cli

gu-stav:
### What does it do?

Fixes the `test:e2e:watch` command. It was broken since the migration to the playwright test runner and is a very convenient tool, when having to work with the e2e tests. The solution has been mentioned in https://github.com/microsoft/playwright/issues/7035#issuecomment-858430649

### Why is it needed?

Developer experience.

### How to test it?

1. Run `yarn test:e2e:watch`
2. Change something in an e2e test file
3. Watch only the tests from that file being re-run

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
   },
   "devDependencies": {
     "@specifyapp/cli": "^0.0.51",
+    "chokidar-cli": "^3.0.0",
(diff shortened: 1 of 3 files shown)


ronronscelestes (approved):
Works well on my side, thanks 🌞

PR #652, opened 2022-07-14
Accordion: Make height of AccordionToggle flexbile and add title ellipsis

gu-stav:
### What does it do?

This PR does a couple of things:

- it ellipsizes the title
  -  to fix a bug mentioned in https://github.com/strapi/strapi/pull/12500#discussion_r817439733 and allows us to pass any string into the title without breaking the component
  - to fix a bug we currently have on repeatable components Screenshot 2022-07-14 at 18 34 42

- it refactors the component, to reduce it's size, since most of the component code was duplicated
- it reduces the explicit `height` and makes it a `min-height` to allow text to flow more naturally

### Why is it needed?

Housekeeping and fixes a bug.

### How to test it?

|Before|After|
|-|-|
| Screen Shot 2022-07-14 at 18 27 17-fullpage | Screen Shot 2022-07-14 at 18 27 07-fullpage |

[Preview](https://design-system-git-enhancement-accordion-flexible-strapijs.vercel.app/?path=/story/design-system-components-accordion--base)

### Related issue(s)/PR(s)

- Refs https://github.com/strapi/strapi/pull/12500

### DIFF
--- a/packages/strapi-design-system/src/Accordion/AccordionToggle.js
+++ b/packages/strapi-design-system/src/Accordion/AccordionToggle.js
@@ -14,6 +14,11 @@ import { getBackground } from './utils';
 const ToggleButton = styled(TextButton)`
   text-align: left;
 
+  // necessary to make the ellipsis prop work on the title
+  > span {
+    max-width: 100%;
+  }
+
   svg {
     width: ${14 / 16}rem;
     height: ${14 / 16}rem;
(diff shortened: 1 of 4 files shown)


maevalienard (approved):
Good catch once again, thank you :D

ronronscelestes (approved):
Nice improvements 🌞

PR #651, opened 2022-07-13
[Breadcrumbs] CrumbSimpleMenu

ronronscelestes:
⚠️ To merge after #646

## What

CrumbSimpleMenu component supporting SimpleMenu usage in Breadcrumbs

PR #653, opened 2022-07-15
Breadcrumbs: Implement v2 version

gu-stav:
### What does it do?

This is an alternative solution & implementation of https://github.com/strapi/design-system/pull/646 and https://github.com/strapi/design-system/pull/651 with a few advantages:

- the single `Crumb*` components are much simpler, because
- the `Divider` logic was abstracted at the `Breadcrumbs` level
- it works with much less style overwrites

I didn't fully understand the API for the other PR - this one follows the idea of:

- `CrumbLink`: is used to render a link
- `Crumb`: is used to render a non-link and can receive an `aria-current` prop, if it is the last one
- `CrumbSimpleMenu`: acts as an alias for `SimpleMenu` with meaningful defaults

[Preview](https://design-system-git-enhancement-breadcrumbs-simplified-strapijs.vercel.app/?path=/story/design-system-components-v2-breadcrumbs--with-menu)

### Why is it needed?

It is just a different implementation.

### DIFF
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
@@ -1,5 +1,6 @@
 import React, { Children } from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 
 import { Box } from '../../Box';
 import { Crumb } from './Crumb';
@@ -8,10 +9,20 @@ import { CrumbSimpleMenu } from './CrumbSimpleMenu';
 import { Divider } from './Divider';
 import { Flex } from '../../Flex';
 
+const AlignedList = styled(Flex)`
+  // CrumbLinks do have padding-x, because they need to have a
+  // interaction effect, which mis-aligns the breadcrumbs on the left.
+  // This normalizes the behavior by moving the first item to left by
+  // the same amount it has inner padding
+  :first-child {
+    margin-left: ${({ theme }) => `calc(-1*${theme.spaces[2]})`};
+  }
+`;
+
 export const Breadcrumbs = ({ label, children, ...props }) => {
   return (
     <Box aria-label={label} {...props}>
-      <Flex as="ol" horizontal>
+      <AlignedList as="ol" horizontal>
         {Children.map(children, (child, index) => {
           const isLast = index + 1 === children.length;
 
@@ -22,7 +33,7 @@ export const Breadcrumbs = ({ label, children, ...props }) => {
             </Flex>
           );
         })}
-      </Flex>
+      </AlignedList>
     </Box>
   );
 };
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
@@ -1,0 +1,42 @@
+import React, { Children } from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+import { Box } from '../../Box';
+import { Crumb } from './Crumb';
+import { CrumbLink } from './CrumbLink';
+import { CrumbSimpleMenu } from './CrumbSimpleMenu';
+import { Divider } from './Divider';
+import { Flex } from '../../Flex';
+
+const ListFlex = styled.ol`
+  display: flex;
+`;
+
+export const Breadcrumbs = ({ label, children, ...props }) => {
+  return (
+    <Box aria-label={label} {...props}>
+      <ListFlex aria-hidden={true}>
+        {Children.map(children, (child, index) => {
+          const isLast = index + 1 === children.length;
+
+          return (
+            <Flex inline as="li">
+              {child}
+              {!isLast && <Divider />}
+            </Flex>
+          );
+        })}
+      </ListFlex>
+    </Box>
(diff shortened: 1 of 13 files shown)


packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <Box aria-label={label} {...props}>
+      <ListFlex aria-hidden={true}>
+        {Children.map(children, (child, index) => {
+          const isLast = index + 1 === children.length;
+
+          return (
+            <Flex inline as="li">
+              {child}
+              {!isLast && <Divider />}
```
gu-stav, line 26: @HichamELBSI This approach should make it fairly easy to make the divider a prop once it is needed.

packages/strapi-design-system/src/v2/Breadcrumbs/Crumb.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Typography } from '../../Typography';
+
+export const Crumb = ({ children, ...props }) => (
+  <Typography variant="pi" textColor="neutral800" fontWeight="bold" {...props}>
```
ronronscelestes, line 7: Crumb shouldn't be always bold
The isLast should be bold regardless of which component it is (Crumb, CrumbLink, CrumbSimpleMenu

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
```diff
@@ -0,0 +1,18 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import { SimpleMenu } from '../SimpleMenu';
+
+import CarretDown from '@strapi/icons/CarretDown';
+
+export const CrumbSimpleMenu = ({ children, ...props }) => (
+  <SimpleMenu noBorder icon={<CarretDown />} size="S" {...props}>
```
ronronscelestes, line 9: SimpleMenu should be `as={IconButton}` and then height of the IconButton reduced

ronronscelestes:
```
const IconButtonCustom = styled(IconButton)`
  height: ${({ theme }) => theme.spaces[3]};
  background-color: transparent;
`;

...


<SimpleMenu as={IconButtonCustom} noBorder icon={<CarretDown />} size="S" {...props}>
```

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
```diff
@@ -0,0 +1,42 @@
...
+
+export const Breadcrumbs = ({ label, children, ...props }) => {
+  return (
+    <Box aria-label={label} {...props}>
+      <ListFlex aria-hidden={true}>
+        {Children.map(children, (child, index) => {
+          const isLast = index + 1 === children.length;
+
+          return (
+            <Flex inline as="li">
```
ronronscelestes, line 24:
```
const Li = styled(Flex)`
  span,
  a {
    color: ${({ theme, isLast }) => (isLast ? theme.colors.neutral800 : theme.colors.neutral600)};
    font-weight: ${({ theme, isLast }) => (isLast ? theme.fontWeights.bold : theme.fontWeights.regular)};
  }
`;

...

<Li isLast={isLast} inline as="li">
```

then we can remove the styling from Crumb/CrumbLink?

ronronscelestes (changes requested):
thank you for working on it!

packages/strapi-design-system/src/v2/Breadcrumbs/Crumb.js
gu-stav: Done in d6f30e4fcfd97bdf2fb73214247ccb5c6ca53a3f

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
gu-stav: I'm already usign the `size="S"` variant, which visually works well for me, don't you think?

Screenshot 2022-07-18 at 08 58 34

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
gu-stav: I don't want to do this, for 2 reasons:

- it would make the last `CrumbLink` bold, which is not a behavior we want (the last `Crumb` should never be a link)
- it introduces a `span` element selector, which is brittle and might affect other elements
- it moves styling logic which belongs into the `Crumb` to the parent

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
gu-stav: We've switched to use slashes in e9e1036c4c2fa601ee99119cbd13775d31b81100

ronronscelestes (changes requested):
LGTM on the code side, I think there are snapshots conflicts to resolve before merging!

ronronscelestes (approved):

PR #655, opened 2022-07-18
Releasing

HichamELBSI:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "1.2.0",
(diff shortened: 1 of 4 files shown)


gu-stav (commented):
I've recently realized that we never update the `@strapi/icons` peer-dependency in the design-system package: https://github.com/strapi/design-system/blob/4cbed35da651e5300ca27cc10e61da2249c678d7/packages/strapi-design-system/package.json#L45

Do you think we should start doing that?

gu-stav (approved):
Thank you! 🚀

PR #656, opened 2022-07-19
Fix ToggleCheckbox value

### LINKED ISSUE #645: Boolean toggle randomly selects the wrong value from where you click

- Node.js version: v14
- NPM version: v6
- Strapi version: v4.2.2
- Database: SQLite3
- Operating system: Linux Mint 20.3

### Describe the bug

Sometimes when selecting a boolean value it will select the opposite value then where your mouse pointer is. This appears to be random per boolean field and it looks like it depends on if the boolean had a value before and you click clear

Loom video shared internally as it contain's customer information. See internal TID1663 or my post in the product slack channel.

Short summary:
1. Create a content-type with a boolean field
2. Try to click on the boolean `false`
3. If the value is properly selected, hit clear
4. Try to click on `true`
5. Notice that the false value is selected instead

Alternatively repeat this logic selecting `true` first, clear, then select `false` if you cannot reproduce.

### Expected behavior

Boolean value should select the value your mouse is on

### Additional context

EE Customer, Internal TID1663

ronronscelestes:
## Issue

Steps to reproduce:
- Set the default value of ToggleCheckbox as true
- Clear the value
- Change/toggle value
- Value will be set to false

## Fix

- Clicking when value is null should change the value to true
- Fixed `Warning: A component is changing a controlled input to be uncontrolled. This is likely caused by the value changing from a defined to undefined, which should not happen` error message

fix #645

### DIFF
--- a/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
+++ b/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
@@ -68,8 +68,6 @@ export const ToggleCheckbox = React.forwardRef(
     const handleChange = (e) => {
       if (disabled) return;
 
-      console.log(e.target.checked);
-
       onChange(e);
     };
 
--- a/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
+++ b/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
@@ -68,6 +68,8 @@ export const ToggleCheckbox = React.forwardRef(
     const handleChange = (e) => {
       if (disabled) return;
 
+      console.log(e.target.checked);
+
       onChange(e);
     };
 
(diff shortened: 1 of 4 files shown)


gu-stav (approved):
Thank you for fixing this!

I'll leave it up to you, but what do you think about starting e2e tests on this component, because the toggle-logic becomes more and more complicated? Doesn't need to be big, but mainly testing the 3 states ...

ronronscelestes:
> Thank you for fixing this!
> 
> I'll leave it up to you, but what do you think about starting e2e tests on this component, because the toggle-logic becomes more and more complicated? Doesn't need to be big, but mainly testing the 3 states ...

was my first time on [e2e tests](https://github.com/strapi/design-system/pull/656/commits/f8a4c7427dc257d8408341fd04f79a09991054cc), let me know if it's look good to you @gu-stav  🌞

gu-stav (approved):
Very nice, thank you!

PR #657, opened 2022-07-19
Breadcrumbs: Update interaction colors

gu-stav:
### What does it do?

This PR updates the interaction colors for `CrumbLink`s and `CrumbSimpleMenu`s from `neutral100` to `neutral200`.

### Why is it needed?

These components are used on a background which is already `neutral100` and therefore the state won't be visible.

### How to test it?

Validate the `hover/ focus` background colors on https://design-system-git-fix-breadcrumbs-interaction-strapijs.vercel.app/?path=/story/design-system-components-v2-breadcrumbs--with-menu

### Related issue(s)/PR(s)

- https://github.com/strapi/strapi/pull/13788
- https://github.com/strapi/design-system/pull/653


ronronscelestes (approved):
LGTM

Issue #440, opened 2021-11-18
Remove peer dependency on react-router-dom


mfrachet:
When installing the DS on a nextjs project for instance and trying to use SubNav, it throws an error because it can't find the react-router-dom library (defined as a peer dependency)

SafaLabash:
I have the same problem.

PR #662, opened 2022-07-25
[Breadcrumbs] PropTypes & Crumb padding

ronronscelestes:
## What

- Fixed some proptypes errors:
  - Breadcrumbs did not accept a component that returns CrumbSimpleMenu (ex: CrumbSimpleMenuAsync for ML Breadcrumbs) + same issue for SimpleMenu components. Opened children Proptypes to `PropTypes.node`
  - CrumbLink Proptypes pattern same as other DS components based on BaseLink
  - SimpleMenu v1 & v2 + Breadcrumbs children proptypes
- Fixed Crumb not aligned because of different padding than `CrumbLink` and `CrumbSimpleMenu` and first child `margin-left`

## Snapshot

Before
Screenshot 2022-07-25 at 15 16 07

After
Screenshot 2022-07-25 at 15 19 09

### DIFF
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
@@ -3,9 +3,6 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
 import { Box } from '../../Box';
-import { Crumb } from './Crumb';
-import { CrumbLink } from './CrumbLink';
-import { CrumbSimpleMenu } from './CrumbSimpleMenu';
 import { Divider } from './Divider';
 import { Flex } from '../../Flex';
 
@@ -38,11 +35,9 @@ export const Breadcrumbs = ({ label, children, ...props }) => {
   );
 };
 
-const crumbType = PropTypes.shape({ type: PropTypes.oneOf([Crumb, CrumbLink, CrumbSimpleMenu]) });
-
 Breadcrumbs.displayName = 'Breadcrumbs';
 
 Breadcrumbs.propTypes = {
-  children: PropTypes.oneOfType([PropTypes.arrayOf(crumbType), crumbType]).isRequired,
+  children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node]).isRequired,
   label: PropTypes.string.isRequired,
 };
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
@@ -23,7 +23,14 @@ export const CrumbLink = ({ children, ...props }) => <StyledLink {...props}>{chi
 
 CrumbLink.displayName = 'CrumbLink';
 
+const requiredPropsCheck = (props, propName, componentName) => {
+  if (!props.href && !props.to && !props.onClick) {
+    return new Error(`One of 'href', 'to' or 'onClick' is required by '${componentName}' component.`);
+  }
+};
+
 CrumbLink.propTypes = {
   children: PropTypes.string.isRequired,
-  href: PropTypes.string.isRequired,
+  href: requiredPropsCheck,
+  to: requiredPropsCheck,
 };
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
@@ -32,5 +32,6 @@ const requiredPropsCheck = (props, propName, componentName) => {
 CrumbLink.propTypes = {
   children: PropTypes.string.isRequired,
   href: requiredPropsCheck,
+  onClick: requiredPropsCheck,
   to: requiredPropsCheck,
 };
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
@@ -23,15 +23,11 @@ export const CrumbLink = ({ children, ...props }) => <StyledLink {...props}>{chi
 
 CrumbLink.displayName = 'CrumbLink';
 
-const requiredPropsCheck = (props, propName, componentName) => {
-  if (!props.href && !props.to && !props.onClick) {
-    return new Error(`One of 'href', 'to' or 'onClick' is required by '${componentName}' component.`);
-  }
+CrumbLink.defaultProps = {
+  to: undefined,
 };
 
 CrumbLink.propTypes = {
   children: PropTypes.string.isRequired,
-  href: requiredPropsCheck,
-  onClick: requiredPropsCheck,
-  to: requiredPropsCheck,
+  to: PropTypes.string,
 };
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
@@ -3,7 +3,7 @@ import React from 'react';
 import styled from 'styled-components';
 
 import { Button } from '../../Button';
-import { SimpleMenu } from '../SimpleMenu';
+import { SimpleMenu, MenuItem } from '../SimpleMenu';
 
 import CarretDown from '@strapi/icons/CarretDown';
 
@@ -22,9 +22,11 @@ export const CrumbSimpleMenu = ({ children, ...props }) => (
   </SimpleMenu>
 );
 
+const menuItemType = PropTypes.shape({ type: PropTypes.oneOf([MenuItem]) });
+
 CrumbSimpleMenu.displayName = 'CrumbSimpleMenu';
 
 CrumbSimpleMenu.propTypes = {
   'aria-label': PropTypes.string.isRequired,
-  children: PropTypes.arrayOf(SimpleMenu),
+  children: PropTypes.oneOfType([PropTypes.arrayOf(menuItemType), menuItemType]).isRequired,
 };
--- a/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
+++ b/packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
@@ -3,7 +3,7 @@ import React from 'react';
 import styled from 'styled-components';
 
 import { Button } from '../../Button';
-import { SimpleMenu, MenuItem } from '../SimpleMenu';
+import { SimpleMenu } from '../SimpleMenu';
 
 import CarretDown from '@strapi/icons/CarretDown';
 
@@ -22,11 +22,9 @@ export const CrumbSimpleMenu = ({ children, ...props }) => (
   </SimpleMenu>
 );
 
-const menuItemType = PropTypes.shape({ type: PropTypes.oneOf([MenuItem]) });
-
 CrumbSimpleMenu.displayName = 'CrumbSimpleMenu';
 
 CrumbSimpleMenu.propTypes = {
   'aria-label': PropTypes.string.isRequired,
-  children: PropTypes.oneOfType([PropTypes.arrayOf(menuItemType), menuItemType]).isRequired,
+  children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node]).isRequired,
 };
--- a/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
+++ b/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
@@ -227,15 +227,13 @@ SimpleMenu.defaultProps = {
 
 SimpleMenu.displayName = 'SimpleMenu';
 
-const menuItemType = PropTypes.shape({ type: PropTypes.oneOf([MenuItem]) });
-
 SimpleMenu.defaultProps = {
   size: 'M',
 };
 
 SimpleMenu.propTypes = {
   as: PropTypes.any,
-  children: PropTypes.oneOfType([PropTypes.arrayOf(menuItemType), menuItemType]).isRequired,
+  children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node]).isRequired,
   id: PropTypes.string,
   label: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.element]).isRequired,
   onClose: PropTypes.func,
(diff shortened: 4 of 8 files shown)


packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
```diff
@@ -38,11 +35,9 @@ export const Breadcrumbs = ({ label, children, ...props }) => {
   );
 };
 
-const crumbType = PropTypes.shape({ type: PropTypes.oneOf([Crumb, CrumbLink, CrumbSimpleMenu]) });
-
 Breadcrumbs.displayName = 'Breadcrumbs';
 
 Breadcrumbs.propTypes = {
-  children: PropTypes.oneOfType([PropTypes.arrayOf(crumbType), crumbType]).isRequired,
+  children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node]).isRequired,
```
ronronscelestes, line 41: I'm not a fan of removing component type check but this is preventing us to create a CrumbSimpleMenuAsync component and passing it to Breadcrumbs without error, I'm open to other ideas 😁

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
```diff
@@ -22,9 +22,11 @@ export const CrumbSimpleMenu = ({ children, ...props }) => (
...
 );
 
+const menuItemType = PropTypes.shape({ type: PropTypes.oneOf([MenuItem]) });
+
 CrumbSimpleMenu.displayName = 'CrumbSimpleMenu';
 
 CrumbSimpleMenu.propTypes = {
   'aria-label': PropTypes.string.isRequired,
-  children: PropTypes.arrayOf(SimpleMenu),
+  children: PropTypes.oneOfType([PropTypes.arrayOf(menuItemType), menuItemType]).isRequired,
```
ronronscelestes, line 31: Weird issue here:
- In storybook it works well
- When using it in Strapi we are still having an issue saying "_wrong children passed to CrumbSimpleMenu + SimpleMenu_", no matter which MenuItem version used (v1 or v2)
- I tried to allow both version but since I have to rename at least one MenuItem component with `as` it still considers it as a different component

Any idea is welcomed - error will keep appear in Strapi without fixing it.
Other solution:
Accepting any node as for Breadcrumbs children but 🤷🏻‍♀️

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
```diff
@@ -23,7 +23,14 @@ export const CrumbLink = ({ children, ...props }) => <StyledLink {...props}>{chi
...
+  if (!props.href && !props.to && !props.onClick) {
+    return new Error(`One of 'href', 'to' or 'onClick' is required by '${componentName}' component.`);
+  }
+};
+
 CrumbLink.propTypes = {
   children: PropTypes.string.isRequired,
-  href: PropTypes.string.isRequired,
+  href: requiredPropsCheck,
+  to: requiredPropsCheck,
```
gu-stav, line 35: I'm wondering if we should get rid of this for the v2 version. I believe this comes from the `Link` refactoring, but we should either unify it, or use the v1, where we can not at the moment.

What do you think?

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
gu-stav, line 41: I'm wondering if the initial definition was wrong and if this would be a better and working fix:

```suggestion
  children: PropTypes.arrayOf(PropTypes.oneOfType([Crumb, CrumbLink, CrumbSimpleMenu])).isRequired,
```

We should be able to make it work without opening it to `node`, as it works for the `SimpleMenu` too.

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
ronronscelestes: Do you mean remove href/to/onClick to be required at all?
I added this fix because saying that href is required doesn't work:
- We can use as={NavLink} for this component, which needs a to prop and not an href which caused proptypes errors
- We need to use this component with as="button" sometimes (e.g. Media Library in CM, we are using a button to change folder and not a link), in this case there won't be href or to prop, only an onClick one
- It is really similar to FolderCard usage which can accept link or button (and link can be with href or to - React side)

maybe I'm missing something? 👀

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
ronronscelestes: I explained this in a comment you can see before your review ⬆️
I first tried the 
`children: PropTypes.arrayOf(PropTypes.oneOfType([Crumb, CrumbLink, CrumbSimpleMenu])).isRequired,`
solution but it doesn't help us in some cases like for ML Breadcrumbs
in there I created a CrumbSimpleMenuAsync component which returns CrumbSimpleMenu - when doing that we still have proptypes error since CrumbSimpleMenuAsync is not a valid component 🤷🏻‍♀️
but I agree I'm not a fan of just opening it to node - but if not we'll have proptypes errors 🥲

while checking the comments on my code I've written before your review could you also check this one because we still have a proptype error here 🙏 
image

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
gu-stav: Phew, I see and didn't think of the `as="button"` case. Could we remove the `href=""` prop then, if we know it won't be needed (I guess that was my mistake initially - I thought we are using href)?

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
gu-stav: > I tried to allow both version but since I have to rename at least one MenuItem component with as it still considers it as a different component

I think this might be the issue. Maybe we need another wrapper then, so that `Crumb`, `CrumbLink` and `CrumbSimpleMenu` are always a `div` (e.g.) and the `as` prop is applied to their children, so that we never change the way the actual container is rendered?

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
ronronscelestes: yes sure, I just thought href could be useful one day but we could update it then if needed 👌

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
ronronscelestes: I'm sorry I don't understand 🙈
MenuItem and Breadcrumbs don't really have the same issue I think.
If we keep saying to Breadcrumbs that it can only allow Crumb, CrumbLink, CrumbSimpleMenu components, then it will keep sending error if we need to add more logic as for CrumbSimpleMenuAsync no?

gu-stav: What I mean is: instead of applying the `as` prop on the top-level DOM node you could try to apply it on a child instead. E.g.

```js
const CrumbLink = (props) => <div><Box {...props} /></div>
```

I'm wondering if the `as` prop creates the issue here, so that the actual components are not recognized by the prop-types.

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
ronronscelestes: Now that I think about it `Link` v2 doesn't require either, do shouldn't make them required at all for consistency or require them in `Link` v2 component too? 
image

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
ronronscelestes: okay I see what you mean but I don't think it will fix the CrumbSimpleMenuAsync situation? 🤔
when I use CrumbSimpleMenuAsync I don't use as prop so I don't think that is the issue, except if I'm missing a piece of knowledge there? 

what I need to do is:
```
const CrumbSimpleMenuAsync = () => {
  // stuff happening here
  return <CrumbSimpleMenu />
}
```

and then:
```
const BreadcrumbsInML = () => {
  return 
  <Breadcrumbs>
    <CrumbSimpleMenuAsync />
  </Breadcrumbs>
}
```

without having Breadcrumbs complaining because I passed a children that is not CrumbSimpleMenu, CrumbLink or Crumb - are we talking about the same thing? 👀

gu-stav: @ronronscelestes I'm also guessing here, but there must be a solution to this problem other than opening up the type to `node` :) Let me do some research ...

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
gu-stav: What happens to a Link component without `href`? Does it render as a span or does it become an invalid element?

packages/strapi-design-system/src/v2/Breadcrumbs/Breadcrumbs.js
ronronscelestes: note: we did some pair-programming on this issue and didn't find a working solution - we will keep `node` type for now

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
ronronscelestes: I would say to keep one of them required, in any case I find it helpful for developers, wdyt?
about href I would still keep it since all DS link components can be used as external link, I would find it weird to force NavLink or button, wdyt?

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
ronronscelestes: @HichamELBSI would love your opinion on this one if you have some time, maybe you have some knowledge about why in Strapi this component would act like we are not passing the right component version 
is something happening during the build that would make it confusing for this component to have 2 MenuItem components? (which I would find weird but that's my only theory so far)

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
gu-stav: Phew, I just realized I'm lost on where we want to go regarding `Link` and `NavLink`. @HichamELBSI I'd need your opinion here and try to summarize:

- `CrumbLink` will always be used with `as={NavLink}` (`NavLink` refers to `react-router-dom`)
- because of that, `CrumbLink` needs a `to` and a `href` prop: `to` is used by `NavLink`, `href` is used by `Link@v2`
- I'd like to get rid of one of these, because it feels like we have an inconsistency which is present in many places and I feel like we should settle for `to` everywhere in v2, since this is what `react-router-dom` is using and I can only imagine a very small number of cases where we'd not that one

But I'm not sure and I'm wondering what your thoughts are ...

ronronscelestes: Just a thought after checking:
- `Link@v2` don't force developers to have an `href` prop because it can be a `NavLink`
- Btw BaseLink is offering a nice isExternal prop that is not showing in Links documentation 
- If we need to make any change on Links components I would say to do it in another PR since I'd like to move on quicker on Breacrumbs so I can also move on with the feature 🥲

- If you think @gu-stav `CrumbLink` shouldn't be use as an external link I can enforce only `to` or `onClick` prop (since it can be a button to as we can see in the ML modal in the CM)
- That won't prevent user to use BaseLink prop (`isExternal`) but they would have a prop error 🤷🏻‍♀️

Tbh I would like to move on with the PR, I'm not sure @HichamELBSI will have time to review it and I would like to finish the feature this sprint, wdyt?

gu-stav: I'm not feeling comfortable to make a final decision on that now with another opinion, tbh, because I'm not sure I understand what the underlying idea is and where we want to go with `v2` ...

In the end, what I'd like to have is a clean interface for the `CrumbLink` component, which doesn't leave room for for ambiguity (which imho having `to` and `href` creates). Once the component is out and released, we won't be able to take the decision back and will have to live it for a long time and therefore I'd argue a solid decision is important here.

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
HichamELBSI: We are having the same issue on Cloud. I personally think that we shouldn't restrict the children proptype and keep the default react children. If we need a children with specific properties, we can do some checks before cloning the children.

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
HichamELBSI: The v2 components were created precisely in order to get rid of the `react-router-dom` dependency. Some users don't/can't use `react-router-dom`. For example, on `NextJS`, we need the href property. That's why our links must be render html links without modification on native properties. This is what we recommend in our documentation. 

Screenshot 2022-07-27 at 11 07 42

https://github.com/strapi/design-system/issues/440

Does it make sense?

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.js
ronronscelestes: thank you for your answer @HichamELBSI!
to make sure I understand "we shouldn't restrict the children proptype": 
`SimpleMenu` should restrict its children to be `MenuItem` components only so we would have `children: PropTypes.node` ?

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.js
ronronscelestes: As discussed in private messages, for now CrumbLink will have the same PropTypes definition as every link based on BaseLink:
- Enforcing only `to` or `href` would prevent Next developers to use the component without PropTypes errors
- In some cases CrumbLink needs to be a button (e.g. Modal navigation)
- We could redefine better PropTypes definition for links component, to be discussed during a DS meeting

@gu-stav @HichamELBSI

gu-stav (approved):
Looks good to me. Thanks for working on it and your patience. Eventhough it doesn't like a big difference, I'm glad we were able to settle on `to` instead of both. 🌻 

Good job!

HichamELBSI (approved):

PR #664, opened 2022-07-26
DatePicker: remove default placeholder value

### LINKED ISSUE #663: Blank date picker's placeholder value is '1/1/1970'

- Node.js version: 16 lts
- NPM version: 8.x
- Strapi version: 4.2.3
- Database: PostgreSQL 13
- Operating system: macOS 12.4

### Describe the bug

Blank date picker always show initial value '1/1/1970'

### Expected behavior

Blank date picker's placeholder value may be better with empty string or 'mm/dd/yyyy'

### Screenshots

Screen Shot 2022-07-15 at 11 07 01 PM

gu-stav:
### What does it do?

This removes the default `placeholder` attribute value for the `DatePicker` component. As has been discussed with @maevalienard and suggested in #663. It also removes some explicit props and passes them through `props` instead.

### Why is it needed?

1. we don't have it on other input fields (consistency)
2. editors can not type into the field (yet) and therefore can not make a mistake
3. the format would need to be localized to be clear in every timezone

### How to test it?

Validate the default placeholder is gone in [the component preview](https://design-system-git-fix-datepicker-placeholder-strapijs.vercel.app/?path=/story/design-system-components-datepicker--base)

### Related issue(s)/PR(s)

- Fixes #663

### DIFF
--- a/packages/strapi-design-system/src/DatePicker/DatePicker.js
+++ b/packages/strapi-design-system/src/DatePicker/DatePicker.js
@@ -13,16 +13,12 @@ export const DatePicker = ({
   ariaLabel,
   initialDate,
   selectedDate,
-  size,
   onChange,
   label,
   selectedDateLabel,
   onClear,
   clearLabel,
   disabled,
-  minDate,
-  maxDate,
-  placeholder,
   id,
   ...props
 }) => {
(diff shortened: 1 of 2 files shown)


maevalienard (approved):
Thank you!

PR #635, opened 2022-06-30
Add indentation styling to SubNavLinkSection child elements

richorrichard:
### What does it do?

Improves visual hierarchy of sub nav panel by adjusting indentation of `<SubNavLinkSection>` child elements, and reducing overall `<SubNavLink>` top and bottom padding. 

WIP, but I believe it's complete, pending review.

### Related issue(s)/PR(s)

Related to https://github.com/strapi/design-system/issues/637

*Before*
Screen Shot 2022-06-30 at 11 21 44 AM

*After*
Screen Shot 2022-06-30 at 11 20 41 AM

PR #668, opened 2022-08-01
Subnav (v2): Improve styling of child links

gu-stav:
### What does it do?

Applies all changes from https://github.com/strapi/design-system/pull/635 to the v2 version of the component.

### Why is it needed?

To be in sync between v1 and v2.

### DIFF
--- a/packages/strapi-design-system/src/SubNav/SubNavLink.js
+++ b/packages/strapi-design-system/src/SubNav/SubNavLink.js
@@ -88,6 +88,5 @@ SubNavLink.propTypes = {
   icon: PropTypes.element,
   isSubSectionChild: PropTypes.bool,
   link: PropTypes.element,
-  subSectionChild: PropTypes.bool,
   withBullet: PropTypes.bool,
 };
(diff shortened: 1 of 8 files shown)


packages/strapi-design-system/src/SubNav/SubNavLink.js
```diff
@@ -88,5 +88,6 @@ SubNavLink.propTypes = {
   icon: PropTypes.element,
   isSubSectionChild: PropTypes.bool,
   link: PropTypes.element,
+  subSectionChild: PropTypes.bool,
```
ronronscelestes, line 91: should it be `isSubSectionChild`?

ronronscelestes (changes requested):
Found a small typo. Other than that looks good, thank you!

HichamELBSI (approved):
LGTM

packages/strapi-design-system/src/SubNav/SubNavLink.js
gu-stav: Good catch. I messed up the rebease - should be fixed.

ronronscelestes (approved):
LGTM 🌞💅

PR #675, opened 2022-08-22
[Fix] Link font

ronronscelestes:
## What

Update Links font from `pi` to `omega` variant

## Snapshots

Before:
image

After:
image


gu-stav (approved):

maevalienard:
That's one quick win! Thanks very much 🎉

PR #676, opened 2022-08-23
Combobox filtering behavior

udimberto:
### What does it do?
Updates the filtering rule for the `Combobox` component.

Before, just showing options which starts with the typed term.
Now, displays options containing any part of the typed term.

### Why is it needed?
Will help out to search for User Repositories at the Strapi Cloud Dashboard.

### How to test it?
Unit tests.
E2E existent tests.

### DIFF
--- a/packages/strapi-design-system/src/Combobox/__tests__/utils.spec.js
+++ b/packages/strapi-design-system/src/Combobox/__tests__/utils.spec.js
@@ -1,27 +1,94 @@
 import { filterOptions } from '../utils';
 
-describe('Combobox | utils', () => {
+describe('Combobox | utils', () => {
   describe('filterOptions', () => {
+    const options = [{ props: { children: 'Hamburger' } }, { props: { children: 'Bagel' } }];
+
     it('should filter the options correctly', () => {
-      const options = [{ props: { children: 'Hamburger' } }, { props: { children: 'Bagel' } }];
       const actual = filterOptions(options, 'ham');
-      const expected = [{ props: { children: 'Hamburger' } }];
+      const expected = [options[0]];
+
       expect(actual).toEqual(expected);
     });
 
-    it('should not filter the options if the search is empty or null', () => {
-      const options = [{ props: { children: 'Hamburger' } }, { props: { children: 'Bagel' } }];
-      const expected = [{ props: { children: 'Hamburger' } }, { props: { children: 'Bagel' } }];
+    it('empty array, without any param', () => {
+      const expected = [];
+      const result = filterOptions();
 
-      expect(filterOptions(options, '')).toEqual(expected);
-      expect(filterOptions(options, null)).toEqual(expected);
+      expect(result).toEqual(expected);
     });
 
-    it('should also filter the options if the search is a number', () => {
-      const options = [{ props: { children: 421 } }, { props: { children: 1285 } }];
-      const expected = [{ props: { children: 1285 } }];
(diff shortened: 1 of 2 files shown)


packages/strapi-design-system/src/Combobox/__tests__/utils.spec.js
```diff
@@ -6,9 +6,48 @@ describe('Combobox | utils', () => {
...
+      });
+
+      it('2 last of 5 items', () => {
+        const result = filterOptions(options, 'strapi');
+        const expected = [options[3], options[4]];
+
+        expect(result).toEqual(expected);
+      });
+
+      it('1 item', () => {
```
gu-stav, line 43: What do you think about adding one more test, to prove the edge-case `filterOptions(options, '')` or `filterOptions(options, null)`is working well?

packages/strapi-design-system/src/Combobox/utils.js
```diff
@@ -30,10 +30,13 @@ export const TreeActions = {
 // filter an array of options against an input string
 // returns an array of options that begin with the filter string, case-independent
 export function filterOptions(options = [], filter, exclude = []) {
-  return filter
+  const equalizedTerm = String(filter || '').toLowerCase();
```
gu-stav, line 33: What do you think about using:

```suggestion
  const equalizedTerm = String(filter ?? '').toLowerCase();
```

instead?

```js
String(0 || '') // ''
String(0 ?? '') // '0'
```

packages/strapi-design-system/src/Combobox/__tests__/utils.spec.js
```diff
@@ -6,9 +6,48 @@ describe('Combobox | utils', () => {
...
       const expected = [{ props: { children: 'Hamburger' } }];
+
       expect(actual).toEqual(expected);
     });
 
+    describe('should filter the options correctly, with partial term', () => {
+      const options = [
+        { props: { children: 'hicham/project-super-cms' } },
+        { props: { children: 'pierre/project-awesome-cms' } },
+        { props: { children: 'udi/cool-cms' } },
```
HichamELBSI, line 17: Not sure that we should keep the Strapier names in the codebase 😅. Can you rename these values please 👍🏽

packages/strapi-design-system/src/Combobox/__tests__/utils.spec.js
HichamELBSI: +1

HichamELBSI (changes requested):

packages/strapi-design-system/src/Combobox/__tests__/utils.spec.js
udimberto: Cool.
Those test-cases already exists.
I've re-order them now, adding one more.
Thank you

packages/strapi-design-system/src/Combobox/utils.js
udimberto: Awesome!
Thank you.

packages/strapi-design-system/src/Combobox/__tests__/utils.spec.js
udimberto: Sorry about that.
Fixed.
Thank you.

HichamELBSI (approved):

gu-stav (approved):
Well done! :)

PR #677, opened 2022-08-24
chore: Update many children prop-types from string to node

gu-stav:
### What does it do?

In the tests of the core repo we see a couple of prop-type errors, because some components can only render `string` for `children` props. This PR opens many of these to be able to receive `node` instead.

### Why is it needed?

While `string` might work, it sometimes might not, if `formatMessage()` renders sub-components as replacements. I think opening `children` doesn't do any harm, because using a node instead of pure text isn't harmful in most cases, such as:

```js
<Checkbox>
  <span>Label</span>
</Checkbox>
```

I think it avoids some frustration in the future and I think `string` is too strict in most cases.

### DIFF
--- a/packages/strapi-design-system/src/Alert/Alert.js
+++ b/packages/strapi-design-system/src/Alert/Alert.js
@@ -116,7 +116,7 @@ Alert.propTypes = {
   /**
    * The body of the `Alert` (Will be rendered under the `Alert` title).
    */
-  children: PropTypes.string.isRequired,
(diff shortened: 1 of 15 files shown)


HichamELBSI (changes requested):
Many thanks for this, we are also having this issue with the children prop of the Loader component. Can you make the changes in it as well please?

HichamELBSI (approved):
Thanks

PR #681, opened 2022-08-25
Add command icon

alexandrebodin:
### What does it do?

Add a command (command key on the keyboard) icon

### Why is it needed?

To be usable for the CMS 

### How to test it?

### Related issue(s)/PR(s)


alexandrebodin:
I experimented with some chunking config I don't count on leaving it in the PR but have a look at the analyze and let me know what you think 👍

packages/strapi-design-system/webpack.config.js
```diff
@@ -52,6 +53,14 @@ module.exports = {
     library: 'strapiDs',
     umdNamedDefine: true,
   },
+  optimization: {
+    splitChunks: {
+      chunks: 'all',
+      maxInitialRequests: 30,
+      maxAsyncRequests: 30,
+      minSize: 20,
```
HichamELBSI, line 61: Curious to know why 20bytes to generate a chunk?

HichamELBSI (changes requested):

packages/strapi-design-system/webpack.config.js
```diff
@@ -28,7 +28,7 @@ const entry = fileNames
     return acc;
   }, {});
 
-// Plugin section
+// Plugin <section></section>
```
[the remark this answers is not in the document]
alexandrebodin: Nothing :D my prettier went crazy I didn't see it

packages/strapi-design-system/webpack.config.js
alexandrebodin: If you don't put that webpack won't consider the shared packages as being "big enough" to put them in a common chunk we can play we this to force chunking. 

I'll move it outside of this pr and open a draft one with this

HichamELBSI (approved):

PR #682, opened 2022-08-26
ToggleInput: Remove clear button, if the field is disabled

gu-stav:
### What does it do?

Does not display the "clear" button if the field is disabled.

### Why is it needed?

To disallow clearing a field, if it is disabled.

### How to test it?

Use the disabled story of "ToggleInput"

### Related issue(s)/PR(s)

- Fixes https://github.com/strapi/design-system/issues/670


HichamELBSI (approved):
Awesome!

Issue #671, opened 2022-08-05
NumberInput: does not handle scientific numbers


gu-stav:
## Bug report

When a decimal with a lot of decimals is used in the `NumberInput`, it renders the internal JS `Number` representation instead of a well formatted string.

Example: `00000000001` -> `1e-10`

The latter can not be edited.

### Required System information

- Strapi Design System version: 1.2.1

### Additional context

- Refs https://github.com/strapi/strapi/pull/13993

PR #683, opened 2022-08-26
NumberInput: Keep scientific numbers after blur/ focus as value

gu-stav:
### What does it do?

Fixes `NumberInput`, if scientific values were entered. 

### Why is it needed?

Currently, if you enter a scientific number into the `NumberInput` the following happens:

- enter `0.00000001`
- blur
- focus
- values switches to `1e-8` and can not be edited any longer

This PR makes sure, we display `0.00000001` even after the focus event.

### How to test it?

I've added an automated test.

### Related issue(s)/PR(s)

- Fixes https://github.com/strapi/design-system/issues/671

### DIFF
--- a/packages/strapi-design-system/src/NumberInput/NumberInput.js
+++ b/packages/strapi-design-system/src/NumberInput/NumberInput.js
@@ -144,7 +144,7 @@ export const NumberInput = React.forwardRef(
 
     const handleFocus = () => {
       if (value !== undefined) {
-        setInputValue(String(numberParserRef.current.parse(inputValue) ?? INITIAL_VALUE));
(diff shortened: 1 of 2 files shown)


HichamELBSI (approved):
Nice one!

dz85:
hey, this pr brought a new issue: https://github.com/strapi/strapi/issues/14396

cos' it seems that the class `NumberParser` doesn't have the method `format` as follow:

https://github.com/adobe/react-spectrum/blob/%40internationalized/number%403.1.1/packages/@internationalized/number/src/NumberParser.ts

gu-stav:
@dz85 Thanks - I think this might be a race condition. I'll move the issue into this repo and then we'll have a look hopefully very soon.

Issue #667, opened 2022-07-28
Unable to add linebreaks to TextInput hints


jclaveau:
### Required System information

- Npm version 8.12.1
- NodeJS version: v16.15.1
- Strapi Design System version: @strapi/design-system@1.2.0
- Browser: Firefox 

### Describe the bug

Adding a JSX node object to the "hint" property of a TextInput generates a warning `Warning: Failed prop type: Invalid prop 'hint' of type 'object' supplied to 'Field', expected` so adding a line break to it is not possible without spaming the console

### Steps to reproduce the behavior

1. Create a TextInput
2. Add a hint with linebreak as string `hint="aaa<br/>bbb"`
![image](https://user-images.githubusercontent.com/1556489/181644754-1501c03a-f57c-482e-8912-4ade3a649ead.png)
3. the "<br/ >" appears on the page
4. Replace it by a JSX object `hint={<span>aaa<br/>bbb</span>}` then it works
![image](https://user-images.githubusercontent.com/1556489/181644824-7d34d10f-7e44-46cc-a1ad-7551405c510c.png)
6. But your console is obfuscated by warnings

### Expected behavior

The console should remain clean

### Screenshots
![image](https://user-images.githubusercontent.com/1556489/181644374-a6184050-513b-44f4-9724-ca406216d38c.png)

### Code snippets

```
<TextInput
  name="myInput"
  hint={
    <span>
      - First line
      <br/>
      - Second line
    </span>
  }
  value={123}
/>
```

### Additional context

Adding node propType for hints would be very useful here https://github.com/strapi/design-system/blob/main/packages/strapi-design-system/src/TextInput/TextInput.js#L59
```
PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.node,
    PropTypes.arrayOf(PropTypes.node),
])
```

jclaveau:
Thank you!

PR #684, opened 2022-08-26
Relax hint prop-types from string to node

gu-stav:
### What does it do?

Allows `node` prop-types to be passed in as `hint` for `TextInput` components.

### Why is it needed?

I think it makes sense to open that prop, so that users can add e.g. multiline-text.

### Related issue(s)/PR(s)

- Fixes https://github.com/strapi/design-system/issues/667

### DIFF
--- a/packages/strapi-design-system/src/CarouselInput/Carousel.js
+++ b/packages/strapi-design-system/src/CarouselInput/Carousel.js
@@ -126,7 +126,7 @@ Carousel.propTypes = {
   actions: PropTypes.node,
   children: PropTypes.node.isRequired,
   error: PropTypes.string,
-  hint: PropTypes.string,
+  hint: PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.arrayOf(PropTypes.node)]),
   label: PropTypes.string.isRequired,
   nextLabel: PropTypes.string.isRequired,
   onNext: PropTypes.func.isRequired,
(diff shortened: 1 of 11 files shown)


HichamELBSI:
I'm not sure about these changes. There is an impact on the design guideline documentation of the components. Has this been validated with @maevalienard, I have the feeling that it goes against our guidelines. In the documentation of the TextInput component, we specify that the hint is a description line.

Also, it opens the component a bit too much to customization. Foe example, a user could put a button instead of a text and completely break the guidelines of the design system in a plugin without any warning.

What's your thought on it?

https://design-system-git-main-strapijs.vercel.app/?path=/docs/design-system-components-textinput--base

gu-stav:
@HichamELBSI I haven't discussed it with Maeva, because it is a prop-type change. It is already possible to pass in `node`/ `[node]`, but the prop-types will throw an error. Tbh, for me `string` is only a more strict `node` prop-type and I don't see any disadvantage if we allow passing `<span>Text</span>` over `Text`. I do see the use-case to allow line-breaks and let's say links for an external documentation ... in hints.

Either way I don't have a strong opinion and if you'd like to keep it as it is I can live with that :)

HichamELBSI:
> @HichamELBSI I haven't discussed it with Maeva, because it is a prop-type change. It is already possible to pass in `node`/ `[node]`, but the prop-types will throw an error. Tbh, for me `string` is only a more strict `node` prop-type and I don't see any disadvantage if we allow passing `<span>Text</span>` over `Text`. I do see the use-case to allow line-breaks and let's say links for an external documentation ... in hints.
> 
> Either way I don't have a strong opinion and if you'd like to keep it as it is I can live with that :)

I think we can move forward with this PR. You are right, this is purely a tech decision a strict proptype won't prevent bad usage. And TBH I'm done with these strict proptypes 😅

HichamELBSI (approved):

PR #686, opened 2022-08-26
chore/webpack chunks

alexandrebodin:
- Experimenting with webpack chunks

### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/packages/strapi-design-system/webpack.config.js
+++ b/packages/strapi-design-system/webpack.config.js
@@ -40,6 +40,7 @@ const externalNodeModules = [];
   externalNodeModules.push(new RegExp(`^${module}(/.+)?$`));
 });
 
+/** @type import('webpack').Configuration */
 module.exports = {
   entry,
   mode: process.env.NODE_ENV,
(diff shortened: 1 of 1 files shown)


HichamELBSI:
@joshuaellis I'd like to have your opinion on this PR please

joshuaellis:
I'm not sure chunking would help? Isn't the issue here that when we build the library we wrap up `Flex` as a `production` ready file that includes `Box` as it's dep, but isn't it too early to be doing this? It then causes `Box` to be included multiple times.

This could always be to do with the fact we're releasing a UMD bundle. I think to solve the issue we probably want to be releasing a `ESM`, `CJS` and `UMD` version of the library.

The only example I can think of that I know works (doesn't duplicate it's imports from it's own lib) is https://github.com/pmndrs/drei where we import `OrbitControls` into the `Stage` component, when you build both the ESM & CJS version are using their respective versions of `import` to bring in the `OrbitControls`, so I assume when people build their packages / apps / libraries only one version of the `OrbitControls` is there.

joshuaellis:
We've merged a beta for using `vite` so i think we can close this.

PR #687, opened 2022-08-29
NumberInput: Allow to pass in a locale for improved number formatting

gu-stav:
### What does it do?

This PR adds a new prop `locale` to `NumberInput` fields, which is used to determine the current number format. Currently we determine the locale from the navigator string. This is not always true and I think in strapi core we should set this to the selected locale. By default the locale from the global navigator object is used.

I wanted to open this PR first to start a discussion on the topic. If we agree on this approach I think we'd need to add a consistent handling for `DatePicker`s too, because they also rely on `navigator` for now.

### Why is it needed?

In Germany, for example, the decimal seperator is `,` instead of `.`. It is very confusing not to be able to use the localized seperator. In order for `Intl` to determine the proper locale, we need to set it to the users preference.

### How to test it?

I've added a new story and automated tests.

### Related issue(s)/PR(s)

- Fixes https://github.com/strapi/design-system/issues/650


HichamELBSI (approved):

gu-stav:
@HichamELBSI Please don't merge this yet. To stay consistent we should apply this to `DatePicker`s too (imo), which is material for a next release.

joshuaellis:
I've added tests to #688 so hopefully they can be merged together.

PR #688, opened 2022-08-29
DatePicker: Allow to pass in a locale

gu-stav:
### What does it do?

Similar to https://github.com/strapi/design-system/pull/687: it allows us to pass in a locale (e.g. from users choice in strapi core), which overrides the one from `navigator`.

### Why is it needed?

To locale fields to the selected locale of a users instead of the one from the browser.

### How to test it?

* I've added some snapshot testing of the input value for EN & DE
* Added a new story to illustrate the locale in DE

### DIFF
--- a/packages/strapi-design-system/src/DatePicker/DatePicker.js
+++ b/packages/strapi-design-system/src/DatePicker/DatePicker.js
@@ -1,4 +1,4 @@
-import React, { useRef, useState } from 'react';
+import React, { useRef, useState, useMemo } from 'react';
 import PropTypes from 'prop-types';
 import CalendarIcon from '@strapi/icons/Calendar';
 import Cross from '@strapi/icons/Cross';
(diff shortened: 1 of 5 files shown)


packages/strapi-design-system/src/DatePicker/DatePicker.js
```diff
@@ -26,7 +28,8 @@ export const DatePicker = ({
   const [visible, setVisible] = useState(false);
   const inputRef = useRef(null);
   const datePickerButtonRef = useRef(null);
-  const formattedDate = selectedDate ? formatDate(selectedDate) : '';
+  const locale = useMemo(() => defaultLocale || getDefaultLocale(), [locale]);
```
HichamELBSI, line 31: Do we really need the useMemo here?

joshuaellis: No I agree, i don't think we do.

HichamELBSI (approved):

PR #690, opened 2022-08-30
Pass the required prop to the combobox FieldLabel

HichamELBSI:

### What does it do?

Allow the combobox to display the required state.

### DIFF
--- a/packages/strapi-design-system/src/Combobox/Combobox.js
+++ b/packages/strapi-design-system/src/Combobox/Combobox.js
@@ -20,25 +20,27 @@ import { KeyboardKeys } from '../helpers/keyboardKeys';
 import { VisuallyHidden } from '../VisuallyHidden';
 
 export const Combobox = ({
+  children: nodes,
   clearLabel,
+  creatable,
   createMessage,
   disabled,
-  hint,
   error,
+  hasMoreItems,
+  hint,
   label,
-  value,
-  onChange,
-  placeholder,
-  creatable,
+  labelAction,
   loading,
   loadingMessage,
+  noOptionsMessage,
+  onChange,
+  onClear,
   onCreateOption,
   onInputChange,
   onLoadMore,
-  noOptionsMessage,
-  hasMoreItems,
-  children: nodes,
-  onClear,
+  placeholder,
+  required,
+  value,
   ...props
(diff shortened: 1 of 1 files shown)


packages/strapi-design-system/src/Combobox/Combobox.js
```diff
@@ -250,7 +251,11 @@ export const Combobox = ({
         {value}
       </VisuallyHidden>
       <Stack spacing={label || hint || error ? 1 : 0}>
-        {label && <FieldLabel id={labelId}>{label}</FieldLabel>}
+        {label && (
+          <FieldLabel required={required} id={labelId}>
```
gu-stav, line 255: Good catch! Could you also check if the `labelAction` is passed through as `action={labelAction}`? It seems to be missing too.

udimberto (approved):
### LGTM


### Observation
There's a strange behavior with the Combobox.
It's broking, when does not have any `ComboboxOption` as child.
But I think it's a problem to another time.

### Good job

packages/strapi-design-system/src/Combobox/Combobox.js
```diff
@@ -20,25 +20,27 @@ import { KeyboardKeys } from '../helpers/keyboardKeys';
...
   onCreateOption,
   onInputChange,
   onLoadMore,
-  noOptionsMessage,
-  hasMoreItems,
-  children: nodes,
-  onClear,
+  placeholder,
+  required,
+  value,
```
HichamELBSI, line 43: ℹ️  Just sorting the props list 😅

packages/strapi-design-system/src/Combobox/Combobox.js
HichamELBSI: Nice catch! Pushed 👍🏽

packages/strapi-design-system/src/Combobox/Combobox.js
gu-stav: 💡  We need an eslint rule for that.

gu-stav (approved):
Looks good to me. We should keep in mind that I think we need to create a `ComboboxInput` component to be consistent with the rest of the input fields at some point.

Well done - thanks!

PR #691, opened 2022-09-05
[Fix] TextButton loading state

ronronscelestes:
## What

Added missing loading state for TextButton
Used the same UI as Button

## Snapshot

image

### DIFF
--- a/packages/strapi-design-system/src/TextButton/TextButton.js
+++ b/packages/strapi-design-system/src/TextButton/TextButton.js
@@ -18,6 +18,7 @@ const rotation = keyframes`
 
 const LoadingWrapper = styled.div`
   animation: ${rotation} 2s infinite linear;
+  will-change: transform;
 `;
 
 const TextButtonWrapper = styled(Flex)`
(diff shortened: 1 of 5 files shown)


maevalienard (approved):
LGTM, thank you very much! 🎉

packages/strapi-design-system/src/TextButton/TextButton.js
```diff
@@ -1,11 +1,25 @@
...
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(359deg);
+  }
+`;
+
+const LoadingWrapper = styled.div`
+  animation: ${rotation} 2s infinite linear;
```
gu-stav, line 20: Could you please add a `will-change: transform;` here too, to animate it smoothly? If that is missing from the others could you add it there too? :)

ronronscelestes: interesting I didn't know about this one
from the documentation I see it should be used only when really necessary/last resort on performance issues, do you think we really need it here? 

image

gu-stav: In this case it will do a good job, because it will move the element into it's own compositor and therefore make sure it always renders at 60fps.

ronronscelestes: okay good to know, on it!

gu-stav (approved):
Thanks!

Issue #674, opened 2022-08-18
Popover element can't properly handle long multi-line labels


florianmrz:
### Required System information

- Npm version: 8.12.1
- NodeJS version: 16.13.2
- Strapi Design System version: 1.2.1
- Browser: Chrome

### Describe the bug

The `Popover` component is not able to properly display long multi-line labels.
I think this happens when the content is longer than the available space for the popover itself and the label is multi-line.
This happened for me in the list view settings, when adding additional fields to be displayed as headers.
The GIF doesn't really justify how it looks, as the dropdown becomes basically unusable.

![chrome_SqzaO4SiVG](https://user-images.githubusercontent.com/8062112/185376934-83cc54dd-d5ec-484a-b444-82798acbd086.gif)

### Steps to reproduce the behavior

1. Check the following code sandbox: https://codesandbox.io/s/silly-burnell-3dn0qv?file=/src/App.js
2. Shrink the preview size and open the dropdown until the issue occurs

![chrome_oUjjqQLn2P](https://user-images.githubusercontent.com/8062112/185378043-60f84e81-785d-42fb-b17e-9aff04718078.gif)

### Expected behavior

The popover should properly display multi-line labels that are longer than the available popover space.

### Screenshots

See GIFs above.

### Code snippets

The issue itself is with the `Popover` component, that's used in the `SimpleMenu` component.

PR #697, opened 2022-09-16
[POC] Convert Popover to Radix-UI popover

joshuaellis:
### What does it do?

This is a POC for the upcoming DS sync, it shows the potential of replacing difficult components (such as Popover / Accordion / Dropdowns) with `radix-ui` components as a base.

### Why is it needed?

Radix-UI is a un-styled design system full of UI primitives enabling user to create accessible interfaces without the technical knowledge (full accessibility is hard and not every dev has the same level of knowledge on the subject). It would mean we could focus on design implementation and adding variations based on the needs of the product.

I haven't resolved any of the tests, nor added the actual `PropTypes` that the radix components expose, hence why this is *definitely* a draft PR.

### Related issue(s)/PR(s)

This came about as a response from #674 where we were using a `ResizeObserver` as the handler for it's floating placement in conjunction with a custom `position` function.

### Other thoughts

* I find it odd that the popover is trapped in a fixed height space, what happens if we want to use a non-list ui in the component? There might be some issues? It feels like the list version warrants being it's own variation of a base `Popover` styled in the Strapi themes?
* This might be part of a larger effort to expand & simplify our design system in a largely framework agnostic way (v2)

### DIFF
--- a/packages/strapi-design-system/package.json
+++ b/packages/strapi-design-system/package.json
@@ -50,6 +50,7 @@
   },
   "dependencies": {
     "@internationalized/number": "^3.1.1",
+    "@radix-ui/react-popover": "^1.0.0",
     "compute-scroll-into-view": "^1.0.17",
     "prop-types": "^15.7.2"
   },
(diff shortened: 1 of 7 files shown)


joshuaellis:
Closing now the engineering session is over.

PR #698, opened 2022-09-16
NumberInput: Fix value assignment during focus

gu-stav:
### What does it do?

It fixes the `NumberInput` field:

- `maximumSignificantDigits` -> `maximumFractionDigits`: The maximum number of fraction digits to use: this is what I wanted in the first place, to fix scientific numbers.
- `numberParserRef` does not have a `.format()` method. I broke that in https://github.com/strapi/design-system/pull/683 and have no idea why the tests worked.

### Why is it needed?

Fixes https://github.com/strapi/design-system/issues/693

### DIFF
--- a/packages/strapi-design-system/src/NumberInput/NumberInput.js
+++ b/packages/strapi-design-system/src/NumberInput/NumberInput.js
@@ -50,7 +50,7 @@ export const NumberInput = React.forwardRef(
     const [inputValue, setInputValue] = useState(value === undefined || value === null ? INITIAL_VALUE : String(value));
     const generatedId = useId('numberinput', id);
     const numberParserRef = useRef(new NumberParser(getDefaultLocale()));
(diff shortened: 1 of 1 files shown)


ronronscelestes (approved):
LGTM!

ronronscelestes (approved):
I caught some behaviors but all of them were there before and might be intended
We will still merge this fix for it to be quickly released

**List of questions for later discussions/fix:**

- when increasing the value with the arrow button, it sometimes changes the values of decimals
- the decimals will go back its previous state if you then decrease it (still with arrow button) 
image
image

also when adding decimals, the decimals sometimes weirdly change (from 3 to 6 here):
image

for large values, when typing everything seems to work fine - but after focusing somewhere else, after 16 digits all other digits will update to zeros (I don't remember if it was intended):
image
image

PR #699, opened 2022-09-20
[Popover]: fix position function by using @floating-ui

joshuaellis:
### What does it do?

* Removes `useResizeObserver` from `Popover` and replaces the functionality with `@floating-ui` library
* Updates `useResizeObserver` to have it's callback based on `useLayoutEffect` to avoid incorrectly styled flashes of content
* Adds a new test simulating a user using the popover tool

### Why is it needed?

#674 was caused because the positioning function was being called based on a resize-observer, in small spaces this fails (as seen in 674).

### Related issue(s)/PR(s)

#674


packages/strapi-design-system/src/helpers/useResizeObserver.js
```diff
@@ -1,7 +1,7 @@
-import { useEffect } from 'react';
+import { useLayoutEffect } from 'react';
 
 export const useResizeObserver = (sources, onResize) => {
-  useEffect(() => {
+  useLayoutEffect(() => {
```
joshuaellis, line 4: This runs before paint of the DOM, it therefore stops flashes of incorrectly styled elements (which was the previous case in Popover)

ronronscelestes:
Works quite well on my machine, no flashes anymore, and nice dynamic positioning! Thanks 💅
2 questions:
- I'm really curious if you know how repositioning (top/bottom) on overflow works with floating UI, I didn't catch any position parameters about it - is it doing it by default?
- This PR is open + there is the radix POC on draft: if we go for radix I guess we won't need floating anymore? 👀 what I mean is do we keep this PR + POC on draft state and talk about it in the next DS sync or did you want to possibly merge this one before talking about radix?

joshuaellis:
hey @ronronscelestes, great questions, see answers below:

> I'm really curious if you know how repositioning (top/bottom) on overflow works with floating UI, I didn't catch any position parameters about it - is it doing it by default?

This is handled by [`floating-ui`](https://floating-ui.com/) (thats the easy answer). In more detail, you enable the `flip` plugin so the hook knows it can position it on the opposite side of the one it's originally attached too (in our case, we default bottom). We then use shift to make sure it's always in the viewport too! The placement only changes if you want to `center` the popover, otherwise it's the default of `bottom-left` which equates the to sticking to the `left` side of our `button`. `offset` is only used to give some spacing between the element and the popover! 

The actual math is handled by the `computePosition` function i believe, you'd have to read the source to understand that more.

> This PR is open + there is the radix POC on draft: if we go for radix I guess we won't need floating anymore? 👀 what I mean is do we keep this PR + POC on draft state and talk about it in the next DS sync or did you want to possibly merge this one before talking about radix?

I personally see the radix PR as a POC to demonstrate how simple it is to use & how we can simplify our APIs for basic use cases to be discussed in the DS Sync, if everyone is onboard then it would be a v2 effort where we would _hopefully_ move the entire design system over to Radix (and any other milestones we want to include on a v2 roadmap). So I would want to merge this to resolve the ticket for sure & if we did move to radix (or an equivalent) then we would be able to remove `floating-ui` 👌🏼

ronronscelestes:
> This is handled by [`floating-ui`](https://floating-ui.com/) (thats the easy answer). In more detail, you enable the `flip` plugin so the hook knows it can position it on the opposite side of the one it's originally attached too (in our case, we default bottom). We then use shift to make sure it's always in the viewport too! The placement only changes if you want to `center` the popover, otherwise it's the default of `bottom-left` which equates the to sticking to the `left` side of our `button`. `offset` is only used to give some spacing between the element and the popover!
> 
> The actual math is handled by the `computePosition` function i believe, you'd have to read the source to understand that more.

Thanks for the details, I'll have a look at the documentation 👀

> I personally see the radix PR as a POC to demonstrate how simple it is to use & how we can simplify our APIs for basic use cases to be discussed in the DS Sync, if everyone is onboard then it would be a v2 effort where we would _hopefully_ move the entire design system over to Radix (and any other milestones we want to include on a v2 roadmap). So I would want to merge this to resolve the ticket for sure & if we did move to radix (or an equivalent) then we would be able to remove `floating-ui` 👌🏼

Makes sense! In my opinion it works very well and good side is that it's very lightweight, if @gu-stav and @HichamELBSI are good with it I'm in!

HichamELBSI (approved):
Seems really good to me, many thanks @joshuaellis 🔥

Issue #642, opened 2022-07-08
Arrow keys do not work inside TextInput component


benderillo:
### Required System information

- NodeJS version: 16.14.0
- Strapi Design System version: 4.2.2
- Browser: Chrome

### Describe the bug

Pressing Arrow Keys inside a `TextInput` element does not move cursor inside the text field. Instead it moves focus to other elements on the page.

### Steps to reproduce the behavior

Here is a piece of React JSX that exhibit this behaviour:

```js
<Table>
<Tbody>
  {entries.map((entry, index) => (
    <Tr key={index}>
      <Td>
        <TextInput
          name={`${index}.title`}
          aria-label="item title"
          onChange={(e) => updateEntry(index, { title: e.target.value })}
          value={entries[index].title}
        />
      </Td>
</Tbody>
</Table>
```
This will render a table with rows having a TextInput component each. 
Try using arrow keys in any of the inputs, and it won't move a cursor within the text but rather move the focus to other elements on the page.

### Expected behavior

Since TextInput is an element for working with text, it should behave like any text editor. 
Arrows should move the cursor within the text and all the shortcuts involving arrow keys should work too, e.g. CMD+SHIFT+ARROW should select entire line, and so on.

### Code snippets
![Screen Shot 2022-07-08 at 9 53 39 AM](https://user-images.githubusercontent.com/1569988/177896076-ace58afc-37e2-4a5d-8e8b-b220e8518fe8.png)

### Additional context

The same behaviour can be seen with TextInput components that are placed inside repeatable component in Content-Manager view.
This is a clue that the issue is really with the `<Table>` component...
If I wrap the `TextInput` in a div that intercepts the `keyDown` and `keyUp` events and stop propagating them up to parent, I can use the `TextInput` as expected.

Issue #614, opened 2022-05-27
KeyboardNavigable component always calls `preventDefault()` on arrow key events


jcberquist:
### Required System information

- Npm version 8.7.0
- NodeJS version: 16.14.2
- Strapi Design System version: 1.1.0
- Browser: Brave 1.39.111

### Describe the bug

`KeyboardNavigable` component always calls `preventDefault()` on arrow key events. I think this is unintended since `preventDefault()` gets called again if the `isValidFocusedElement()` check passes.

### Steps to reproduce the behavior

Please see the discussion here https://github.com/strapi/strapi/issues/13342 for an instance of the issue. If a form input field is contained in a `KeyboardNavigable` component, then the arrow keys will not work to move the cursor in that field. As the original issue there has it, using the latest Strapi version:

1. Content-Type builder: Create a singular type and add a repeatable component with a text input
2. Content Manager: Add a component, type anything in the input field and try to move the cursor with the arrow keys.

### Expected behavior

`KeyboardNavigable` component calls `preventDefault()` only when the `isValidFocusedElement()` check passes.

yanniskadiri:
Closing this one and relating it to #642 as we need to fix arrow key handling in both cases.

PR #702, opened 2022-09-23
[KeyboardNavigable & RawTable]: fix keyboard navigation on child elements

joshuaellis:
### What does it do?

This PR does two things –

* removes the immediate call of `preventDefault()` in `KeyboardNavigable` which hijacks all keyboard events from inputs wrapped inside e.g. `Accordion` (issues https://github.com/strapi/design-system/issues/614 & https://github.com/strapi/strapi/issues/12432)
* Reworks `RawTable` because the aria-pattern implementation was missing handle cases for when cells had fields within that have their own keyboard navigation / more than one focussable field (issue https://github.com/strapi/design-system/issues/642)
* Also sets up so if you click on a focusable child of a `RawTable Cell` then the focussing starts at that point

### Why is it needed?

Users were unable to do a couple of things:

* Were unable to use keyboard navigation in fields within Accordions (e.g. to move the caret)
* Were unable to use keyboard navigation in RawTable Cells (e.g. to move the caret)

The aria pattern was incomplete because when multiple focusable fields are inside a cell the cell should be "activated" before the content is focusable (see image below)

CleanShot 2022-09-23 at 10 56 37@2x

### How to test it?

I've added extensive tests for `RawTable` to validate the API changes for the ARIA design pattern

### Related issue(s)/PR(s)

* resolves https://github.com/strapi/design-system/issues/642
* resolves https://github.com/strapi/strapi/issues/12432

### DIFF
--- a/packages/strapi-design-system/src/RawTable/RawCell.js
+++ b/packages/strapi-design-system/src/RawTable/RawCell.js
@@ -27,6 +27,30 @@ export const RawTd = ({ coords, as, ...props }) => {
       (focusableNodes.length === 1 && getFocusableNodesWithKeyboardNav(focusableNodes).length === 0)
     ) {
       return;
+      /**
+       * This allows cells that **only** have buttons in them to still be
+       * navigable with the keyboard arrow keys (left / right) as if they were grid cells.
+       *
+       * If there are nextNodes (next child node) then we stop the table's keyboard navigation
+       * handlers from happening.
+       */
+    } else if (focusableNodes.length > 1 && !Boolean(focusableNodes.find((node) => node.tagName !== 'BUTTON'))) {
+      e.preventDefault();
+      const focussedButtonIndex = focusableNodes.findIndex((node) => node === document.activeElement);
+      if (e.key === KeyboardKeys.RIGHT) {
+        const nextNode = focusableNodes[focussedButtonIndex + 1];
+        if (nextNode) {
+          e.stopPropagation();
+          nextNode.focus();
+        }
+      } else if (e.key === KeyboardKeys.LEFT) {
+        const nextNode = focusableNodes[focussedButtonIndex - 1];
+        if (nextNode) {
+          e.stopPropagation();
+          nextNode.focus();
+        }
+      }
+      return;
     }
 
     if (e.key === KeyboardKeys.ENTER && !isActive) {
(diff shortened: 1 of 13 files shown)


packages/strapi-design-system/src/RawTable/RawCell.js
```diff
@@ -1,51 +1,151 @@
...
+      (focusableNodes.length === 1 && getFocusableNodesWithKeyboardNav(focusableNodes).length !== 0) ||
+      focusableNodes.length > 1
+    ) {
+      tdRef.current.setAttribute('tabIndex', !isActive && isFocused ? 0 : -1);
+
+      focusableNodes.forEach((node, index) => {
+        node.setAttribute('tabIndex', isActive ? 0 : -1);
+        /**
+         * When a cell is active we want to focus the
+         * first focusable element simulating a focus trap
```
ronronscelestes, line 72: to make sure I understand, was your goal to have a focus trap behaviour when entering into a cell with focusable elements?
when entering this cell:
image

after tabbing through all focusable elements:
image

I end up in the Table footer:
image

I don't have a strong opinion on it, I'm able to recover my previous position with `tab reverse` then `Enter`

ronronscelestes (commented):
Wow thank you for this huge fix 🔥
Tested it, works well on my side, I added one question

packages/strapi-design-system/src/RawTable/RawCell.js
HichamELBSI: That's the normal behaviour because pressing `tab` will move the focus to the next "focusable" element but I agree that it is a weird behavior in this case. Do you think we could fix it? @joshuaellis

joshuaellis: That then breaks the aria-pattern, which is sort of the cause of this issue (note point 1)

CleanShot 2022-09-26 at 14 51 02@2x

Where we originally would go straight to the `input` but that's not expected behaviour and the keyboard navigation prevents default actions so people can't move the caret.

IMHO whilst we think it's a bit odd for behaviour I do strongly believe we should adhere to the guidelines. I think changing for button is tolerable, but we should draw the line there. 

We don't have control over what people put in their tables anyway so if someone does challenge the implementation then at the very least we have the standards documentation to point too.

joshuaellis: Link to documentation – https://www.w3.org/WAI/ARIA/apg/patterns/grid/

HichamELBSI: Good and strong points. I'm ok to restrict it to buttons only, thanks @joshuaellis 🔥

HichamELBSI (approved):
Awesome work @joshuaellis! Many thanks 😄

joshuaellis:
@HichamELBSI I've sorted the merge conflicts & added in the behaviour for exclusively multiple buttons in a table cell as well as adding a new test to validate this behaviour. When the checks have passed this should be good to merge 👍🏼 

Thanks for the feedback @HichamELBSI & @ronronscelestes!

scottydev:
Just updated and can confirm this fix works beautifully. Thanks for taking the time to do this @joshuaellis !

PR #703, opened 2022-09-23
Refactoring the DateTimePicker and the DatePicker and TimePicker state logic

simotae14:
### What does it do?

We want to move the Code of the DateTimePicker from the helper-plugin to the design system because it is better to divide the Design stuffs from the helper plugin.

### Decision
I think there is no reason to maintain the DateTimePicker inside the helper-plugin because it is a Design Component that can be used in other places. In addiction, in future we need to refactor it and the other field components to prevent a little bit the props drilling and transform them into controlled components.

### Consequences
I think we can have a more clear and correct division between components into the different repos. We need also to take care about the possible errors with the old versions of the Design System and strapi

Refactor the fields in order to be able to display an error state (red border) without displaying the error message.

Move the DateTimePicker in the Design System and have a beta components system that would display a warning to the user that this component is still under definition/development and that it is not recommended to use it because the code and the API will change soon.

### How to test it?
We need to test if we make some regressions in the code, in every section where we can use the DateTimePicker, how it works with content from another locale

### Why is it needed?
We need to move the logic from the helper-plugin to the Design System project


packages/strapi-design-system/src/Field/FieldInput.js
```diff
@@ -47,7 +47,9 @@ const Input = styled.input`
 `;
 
 export const InputWrapper = styled(Flex)`
-  border: 1px solid ${({ theme, hasError }) => (hasError ? theme.colors.danger600 : theme.colors.neutral200)};
+  border: 1px solid
+    ${({ theme, hasError, showErrorBorder }) =>
```
ronronscelestes, line 51: maybe I missed a discussion about it - I was curious about why adding this new `showErrorBorder` prop?

ronronscelestes (commented):
Neat thank you - seems to work well on my side
Added a comment about a new prop for error state

packages/strapi-design-system/src/Field/FieldInput.js
simotae14: Hello, because Hicham suggests me to handle differently how we show errors in the DateTimePicker so I thought it was good to just use a boolean to show only the red border and not the error messages for the DatePicker and TimePicker

packages/strapi-design-system/global-setup.js
```diff
@@ -0,0 +1,4 @@
+module.exports = async () => {
+  process.env.TZ = 'UTC';
+  process.env.LANG = 'en_US.UTF-8';
```
gu-stav, line 3: Thank you 🌻

packages/strapi-design-system/src/DatePicker/DatePicker.js
```diff
@@ -64,7 +64,7 @@ export const DatePicker = ({
           <DatePickerButton
             ref={datePickerButtonRef}
             onClick={toggleVisibility}
-            aria-label={selectedDate ? selectedDateLabel(formatDate(selectedDate, locale)) : label}
+            aria-label={selectedDate ? selectedDateLabel(formatDate(selectedDate)) : label ? label : ariaLabel}
```
gu-stav, line 67: Why did you remove `locale` here?

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,189 @@
...
+
+  return (
+    <Field name={name} role="group" aria-labelledby="datetime-label" hint={hint} error={error}>
+      <Stack spacing={1}>
+        {label && (
+          <Flex>
+            <FieldLabel required={required} id="datetime-label">
+              {label}
+            </FieldLabel>
+            {labelAction && <LabelAction paddingLeft={1}>{labelAction}</LabelAction>}
```
gu-stav, line 121:
```suggestion
            <FieldLabel required={required} action={labelAction} id="datetime-label">
              {label}
            </FieldLabel>
```

packages/strapi-design-system/src/Field/FieldInput.js
gu-stav: Why would we not want to show the error message for these fields?

simotae14: because we want to show just the Error message in the DateTimePicker

packages/strapi-design-system/global-setup.js
simotae14: a pleasure

packages/strapi-design-system/src/DatePicker/DatePicker.js
simotae14: my bad now it is back again

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
simotae14: done

packages/strapi-design-system/src/Field/FieldInput.js
joshuaellis: I think the confusion here is you're passing a prop to the component to show an error but normally its handled by the context. It creates two ways of _almost_ doing the same thing.

Maybe the `error` from the context if a `string` should show a message otherwise if it's a `boolean` should just change the styling. 

This would create a more uniform API to handle field errors as opposed to splintering. What do others think?

ronronscelestes: just to be really specific:
- if `error === string`: error message + error styling
- if `error === boolean`: error styling

is that it @joshuaellis ?

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,180 @@
...
+      dateToSet.setHours('00');
+      dateToSet.setMinutes('00');
+    }
+
+    if (onChange) {
+      onChange(dateToSet);
+    }
+  };
+
+  useEffect(() => {
```
joshuaellis, line 92: I find the API to be weird, we're managing the state in the component, but it looks like we're actually supposed to be managing the component outside (via props). If the component is supposed to be designed in a way that you control it from outside then i think using the `state` isn't necessary and it can be derived from the props.

Wdyt?

joshuaellis (changes requested):

packages/strapi-design-system/src/Field/FieldInput.js
joshuaellis: Yes, exactly @ronronscelestes, that way there is one way to do things 😄

ronronscelestes: I like it 💃

HichamELBSI: I like the idea @joshuaellis

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,180 @@
...
+            <FieldLabel required={required} action={labelAction} id="datetime-label">
+              {label}
+            </FieldLabel>
+          </Flex>
+        )}
+        <Stack horizontal spacing={2}>
+          <DatePicker
+            data-testid="datetimepicker-date"
+            name={name}
+            ariaLabel={label || ariaLabel}
```
HichamELBSI, line 119: As we already have the group field label for this formgroup, we shouldn't be able to pass other labels to the inputs. Can you rework it please?

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,180 @@
...
+      if (onChange) {
+        onChange(dateToSet);
+      }
+    } else {
+      const dateToSet = new Date(e);
+      setTimeValue(`${dateToSet.getHours()}:${dateToSet.getMinutes()}:${dateToSet.getSeconds()}`);
+
+      if (onChange) {
+        onChange(dateToSet);
+      }
```
HichamELBSI, line 50: all these conditions has to be tested please.

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,180 @@
...
+    dateToSet.setHours(e.split(':')[0]);
+    dateToSet.setMinutes(e.split(':')[1]);
+
+    if (!dateValue) {
+      setDateValue(new Date());
+    }
+
+    if (onChange) {
+      onChange(dateToSet);
+    }
```
HichamELBSI, line 67: Same here about the tests

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,180 @@
...
+
+    if (dateValue) {
+      dateToSet = new Date(dateValue);
+      dateToSet.setHours('00');
+      dateToSet.setMinutes('00');
+    }
+
+    if (onChange) {
+      onChange(dateToSet);
+    }
```
HichamELBSI, line 89: Same here about the tests

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,180 @@
...
+      const parsedData = parseDate(value);
+      setDateValue(parsedData);
+      setTimeValue(
+        parsedData ? `${parsedData.getHours()}:${parsedData.getMinutes()}:${parsedData.getSeconds()}` : null,
+      );
+    } else {
+      setDateValue(undefined);
+      setTimeValue(undefined);
+    }
+  }, [value]);
```
HichamELBSI, line 103: This use effect + conditions has to be tested, please

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
joshuaellis: But wouldn't that be passed through via the `value` prop?

packages/strapi-design-system/src/DatePicker/DatePicker.js
```diff
@@ -64,7 +64,7 @@ export const DatePicker = ({
           <DatePickerButton
             ref={datePickerButtonRef}
             onClick={toggleVisibility}
-            aria-label={selectedDate ? selectedDateLabel(formatDate(selectedDate, locale)) : label}
+            aria-label={selectedDate ? selectedDateLabel(formatDate(selectedDate, locale)) : label ? label : ariaLabel}
```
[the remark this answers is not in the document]
simotae14: done, let me know if it is better, thanks

packages/strapi-design-system/src/DatePicker/DatePicker.js
```diff
@@ -103,6 +103,7 @@ export const DatePicker = ({
 DatePicker.defaultProps = {
   ariaLabel: undefined,
   clearLabel: undefined,
+  showErrorBorder: false,
```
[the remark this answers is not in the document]
simotae14: yes, you right, I have removed the prop type definition for it

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,180 @@
...
+  hint: PropTypes.string,
+  label: PropTypes.string,
+  labelAction: PropTypes.element,
+  name: PropTypes.string.isRequired,
+  onChange: PropTypes.func.isRequired,
+  onClear: PropTypes.func,
+  required: PropTypes.bool,
+  size: PropTypes.oneOf(['S', 'M']),
+  step: PropTypes.number,
+  value: PropTypes.instanceOf(Date),
```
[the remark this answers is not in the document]
simotae14: ok but I can't see an example in the other Components

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
```diff
@@ -0,0 +1,51 @@
...
+    });
+
+    test('initial data A11y', async ({ page }) => {
+      // This is the URL of the Storybook Iframe
+      await page.goto('/iframe.html?id=design-system-components-datetimepicker--initial-data&viewMode=story');
+      await injectAxe(page);
+      await checkA11y(page);
+    });
+  });
+});
```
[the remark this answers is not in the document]
simotae14: ok sorry, I don't have a lot of experience with axe

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
gu-stav: @joshuaellis I agree this is confusing, but right now (due to the lack of controlled components) we have this in several places, such as https://github.com/strapi/design-system/pull/644

I'd love to get rid of it in the next major iteration, but it requires quite a bit of work to make all components consistent ... :(

packages/strapi-design-system/src/Field/FieldInput.js
simotae14: I tried to implement @joshuaellis idea, what do you think? https://github.com/strapi/design-system/pull/703/commits/b3361c7a2f09e34a13c38db103818fd48e9b20ff

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
joshuaellis: Lets leave as is then

packages/strapi-design-system/src/Field/FieldInput.js
joshuaellis: LGTM

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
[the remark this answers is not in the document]
HichamELBSI: https://github.com/strapi/design-system/blob/main/packages/strapi-design-system/src/Box/BoxProps.js

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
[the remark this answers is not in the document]
HichamELBSI: No worries, Axe is just used to do a11y check on a page. Here you are just running axe on every test but it is always the same page.

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
```diff
@@ -0,0 +1,51 @@
...
+      await injectAxe(page);
+      await checkA11y(page);
+    });
+
+    test('initial data A11y', async ({ page }) => {
+      // This is the URL of the Storybook Iframe
+      await page.goto('/iframe.html?id=design-system-components-datetimepicker--initial-data&viewMode=story');
+      await injectAxe(page);
+      await checkA11y(page);
+    });
```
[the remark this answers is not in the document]
simotae14: ok

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
simotae14: thanks, now it is tested

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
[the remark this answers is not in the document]
simotae14: Could you please take a look now and give me your comments, thanks

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,180 @@
...
+            size={size}
+            onClear={onClear && handleDateClear}
+            clearLabel={clearLabel}
+            disabled={disabled}
+            {...props}
+          />
+          <TimePicker
+            data-testid="datetimepicker-time"
+            size={size}
+            aria-label={label || ariaLabel}
```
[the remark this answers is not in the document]
simotae14: yes but they are required inside the TextInput component, do we want to change the Props of the TextInput to not be required?

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,181 @@
...
+        <FieldHint />
+        <FieldError id="datetimepicker" />
+      </Stack>
+    </Field>
+  );
+};
+
+DateTimePicker.defaultProps = dateTimePickerDefaultProps;
+
+DateTimePicker.propTypes = dateTimePickerPropTypes;
```
joshuaellis, line 149: How come this is defined again? 😄

simotae14: 😅 I forgot to cancel the previous definition, thanks @joshuaellis

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,148 @@
...
+      setTimeValue(undefined);
+    }
+  }, [value]);
+
+  return (
+    <Field name={name} role="group" aria-labelledby="datetime-label" hint={hint} error={error}>
+      <Stack spacing={1}>
+        {label && (
+          <Flex>
+            <FieldLabel required={required} action={labelAction} id="datetime-label">
```
gu-stav, line 106: I think we can not rely on this id here, because multiple `DateTimePicker`s might be used on the same page. There is a custom `useId()` hook somewhere, which you could utilize maybe?

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,148 @@
...
+      setDateValue(undefined);
+      setTimeValue(undefined);
+    }
+  }, [value]);
+
+  return (
+    <Field name={name} role="group" aria-labelledby="datetime-label" hint={hint} error={error}>
+      <Stack spacing={1}>
+        {label && (
+          <Flex>
```
gu-stav, line 105: Is this `Flex` wrapper still needed, now that `labelAction` is part of `FieldLabel`?

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,148 @@
...
+          </Flex>
+        )}
+        <Stack horizontal spacing={2}>
+          <DatePicker
+            data-testid="datetimepicker-date"
+            name={name}
+            ariaLabel={label || ariaLabel}
+            error={typeof error === 'string'}
+            selectedDate={dateValue}
+            selectedDateLabel={(formattedDate) => `Date picker, current is ${formattedDate}`}
```
gu-stav, line 118: This needs to be translateable and therefore we will need to pass it in as a prop.

packages/strapi-design-system/yarn.lock
```diff
@@ -12653,4 +12653,4 @@ yocto-queue@^0.1.0:
 zwitch@^1.0.0:
   version "1.0.5"
   resolved "https://registry.yarnpkg.com/zwitch/-/zwitch-1.0.5.tgz#d11d7381ffed16b742f6af7b3f223d5cd9fe9920"
-  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
+  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
```
gu-stav, line 12656: Why did that hash change?

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.spec.js
```diff
@@ -0,0 +1,106 @@
...
+      <ThemeProvider theme={lightTheme}>
+        <DateTimePicker
+          step={15}
+          onChange={() => {}}
+          name="datetimepicker"
+          label="Date time picker"
+          hint="This is a super description"
+        />
+      </ThemeProvider>,
+    );
```
gu-stav, line 100: Since the setup is repeated repeatedly, what do you think about using a setup for that?

```js
function setup(props) => render(
        <DateTimePicker
          {...props}
        />
      </ThemeProvider>,
)
```

and then re-use it

```js
test('', async () {
  const { rerender } = setup({
   step: 15,
   /// ...
  })
})
```

It would help to reduce the boilerplate per test ...

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.js
```diff
@@ -71,7 +71,7 @@ export const DatePickerCalendar = ({
           </Box>
           <RawTable colCount={7} rowCount={weeks.length + 1} initialCol={activeCol} initialRow={activeRow} role="grid">
             <RawThead>
-              <RawTr>
+              <RawTr aria-rowindex={1}>
```
HichamELBSI, line 74: Not sure to understand why you've added this static rowindex?

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Clear Label
+   */
+  clearLabel: PropTypes.string,
+  /**
+   * Disabled attribute
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Error message or boolean
```
HichamELBSI, line 34: Please describe why it can be a boolean and how the message will be rendered.

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+  labelAction: undefined,
+  onClear: undefined,
+  required: false,
+  size: 'M',
+  step: 1,
+  value: undefined,
+};
+export const dateTimePickerPropTypes = {
+  /**
+   * Aria Label
```
HichamELBSI, line 22: Aria label

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+   */
+  ariaLabel: PropTypes.string,
+  /**
+   * Clear Label
```
HichamELBSI, line 26: Clear label

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Disabled attribute
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Error message or boolean
+   */
+  error: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
+  /**
+   * Hint text
```
HichamELBSI, line 38: Where this hint text will be shown

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+   */
+  hint: PropTypes.string,
+  /**
+   * Label
```
HichamELBSI, line 42: What label ?

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+   */
+  label: PropTypes.string,
+  /**
+   * Label Action
```
HichamELBSI, line 46: Please describe what's this label is for.

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+  size: PropTypes.oneOf(['S', 'M']),
+  /**
+   * Step value
+   */
+  step: PropTypes.number,
+  /**
+   * Value wich is a Date value
+   */
+  value: PropTypes.instanceOf(Date),
+};
```
HichamELBSI, line 77: Can you write sentences to document the properties please because you just put the names of the properties each time and there is no use for these descriptions.

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Label Action
+   */
+  labelAction: PropTypes.element,
+  /**
+   * Name, required
+   */
+  name: PropTypes.string.isRequired,
+  /**
+   * onChange function
```
HichamELBSI, line 54: How to use it, what parameters, etc

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Required field boolean
+   */
+  required: PropTypes.bool,
+  /**
+   * Size of the Field (possible values S anf M)
+   */
+  size: PropTypes.oneOf(['S', 'M']),
+  /**
+   * Step value
```
HichamELBSI, line 70: What is this step for ? Can you add some precision please?

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
HichamELBSI: These components will be used by external developers. There is a serious lack of documentation for the Design System components. We need to make an effort to write clear documentation.

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
```diff
@@ -0,0 +1,137 @@
...
+    test.beforeEach(async ({ page }) => {
+      // This is the URL of the Storybook Iframe
+      await page.goto('/iframe.html?id=design-system-components-datetimepicker--base&viewMode=story');
+      await injectAxe(page);
+    });
+    test('change the date and see if also the time will be changed when both are empty', async ({
+      page,
+      browserName,
+    }) => {
+      test.skip(browserName === 'webkit', 'Still working on it');
```
HichamELBSI, line 16: Why do you skip `webkit` browsers 🤔 ?

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
```diff
@@ -0,0 +1,137 @@
+const { injectAxe } = require('axe-playwright');
+
+const { test, expect } = require('@playwright/test');
+
+test.describe.parallel('DateTimePicker', () => {
+  test.describe('without initial data', () => {
+    test.beforeEach(async ({ page }) => {
+      // This is the URL of the Storybook Iframe
+      await page.goto('/iframe.html?id=design-system-components-datetimepicker--base&viewMode=story');
+      await injectAxe(page);
```
HichamELBSI, line 10: Why are you injecting axe before each test ?

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
```diff
@@ -0,0 +1,137 @@
...
+
+      await page.keyboard.press('Tab');
+      await page.keyboard.press('Enter');
+      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('Tab');
+      await page.keyboard.press('Tab'); // move to the Su column
+      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('ArrowDown'); // move to the third Sunday of the month
+      await page.keyboard.press('Enter'); // to select the date
```
HichamELBSI, line 27: I'm not sure to understand why you need to do all these keyboard press? Can't you just do that with a click ?

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
```diff
@@ -0,0 +1,137 @@
...
+      await page.focus('[aria-labelledby="datetime-label"] > div > div:nth-child(2) > div');
+
+      await page.keyboard.press('Tab');
+      await page.keyboard.press('Tab');
+      await page.keyboard.press('Tab');
+      await page.keyboard.press('Enter');
+      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('ArrowDown');
+      await page.keyboard.press('Enter'); // to select the time
```
HichamELBSI, line 46: Same question about all the keyboard press and same for the next tests.

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
```diff
@@ -0,0 +1,137 @@
...
+      await page.keyboard.press('Enter'); // to select the date
+      // check if the time is selected with 00:00
+      const timePickerValue = page.locator('span:has-text("00:00")');
+      expect(await timePickerValue.innerText()).toBe('00:00');
+    });
+    test('change the time and see if also the date will be changed when both are empty', async ({
+      page,
+      browserName,
+    }) => {
+      test.skip(browserName === 'webkit', 'Still working on it');
```
HichamELBSI, line 36: Same question about skip and same for the next tests

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
```diff
@@ -0,0 +1,137 @@
...
+      const today = new Date().toLocaleDateString('en-US');
+      const datePickerValue = page.locator('[name="datetimepicker"]');
+      expect(await datePickerValue.inputValue()).toBe(today);
+    });
+  });
+  test.describe('with initial data', () => {
+    test.beforeEach(async ({ page }) => {
+      // This is the URL of the Storybook Iframe
+      await page.goto('/iframe.html?id=design-system-components-datetimepicker--initial-data&viewMode=story');
+      await injectAxe(page);
```
HichamELBSI, line 57: Same question about injectAxe before each

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
HichamELBSI: And you are not doing any `checkA11y`, injecting axe is useless without `checkA11y`

packages/strapi-design-system/src/Select/Select.js
```diff
@@ -53,7 +53,8 @@ export const Select = ({
 
   const labelId = `${generatedId}-label`;
   const contentId = `${generatedId}-content`;
-  const ariaDescribedBy = error ? `${generatedId}-error` : hint ? `${generatedId}-hint` : undefined;
+  const ariaDescribedBy =
+    error && typeof error === 'string' ? `${generatedId}-error` : hint ? `${generatedId}-hint` : undefined;
```
HichamELBSI, line 57: These nested ternaries aren't easy to read. IMO you can do something like that

```suggestion
  const hasStringError = typeof error === 'string';
  const errorMessageId = hasStringError ? `${generatedId}-error` : undefined;
  const hintMessageId = hint ? `${generatedId}-hint` : undefined;
  
  const ariaDescribedBy = errorMessageId || hintMessageId;
```

packages/strapi-design-system/src/Select/Select.js
```diff
@@ -149,7 +150,7 @@ export const Select = ({
 
   return (
     <Field hint={hint} error={error} id={generatedId}>
-      <Stack spacing={label || hint || error ? 1 : 0}>
+      <Stack spacing={label || hint || typeof error === 'string' ? 1 : 0}>
```
HichamELBSI, line 153:
```suggestion
      <Stack spacing={label || hint || hasStringError ? 1 : 0}>
```

HichamELBSI: Maybe you can also set the spacing like that but it has to be tested:

```suggestion
      <Stack spacing={label || ariaDescribedBy ? 1 : 0}>
```

HichamELBSI (changes requested):

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: yes I forgot to remove it

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: it was to check the navigation with keyboard, but ok, I can only try with the click

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.js
simotae14: because we had some warnings inside the tests, aria-rowindex is a prop required

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
simotae14: done, thanks @gu-stav

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
simotae14: thanks, just removed

packages/strapi-design-system/yarn.lock
simotae14: I don't know because the hash is the same

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.spec.js
simotae14: done in another way, let me know what do you think

packages/strapi-design-system/src/Select/Select.js
simotae14: I like it, thanks a lot!

packages/strapi-design-system/src/Select/Select.js
simotae14: no because _ariaDescribedBy_ value is
`const ariaDescribedBy = errorMessageId || hintMessageId;`

and _hintMessageId_ value is
`hint ? `${generatedId}-hint` : undefined`

but in the condition we need hint and not the _hintMessageId_ value

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: Basically, I have implemented the navigation with the keyboard because I have found also some bugs inside the Component code with it, but if you are not ok with it I can refactor the code using a simple click event

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: I had to use the skip because in webkit we have the tests stuck, we have the same strange behavior and used skip in another test inside the DS repo

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: Basically, I have implemented the navigation with the keyboard because I have found also some bugs inside the Component code with it, but if you are not ok with it I can refactor the code using a simple click event

I had to use the skip because in webkit we have the tests stuck, we have the same strange behavior and used skip in another test inside the DS repo

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: removed thanks for the comment

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: removed, thanks for the comment

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: ok clear

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: ok I have added a more understandable comment

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: ok I have added a more understandable comment

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: ok I have added a more understandable comment

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: it is passed inside the Field wrapper component

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: ok I have added a more understandable comment

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: ok I have added a more understandable comment

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: ok I have added a more understandable comment

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: maybe we need to refactor also _FlexProps_ and _TypographyProps_

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
```
christiancp100, line 1: Is a separate file for props needed?

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Clear Label used for the X button inside both, the TimePicker and the DatePicker
+   */
+  clearLabel: PropTypes.string,
+  /**
+   * Disabled attribute
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Error message or boolean, if it is a boolean we will hide the error message for the DatePicker and the TimePicker and show just the one that we have defined in the DateTimePicker
```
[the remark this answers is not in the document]
simotae14: basically the error message inside the DateTimePicker is always a boolean, I need to change the comment...It is a boolean because we want to hide the Error messages for the DatePicker and TimePicker

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
simotae14: No, tbh I am not sure it is something that we need inside the DateTimePicker, but maybe you can give me more info about it, I saw you added it in the DateTimePicker inside the helper plugin

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+   */
+  error: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
+  /**
+   * Hint text used inside the Field wrapper component
```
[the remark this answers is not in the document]
simotae14: ok thanks

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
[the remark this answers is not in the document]
simotae14: Sorry, I confused the errors, I can wrote a better comment

packages/strapi-design-system/src/DateTimePicker/DateTimePickerProps.js
```diff
@@ -0,0 +1,80 @@
...
+  step: 1,
+  value: undefined,
+};
+export const dateTimePickerPropTypes = {
+  /**
+   * Aria Label used by the DatePicker and the TimePicker inside the DateTimePicker component
+   */
+  ariaLabel: PropTypes.string,
+  /**
+   * Clear Label used for the X button inside both, the TimePicker and the DatePicker
```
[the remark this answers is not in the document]
simotae14: thanks

packages/strapi-design-system/src/DateTimePicker/parseDate.js
```diff
@@ -0,0 +1,11 @@
+// Return a Date if the passed param is a date valid format
+
+export const parseDate = (date) => {
+  const timestamp = Date.parse(date);
+
+  if (Number.isNaN(timestamp) === false) {
```
christiancp100, line 6:
```suggestion
  if (!Number.isNaN(timestamp)) {
```

packages/strapi-design-system/src/Select/Select.js
```diff
@@ -149,7 +153,7 @@ export const Select = ({
 
   return (
     <Field hint={hint} error={error} id={generatedId}>
-      <Stack spacing={label || hint || error ? 1 : 0}>
+      <Stack spacing={label || hint || hasStringError ? 1 : 0}>
```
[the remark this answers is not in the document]
simotae14: I think the ternary solution is more readable but I like also yours

packages/strapi-design-system/src/DatePicker/DatePicker.js
```diff
@@ -64,7 +64,7 @@ export const DatePicker = ({
           <DatePickerButton
             ref={datePickerButtonRef}
             onClick={toggleVisibility}
-            aria-label={selectedDate ? selectedDateLabel(formatDate(selectedDate, locale)) : label}
+            aria-label={selectedDate ? selectedDateLabel(formatDate(selectedDate, locale)) : label || ariaLabel}
```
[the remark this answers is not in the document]
simotae14: I know it seems strange but I had to add the ariaLabel just for the DateTimePicker case, before we used only the label value

packages/strapi-design-system/src/DateTimePicker/parseDate.js
simotae14: ok

packages/strapi-design-system/src/DatePicker/DatePicker.js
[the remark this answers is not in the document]
christiancp100: I mean doing this, because if you have an aria-label it won't be selected unless there is not label

```suggestion
            aria-label={selectedDate ? selectedDateLabel(formatDate(selectedDate, locale)) : ariaLabel || label}
```

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,148 @@
...
+  onClear,
+  name,
+  required,
+  size,
+  step,
+  value,
+  ...props
+}) => {
+  const initialDate = parseDate(value);
+  const [dateValue, setDateValue] = useState(initialDate);
```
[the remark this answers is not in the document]
simotae14: if you want I can call the parseDate directly inside the useState definition but tbh I think it is better to have two instructions just for readability (and I think it is suggested in the article to declare the useState and useEffect at the beginning of the Component Declaration not at the first line) but thanks for your comment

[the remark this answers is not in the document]
joshuaellis: I agree with @simotae14

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
simotae14: thanks Hicham but as I said in another following comment, the TextInput component gives me error if I don't pass the label

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: done

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: done

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.e2e.js
simotae14: ok

HichamELBSI (approved):
Good job @simotae14 !!

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -0,0 +1,148 @@
...
+            value={timeValue}
+            onChange={handleTimeChange}
+            onClear={onClear && handleTimeClear}
+            clearLabel={clearLabel}
+            disabled={disabled}
+            step={step}
+          />
+        </Stack>
+        <FieldHint />
+        <FieldError id="datetimepicker" />
```
gu-stav, line 140: This static ID needs to be replace with a dynamic one.

packages/strapi-design-system/src/Select/Select.js
```diff
@@ -223,6 +228,7 @@ export const Select = ({
                 onMouseDown={handleMouseDown}
                 tabIndex={-1}
                 disabled={disabled}
+                title="Carret Down Button"
```
gu-stav, line 231: What is this? In any case these need to be translateable ...

gu-stav (changes requested):
Hope the two last (minor) comments :) Great work, @simotae14!

joshuaellis (approved):

packages/strapi-design-system/src/Select/Select.js
simotae14: I need it to fix the ay11 problems

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
simotae14: done

christiancp100 (approved):

packages/strapi-design-system/src/Select/Select.js
gu-stav: I understand that, but this string needs to be translateable. The DS should not assume a language and allow to pass in all localized strings through props.

simotae14: ok now I pass it as a prop

gu-stav (approved):

packages/strapi-design-system/src/DateTimePicker/__tests__/__snapshots__/DateTimePicker.spec.js.snap
```diff
@@ -402,7 +402,7 @@ exports[`DateTimePicker snapshots the component 1`] = `
                     class="c17"
                     id="timepicker-5-content"
                   >
-                    12:00
+                    10:00
```
gu-stav, line 405: Just wondering: why does this need an update suddenly? Is this test flaky?

simotae14: I think the problem was my TMZ config

PR #704, opened 2022-09-26
Fix Breadcrumb with one Crumb

HichamELBSI:
### What does it do?

Fix the breadcrumb when only one link is displayed.

### Why is it needed?

To avoid rendering the `/` when the breadcrumb contains only one link


ronronscelestes (approved):
Nice catch thanks 💅

joshuaellis (approved):

PR #708, opened 2022-09-28
Fix Breadcrumb with conditional links

HichamELBSI:

### What does it do?

Flatten the breadcrumb children and add tests.

### Why is it needed?

To avoid falsy values to be taken into account during the children.map


ronronscelestes (approved):
LGTM, thanks for the fix 🙏

joshuaellis (approved):

PR #709, opened 2022-09-30
Add placement props to Popover component

HichamELBSI:

### What does it do?

Since the Popover is handled by `floating-ui`, we are now able to choose where the popover will be handled. This update will be helpful for the Strapi Cloud app

### DIFF
--- a/packages/strapi-design-system/src/Popover/Popover.js
+++ b/packages/strapi-design-system/src/Popover/Popover.js
@@ -8,6 +8,21 @@ import { Portal } from '../Portal';
 
 import { useIntersection } from '../helpers/useIntersection';
 
+export const POPOVER_PLACEMENTS = [
+  'top',
+  'top-start',
+  'top-end',
+  'right',
+  'right-start',
+  'right-end',
+  'bottom',
+  'bottom-start',
+  'bottom-end',
+  'left',
+  'left-start',
+  'left-end',
+];
+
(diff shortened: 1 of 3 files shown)


packages/strapi-design-system/src/Popover/Popover.js
```diff
@@ -122,6 +133,10 @@ const popoverProps = {
    * The callback invoked after a scroll to the bottom of the popover content.
    */
   onReachEnd: PropTypes.func,
+  /**
+   * The popover position
+   */
+  placement: PropTypes.string,
```
joshuaellis, line 139: I think this prop can be better defined, they're specific options available:

```ts
type Placement =
  | 'top'
  | 'top-start'
  | 'top-end'
  | 'right'
  | 'right-start'
  | 'right-end'
  | 'bottom'
  | 'bottom-start'
  | 'bottom-end'
  | 'left'
  | 'left-start'
  | 'left-end';
```

joshuaellis (changes requested):
Reckon we can be tighter on PropTypes but otherwise looks good, glad this library addition has helped 🤙🏻

packages/strapi-design-system/src/Popover/Popover.js
HichamELBSI: Awesome thank you, I was looking for this 😄

joshuaellis (approved):

PR #711, opened 2022-10-05
[fix][Modal | Dialog]: Add dismissible layer component

joshuaellis:
### What does it do?

Add a new component to the `design-system` that handles "Dismissible Layers" commonly found with modals / dialogs etc. This component handles global listeners for `escape` and `pointerDown` the latter is important because a "click" event causes a false closure, see https://github.com/strapi/strapi/issues/13044#issuecomment-1225900907

### Why is it needed?

* Modals shouldn't close on a click event you want to know _where_ the mouse pressed down not where it finished.

### Additional thoughts

There's more you _can_ do in this area, but I see this as a temporary fix whilst we decide on the progression of this lib.

### Related issue(s)/PR(s)

* partially https://github.com/strapi/strapi/issues/13044


packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
```diff
@@ -0,0 +1,73 @@
...
+  onPointerDownOutside: PropTypes.func,
+};
+
+/**
+ * A custom hook that converts a callback to a ref to avoid triggering re-renders when passed as a
+ * prop or avoid re-executing effects when passed as a dependency
+ *
+ * Stolen from @radix-ui/react-use-callback-ref
+ */
+function useCallbackRef(callback) {
```
ronronscelestes, line 59: do you think it would be worth adding it to DS utils? @joshuaellis

ronronscelestes (commented):
works well on my side 💃

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
joshuaellis: I haven't used it in other components, so I suppose the question is "is there a need for it"? Happy to take the consensus of the group 👍🏼

joshuaellis:
Also should this be added to other components e.g. Popover? IIRC those menus behave similarly, they're supposed to have a focus capture and the same dismissible properties...?

ronronscelestes:
> Also should this be added to other components e.g. Popover? IIRC those menus behave similarly, they're supposed to have a focus capture and the same dismissible properties...?

Sounds good indeed 👀

joshuaellis:
> > Also should this be added to other components e.g. Popover? IIRC those menus behave similarly, they're supposed to have a focus capture and the same dismissible properties...?
> 
> Sounds good indeed 👀

I agree @ronronscelestes, @gu-stav what do you think?

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
```diff
@@ -0,0 +1,73 @@
...
+      {children}
+    </div>
+  );
+};
+
+DismissableLayer.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  onEscapeKeyDown: PropTypes.func,
+  onPointerDownOutside: PropTypes.func,
```
gu-stav, line 50: Are these required for the component to work?

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
gu-stav: As long as we only need it here, I'm okay with keeping it local. Tests (for the whole component) would be very nice though :)

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
```diff
@@ -0,0 +1,73 @@
...
+
+  return (
+    <div ref={layerRef} className={className}>
+      {children}
+    </div>
+  );
+};
+
+DismissableLayer.propTypes = {
+  children: PropTypes.node,
```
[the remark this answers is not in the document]
joshuaellis: It could be, my view on required is:
* Does it need a default?
* Will the component break without it
So i'd say it doesn't have to be, but if you think that's more inline with other components then it can be?

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
joshuaellis: The component will not crash if you don't pass them. But equally the behaviour will not be what you expect if you don't pass them, additionally we can't check that you even do what we expect (close the layer) with those functions, so imo as long as the component doesn't crash we're all good.

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
```diff
@@ -0,0 +1,73 @@
...
+      }
+    };
+
+    document.addEventListener('pointerdown', handlePointerDownOutside);
+
+    return () => document.removeEventListener('pointerdown', handlePointerDownOutside);
+  }, [onPointerDownOutsideHandler]);
+
+  return (
+    <div ref={layerRef} className={className}>
```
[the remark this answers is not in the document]
joshuaellis: We can if you prefer, I find spreading all the props in a non-typed env vague, it sort of suggests you can pass what you feel like and it gets added to the `div` but in reality, we don't want things like `style` to be passed imo. wdyt?

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
```diff
@@ -0,0 +1,73 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
joshuaellis: Sure thing 👍🏼 This is a hangover from using TS so much (there was an issue with importing & types)

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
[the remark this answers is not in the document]
gu-stav: I see what you mean: is it necessary to pass in a `className`? If not we could remove the prop and would have a nice and strict interface :)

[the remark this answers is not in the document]
joshuaellis: I think it is necessary because otherwise you can't do `styled(DismissibleLayer)`, which I think is important to avoid unnecessary DOM nesting 🤔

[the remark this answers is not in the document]
gu-stav: I see. Let's keep it the way you build it then 👍🏼

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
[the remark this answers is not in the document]
gu-stav: Yeah I think it would be beneficial to see from the prop-types that the component should have children and I think that is the way we handle it elsewhere in the DS.

packages/strapi-design-system/src/DismissableLayer/DimissableLayer.js
gu-stav: Thanks. If the component behaves unexpectedly, let's make these required and strict (we can open it anytime) and avoid the cost of having them optional. Wdyt?

joshuaellis:
yep I agree, made this change.

packages/strapi-design-system/src/ModalLayout/__tests__/ModalLayout.spec.js
```diff
@@ -462,9 +462,7 @@ describe('ModalLayout', () => {
         max-height: 60vh;
       }
 
-      <body
-        class="lock-body-scroll"
-      >
+      <body>
```
gu-stav, line 465: I'm wondering about this change. According to `useLockScroll()` the class should still be there, don't you think?

gu-stav (approved):
Added one last (TM) question. If you think it is not a problem I'm good with merging this PR :) Thanks for working on it. That's a valuable addition! 🌻

packages/strapi-design-system/src/ModalLayout/__tests__/ModalLayout.spec.js
joshuaellis: no you're right, I didn't realise `useLockScroll` required an argument whether to do it or not. I've set this to true because Modals must be removed from the tree unlike `Dialog`.

PR #712, opened 2022-10-06
fix: link active & hover not affecting Typography

joshuaellis:
### What does it do?

Specifically targets the `Typography` class nested in the `Link` component

### Why is it needed?

Styles applied for `hover` and `active` were not affecting the text color accurately.

### Related issue(s)/PR(s)

CONTENT-477

### DIFF
--- a/packages/strapi-design-system/src/Link/Link.js
+++ b/packages/strapi-design-system/src/Link/Link.js
@@ -12,18 +12,34 @@ const LinkWrapper = styled.a`
   align-items: center;
   text-decoration: none;
   pointer-events: ${({ disabled }) => (disabled ? 'none' : undefined)};
+
+  ${Typography} {
+    transition: color 150ms ease-out;
+  }
+
   svg path {
     fill: ${({ disabled, theme }) => (disabled ? theme.colors.neutral600 : theme.colors.primary600)};
   }
+
   svg {
     font-size: ${10 / 16}rem;
   }
-  &:hover,
-  &:active {
-    color: ${({ theme }) => theme.colors.primary800};
+
+  &:hover {
+    ${Typography} {
+      color: ${({ theme }) => theme.colors.primary500};
+    }
   }
+
+  &:active {
+    ${Typography} {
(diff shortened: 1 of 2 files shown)


packages/strapi-design-system/src/Link/Link.js
```diff
@@ -12,18 +12,34 @@ const LinkWrapper = styled.a`
   align-items: center;
   text-decoration: none;
   pointer-events: ${({ disabled }) => (disabled ? 'none' : undefined)};
+
+  ${Typography} {
```
ronronscelestes, line 16: 😍

packages/strapi-design-system/src/Link/Link.js
```diff
@@ -12,18 +12,34 @@ const LinkWrapper = styled.a`
...
   }
+
   svg {
     font-size: ${10 / 16}rem;
   }
-  &:hover,
+
+  &:hover {
+    ${Typography} {
+      color: ${({ theme }) => theme.colors.primary500};
```
ronronscelestes, line 30: should it not impact the icons too? 👀

image
image

maevalienard: That would be nice indeed if feasible

joshuaellis: I wasn't sure and thanks both for letting me know, i'll make those changes 😎

gu-stav: You also may want to add `:focus,`?

packages/strapi-design-system/src/Link/Link.js
```diff
@@ -12,18 +12,33 @@ const LinkWrapper = styled.a`
...
     font-size: ${10 / 16}rem;
   }
-  &:hover,
+
+  ${Typography} {
+    transition: color 150ms ease-out;
+    color: currentColor;
+  }
+
+  &:hover {
```
gu-stav, line 31: Do you think we should add `&:focus` too?

ronronscelestes (approved):
LGTM (the transition ❤️)

Issue #692, opened 2022-09-11
Unbroken CarouselImage alt text overflows out of CarouselInput


rmirville:
### Required System information

- Npm version: 8.15.1
- NodeJS version: 16.16.0
- Strapi Design System version: 1.2.2
- Browser: Chrome v105.0.5195.102

### Describe the bug

When a `CarouselInput`'s slide image src can't be found, long alt text overflow out of the image.

### Steps to reproduce the behavior

Sandbox: https://codesandbox.io/s/affectionate-austin-bv7un1?file=/src/App.js

1. Create a `CarouselInput`, with `selectedSlide={0}` and a fixed css width.
2. Inside the `CarouselInput`, make a `CarouselSlide` containing a `CarouselImage`.
3. Give the `CarouselImage`:
    * an invalid src attribute, and
    * alt text with one, unbroken string longer than the `CarouselInput` width.

### Expected behavior

The text should wrap within the bounds of the image.

### Screenshots

![screenshott](https://user-images.githubusercontent.com/1623207/189540677-daa6d6a5-626d-4113-8e1e-0e8d80fbc47e.png)

### Code snippets

```jsx
import {
  CarouselImage,
  CarouselInput,
  CarouselSlide
} from "@strapi/design-system/CarouselInput";

export default function App() {
  return (
    <CarouselInput selectedSlide={0} style={{ width: "242px" }}>
      <CarouselSlide label="1 of 1 slides">
        <CarouselImage
          src="invalid.jpg"
          alt="abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz"
        />
      </CarouselSlide>
    </CarouselInput>
  );
}
```

### Additional context

This bug is a replication of [strapi/strapi issue 14324](https://github.com/strapi/design-system/issues/722)

vaibhav-305:
Hey, I would like to work on this issue, can u assign me this

INNOVATIVEGAMER:
Thank you @rmirville for reporting this issue.
I think there are multiple solutions exist here =>
1. We just hide the overflown text
2. We add a horizontal scroll so that overflown text can be read.
3. We can wrap the text in box if it still overflows => 1. We can hide the overflown text 2. We can add vertical scroll bar for overflown text.

My suggestion would be 3.1
As images alt text are not that long and if they are still that big they can't fit into box, they can be hidden because adding a scroll bar to the current design may destroy a lot of structure
ex.- 
[...]

PR #713, opened 2022-10-06
[Fix]: unbroken carousel image alt text overflows out of carousel input

INNOVATIVEGAMER:
### What does it do?

Added ImageBase styled for Box if Box has as("img") attribute to show alt text (ellipsis for long texts) & added tooltip over carousel image for alt text.

### Why is it needed?

When a CarouselInput's slide image src can't be found, long alt text overflow out of the image.

### How to test it?

In CarouselInput replace src with an empty string and place long alt text.

### Related issue(s)/PR(s)

Design-System #692

Screenshot (41)

### DIFF
--- a/packages/strapi-design-system/src/Box/Box.js
+++ b/packages/strapi-design-system/src/Box/Box.js
@@ -1,5 +1,6 @@
 import styled from 'styled-components';
 import handleResponsiveValues from '../helpers/handleResponsiveValues';
+import { ellipsisStyle, handleColor } from '../Typography/utils';
 import { boxPropTypes, boxDefaultProps } from './BoxProps';
 
 /**
@@ -9,7 +10,20 @@ const transientProps = {
   color: true,
 };
 
-export const Box = styled.div.withConfig({
+// ImageBase has been added here to check if Box is used as "img" tag
+// If so, to handle alt text overflow following properties has been added
+const ImageBase = styled.div`
+  ${({ as, theme, color }) => {
+    if (as === 'img')
+      return `
+    width: stretch;
+    ${ellipsisStyle({ ellipsis: true })}
+    color: ${handleColor({ theme, textColor: color })};
+    `;
+  }};
+`;
+
+export const Box = styled(ImageBase).withConfig({
   shouldForwardProp: (prop, defPropValFN) => !transientProps[prop] && defPropValFN(prop),
 })`
   // Font
--- a/packages/strapi-design-system/src/Box/Box.js
+++ b/packages/strapi-design-system/src/Box/Box.js
@@ -1,6 +1,5 @@
 import styled from 'styled-components';
 import handleResponsiveValues from '../helpers/handleResponsiveValues';
-import { ellipsisStyle, handleColor } from '../Typography/utils';
 import { boxPropTypes, boxDefaultProps } from './BoxProps';
 
 /**
@@ -10,20 +9,7 @@ const transientProps = {
   color: true,
 };
 
-// ImageBase has been added here to check if Box is used as "img" tag
-// If so, to handle alt text overflow following properties has been added
-const ImageBase = styled.div`
-  ${({ as, theme, color }) => {
-    if (as === 'img')
-      return `
-    width: stretch;
-    ${ellipsisStyle({ ellipsis: true })}
-    color: ${handleColor({ theme, textColor: color })};
-    `;
-  }};
-`;
-
-export const Box = styled(ImageBase).withConfig({
+export const Box = styled.div.withConfig({
   shouldForwardProp: (prop, defPropValFN) => !transientProps[prop] && defPropValFN(prop),
 })`
   // Font
--- a/packages/strapi-design-system/src/CarouselInput/CarouselImage.js
+++ b/packages/strapi-design-system/src/CarouselInput/CarouselImage.js
@@ -24,6 +24,6 @@ CarouselImage.defaultProps = {
 };
 
 CarouselImage.propTypes = {
-  src: PropTypes.string,
   alt: PropTypes.string,
+  src: PropTypes.string,
 };
--- a/packages/strapi-design-system/src/CarouselInput/CarouselImage.js
+++ b/packages/strapi-design-system/src/CarouselInput/CarouselImage.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { Box } from '../Box';
 import styled from 'styled-components';
@@ -19,13 +19,22 @@ const StyledImage = styled(Box)`
   `};
 `;
 
-export const CarouselImage = (props) => (
-  <Tooltip description={props.alt ?? ''}>
-    <ImageBase>
-      <StyledImage as="img" {...props} />
-    </ImageBase>
-  </Tooltip>
-);
+export const CarouselImage = (props) => {
+  const [isError, setIsError] = useState(false);
+
+  const handleImageError = () => setIsError(true);
+
+  if (isError)
+    return (
+      <Tooltip description={props.alt ?? ''}>
+        <ImageBase>
+          <StyledImage as="img" {...props} onError={handleImageError} />
+        </ImageBase>
+      </Tooltip>
+    );
+
+  return <StyledImage as="img" {...props} onError={handleImageError} />;
+};
 
 CarouselImage.defaultProps = {
   src: undefined,
--- a/packages/strapi-design-system/src/CarouselInput/CarouselImage.js
+++ b/packages/strapi-design-system/src/CarouselInput/CarouselImage.js
@@ -1,4 +1,29 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 import { Box } from '../Box';
+import styled from 'styled-components';
+import { Tooltip } from '../Tooltip';
 
-export const CarouselImage = (props) => <Box as="img" height="100%" {...props} />;
+// ImageBase is needed since tooltip requires box level element to work upon
+// Also img component can not be directly worked with tooltip otherwise it will span the whole window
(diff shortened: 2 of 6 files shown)


packages/strapi-design-system/src/Box/Box.js
```diff
@@ -9,7 +10,20 @@ const transientProps = {
   color: true,
 };
 
-export const Box = styled.div.withConfig({
+// ImageBase has been added here to check if Box is used as "img" tag
+// If so, to handle alt text overflow following properties has been added
+const ImageBase = styled.div`
```
joshuaellis, line 15: I wouldn't put this here, see comment on the component.

INNOVATIVEGAMER:
@joshuaellis Can you please comment above what to do next ?

packages/strapi-design-system/src/CarouselInput/CarouselImage.js
```diff
@@ -1,4 +1,29 @@
...
+// ImageBase is needed since tooltip requires box level element to work upon
+// Also img component can not be directly worked with tooltip otherwise it will span the whole window
+const ImageBase = styled.div`
+  display: grid;
+`;
+
+export const CarouselImage = (props) => (
+  <Tooltip description={props.alt ?? ''}>
+    <ImageBase>
+      <Box as="img" {...props} />
```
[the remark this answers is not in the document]
joshuaellis: I prefer to take the reactive approach as opposed to the preemptive approach to reduce complexities, so for now lets localise it to the component.

joshuaellis:
> @joshuaellis Can you please comment above what to do next ?

Hey @INNOVATIVEGAMER, unfortunately managing community PRs / Issues isn't the only thing I have to do at work so please do be patient with waiting for feedback, I haven't forgotten about you 😄 I've left a comment inline but there is another comment I left you didn't reply too.

Typically you could be waiting a day or two for a reply just a heads up.

packages/strapi-design-system/src/CarouselInput/CarouselImage.js
```diff
@@ -1,4 +1,29 @@
...
+import PropTypes from 'prop-types';
 import { Box } from '../Box';
+import styled from 'styled-components';
+import { Tooltip } from '../Tooltip';
 
-export const CarouselImage = (props) => <Box as="img" height="100%" {...props} />;
+// ImageBase is needed since tooltip requires box level element to work upon
+// Also img component can not be directly worked with tooltip otherwise it will span the whole window
+const ImageBase = styled.div`
+  display: grid;
```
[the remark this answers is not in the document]
INNOVATIVEGAMER: Ya I will update this also

INNOVATIVEGAMER:
Sorry @joshuaellis . I will be more patient onwards. I just sometimes forget this is a community PR and not my official work. Again sorry for the inconvenience.

I will update PR with the changes you asked fixing it only on the component level.

packages/strapi-design-system/src/CarouselInput/CarouselImage.js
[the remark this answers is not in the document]
INNOVATIVEGAMER: Replaced with `styled(Box)`. As for `display: grid`. It is to do something with Tooltip span and img in its child. 
For all other display options for div between Tooltip and img tag. Image was getting spanned over entire screen but only with grid option it work as expected.

Just another day of war with CSS 😭

joshuaellis (commented):
So i don't think the implementation is wrong per se, but the issue is now when you hover over the image you always get the tooltip 🤔 @gu-stav be helpful to get your recommendation on this one?

INNOVATIVEGAMER:
@joshuaellis @gu-stav 
So I think there are 2 ways ahead =>
1. Check if the `img` has valid `src` attribute using `error` event handler on the component, if not then show the tooltip (alt text)
2. Show tooltip (alt text) all the time

wdyt ?

INNOVATIVEGAMER:
I think @maevalienard can comment here as it is kind of small design update

joshuaellis:
> Check if the img has valid src attribute using error event handler on the component, if not then show the tooltip (alt text)

Yes, I would use the `onError` handler to check when there's an issue then ensure the text behaves as we expect.

maevalienard:
Agree with Josh here, I wouldn't display the alt text all the time at hover. This is "just" the alt text, I would rather have the name of the asset on hover than the alt text for instance (not saying that we have to do that, it was just an example) :)

INNOVATIVEGAMER:
@joshuaellis Made the requested changes. Thank you for helping me finding the issues in implementation :)

packages/strapi-design-system/src/CarouselInput/CarouselImage.js
```diff
@@ -1,4 +1,47 @@
...
+export const CarouselImage = (props) => {
+  const [isError, setIsError] = useState(false);
+
+  const handleImageError = () => setIsError(true);
+
+  if (isError)
+    return (
+      <Tooltip description={props.alt ?? ''}>
+        <ImageBase>
+          <StyledImage as="img" {...props} onError={handleImageError} />
```
joshuaellis, line 31: I think this would incur an infinite loop, whereby the image you're continuously trying to load will call `setIsError` imo this handler can be omitted.

packages/strapi-design-system/src/CarouselInput/CarouselImage.js
```diff
@@ -1,4 +1,47 @@
...
+};
+
+CarouselImage.defaultProps = {
+  src: undefined,
+  alt: undefined,
+};
+
+CarouselImage.propTypes = {
+  alt: PropTypes.string,
+  src: PropTypes.string,
```
joshuaellis, line 46: Both of these should be required.

joshuaellis (changes requested):
Other than the minor comments this looks fine, I think you should add some `unit` tests that test the behaviour:

* If an image load successfully there is no tooltip
* If an image fails to load there is a tooltip

packages/strapi-design-system/src/CarouselInput/CarouselImage.js
INNOVATIVEGAMER: This will not result in infinite loop since state goes from false -> false.
But I also think this is unnecessary so will remove it.

packages/strapi-design-system/src/CarouselInput/CarouselImage.js
INNOVATIVEGAMER: Noted.

INNOVATIVEGAMER:
@joshuaellis I need some help writing the test cases. Can you please help me here.
Test case =>
```
it('Show tooltip when CarouselImage source is not working', async () => {
    render(
        <CarouselInput
          label={`Carousel of numbers 1/1)`}
          selectedSlide={0}
          previousLabel="Previous slide"
          nextLabel="Next slide"
          onNext={() => {}}
          onPrevious={() => {}}
          hint="Description line"
          style={{ width: '242px' }}
        >
          <CarouselSlide label="1 of 1 slides">
            <CarouselImage
              src={''}
              data-testid="carousel image"
              alt="Lorem ipsum dolor, sit amet consectetur adipisicing elit. Modi, nihil."
            />
          </CarouselSlide>
        </CarouselInput>
      </ThemeProvider>,
    );

    fireEvent.focus(screen.getByTestId('carousel image'));

    const tooltipText = await screen.findByText(
      'Lorem ipsum dolor, sit amet consectetur adipisicing elit. Modi, nihil.',
    );

    expect(tooltipText).toBeInTheDocument();
  });
```

This test case is always failing saying `Error: Unable to find an element with the text: Lorem ipsum dolor, sit amet consectetur adipisicing elit. Modi, nihil.. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.`

I don't know where I am doing wrong. I also tried `fireEvent.mouseOver()` but it also results in the same output.

joshuaellis:
Off the top of my head the test probably wants to be something like this:

```ts
it('shows a tooltip with the alt text when the image fails to load', async () => {
    render(
            <CarouselImage
              src={''}
              alt="image"
            />
      </ThemeProvider>,
    );
	
	const imgEl = screen.getByRole('img')

	fireEvent.mouseOver(imgEl)

    const tooltipText = await waitFor(() => expect(screen.findByText('image')).toBeTruthy())
  });
```

I think? Basically you need to wait for the tooltip too show, but i would also simplify the test by a lot, lets not over complicate testing behaviour.

INNOVATIVEGAMER:
> ```ts
> expect(screen.findByText('image')).toBeTruthy()
> ```

This results in always passing test case as `findBy` methods return a promise which is always truthful.
Also the `tooltipText` shows `undefined` when logged.
This is where I am struggling. We need to wait to tooltip to show. I think this is because there is some delay in showing tooltip.
I also referred this answer but not successful in our case though => https://stackoverflow.com/a/66442091

joshuaellis:
I've added a test case, it needed to be e2e because JSDOM doesn't try to load resources (e.g. an image). there was also a bug in safari where the image would disappear when rendered in the tooltip so i've made the necessary changes to resolve this.

joshuaellis (approved):

INNOVATIVEGAMER:
Thank you @joshuaellis for your support in this PR. Sorry that you need to push some final fixes by yourself 😓.
The journey of this PR went from simple fix to complex test cases though. I have learnt a lot from my first PR in strapi-design-system working with you 😄.

joshuaellis:
> Thank you @joshuaellis for your support in this PR. Sorry that you need to push some final fixes by yourself 😓.
> 
> The journey of this PR went from simple fix to complex test cases though. I have learnt a lot from my first PR in strapi-design-system working with you 😄. 

All good, thanks for getting the ball rolling and keeping it going! 

Congratulations on your PR being merged 💪🏼

Issue #661, opened 2022-07-25
MenuItem as Link component does not support external links


HichamELBSI:
### Required System information

- Npm version: 8.5.0
- NodeJS version: 16.4.2
- Strapi Design System version: 1.2.0
- Browser: Chrome

### Describe the bug

The MenuItem component support internal links but not externals.

### Steps to reproduce the behavior

1. Use the SimpleMenu and MenuItem components
2. Try to add a href/isExternal props to use the MenuItem component as an external link.

INNOVATIVEGAMER:
hey @joshuaellis Can you please assign this to me?
Also, I think it's an enhancement and not a bug since this will add external link support to the current "MenuItem" component.

Tasks =>
1. Add support for external links to "MenuItem" component.
2. Add/Update documentation for the same in storybook
3. Add test cases to test the working of the external link

joshuaellis:
Hey @INNOVATIVEGAMER 

Tasks sound good, I'd probably remove the ternary operator for an `if` statement and look for the `isExternal` prop to know if it should be an anchor.

PR #714, opened 2022-10-06
Added support for external link in simple menu

INNOVATIVEGAMER:
### What does it do?

1. Added `isExternal` , `href` props to be supoorted by `MenuItem` component in `SimpleMenu`.
2. Added testcase to test external link is working or not in SimpleMenu test suite.
3. Updated story `With links` of SimpleMenu to document newly added behavior of external link.

### Why is it needed?

Added support for external link in simple menu

### How to test it?

Added testcase in SimpleMenu.spec.js to test the behaviour.
Also you can test it by simple adding `href` & `isExternal` attributes to any `MenuItem` component. By clicking on such button user will be navigated to that link in another tab.

### Related issue(s)/PR(s)

Strapi Design System - #661 

Screenshot (44)


ronronscelestes (commented):
OptionLink is based on BaseLink, which already supports external links

```js
<MenuItem isLink isExternal href="https://strapi.io/">Here</MenuItem>
```

image

maybe we could improve the documentation of the component instead?

joshuaellis:
> OptionLink is based on BaseLink, which already supports external links

Is it? It's a `styled(NavLink)` looking at `SimpleMenu.js`.

ronronscelestes:
> Is it? It's a `styled(NavLink)` looking at `SimpleMenu.js`.

ahh was looking at v2/SimpleMenu API 😶‍🌫️

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -25,6 +26,11 @@ const OptionLink = styled(NavLink)`
   ${getOptionStyle}
 `;
 
+const OptionExternalLink = styled.div`
```
ronronscelestes, line 29: I think we could use the `BaseLink` as it is done in `v2/SimpleMenu`, wdyt?

```js
const OptionExternalLink = styled(BaseLink)`
  text-decoration: none;
  ${getOptionStyle}
`;
```

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -60,30 +66,47 @@ export const MenuItem = ({ children, onClick, to, isFocused, ...props }) => {
...
+        <Box padding={2}>
+          <Typography>{children}</Typography>
+        </Box>
+      </OptionLink>
+    );
+  else if (href && isExternal)
+    return (
+      <OptionExternalLink {...menuItemProps}>
+        <Box padding={2}>
+          <Link href={href} isExternal={isExternal}>
```
ronronscelestes, line 81: following my previous comment, wdyt about: 
```js
    <OptionExternalLink href={href} isExternal={isExternal} {...menuItemProps}>
      <Box padding={2}>
        <Typography>{children}</Typography>
      </Box>
    </OptionExternalLink>
```

ronronscelestes (changes requested):
thank you for working on this 😊

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
INNOVATIVEGAMER: Ya we can do that. Just a small difference there will be though.
We will not get that style for external link and icon as I have shown in screenshot if we use `BaseLink`.
It will be a shown as just simple link.

I think current implementation with icon and different style is more user friendly though ;)

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
INNOVATIVEGAMER: Ya we can do that.
Please reply to above comment so we can proceed with final solution.

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
ronronscelestes: mmh I see - I would say that we should have the same design as the v2 version though, so without external link styling 🤔
if we decide to change this one I guess we would need to update the v2 version?
that would be a design decision for @maevalienard

maevalienard: I like it with the icon, I think we understand better what will be the action :) The rest is fine to me. Thanks!

ronronscelestes: thank you @maevalienard 
I guess we should then add the same icon when external links for `v2/SimpleMenu`
wdyt @INNOVATIVEGAMER?

INNOVATIVEGAMER: Ya, Sure. I will update the PR.

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -60,30 +66,47 @@ export const MenuItem = ({ children, onClick, to, isFocused, ...props }) => {
...
-  );
+  if (to)
+    return (
+      <OptionLink to={to} {...menuItemProps}>
+        <Box padding={2}>
+          <Typography>{children}</Typography>
+        </Box>
+      </OptionLink>
+    );
+  else if (href && isExternal)
```
ronronscelestes, line 77: after checking `v1` of `Link` component doesn't hold any `isExternal` logic 👀
```js
  else if (href)
    return (
      <OptionExternalLink {...menuItemProps}>
        <Box padding={2}>
          <ExternalLink href={href}>
            <Typography textColor="danger200">{children}</Typography>
          </ExternalLink>
        </Box>
      </OptionExternalLink>
    );
```

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
ronronscelestes, line 81: we should try to keep the default MenuItem styling here, one solution could be: 
```js
const ExternalLink = styled(Link)`
  color: currentColor;
  :hover {
    color: currentColor;
  }
`;
```

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
```diff
@@ -6,11 +6,12 @@ import { Typography } from '../../Typography';
 import { Box } from '../../Box';
 import { Flex } from '../../Flex';
 import { Button } from '../../Button';
-import { BaseLink } from '../../BaseLink';
 import { Popover, POPOVER_PLACEMENTS } from '../../Popover';
 import { getOptionStyle } from './utils';
 import { useId } from '../../helpers/useId';
 import { KeyboardKeys } from '../../helpers/keyboardKeys';
+import { Link } from '../../Link';
+import { NavLink } from 'react-router-dom';
```
ronronscelestes, line 14: `NavLink` shouldn't be included in the `v2` components - the idea of v2 components was removing the dependency to `react-router-dom` as it is not compatible with other routing systems (e.g. Next)

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
```diff
@@ -6,11 +6,12 @@ import { Typography } from '../../Typography';
 import { Box } from '../../Box';
 import { Flex } from '../../Flex';
 import { Button } from '../../Button';
-import { BaseLink } from '../../BaseLink';
 import { Popover, POPOVER_PLACEMENTS } from '../../Popover';
 import { getOptionStyle } from './utils';
 import { useId } from '../../helpers/useId';
 import { KeyboardKeys } from '../../helpers/keyboardKeys';
+import { Link } from '../../Link';
```
ronronscelestes, line 13: we should use the `v2/Link` not the `v1/Link`

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
```diff
@@ -20,7 +21,12 @@ const OptionButton = styled.button`
   ${getOptionStyle}
 `;
 
-const OptionLink = styled(BaseLink)`
+const OptionLink = styled(NavLink)`
```
ronronscelestes, line 24: we should use `BaseLink` instead of `NavLink` here

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
```diff
@@ -60,33 +66,50 @@ export const MenuItem = ({ as, children, onClick, isFocused, isLink, ...props })
...
-      </Box>
-    </OptionLink>
-  ) : (
-    <OptionButton onKeyDown={handleKeyDown} onMouseDown={onClick} type="button" {...menuItemProps}>
-      <Box padding={2}>
-        <Typography>{children}</Typography>
-      </Box>
-    </OptionButton>
-  );
+  if (to)
```
ronronscelestes, line 69: I'm starting to think we won't be able to differentiate internal and external link with a prop like `to` or `href`
the all idea of this component was to remove any dependency to `react-router-dom`
with this new structure we wouldn't support routing system like Next's which we should do as described by the DS documentation:
```js

import { SimpleMenu, MenuItem } from '@strapi/design-system/SimpleMenu';
import NextLink from 'next/link';

<SimpleMenu label={Label}>
  <NextLink href="/home" passHref>
    <MenuItem>Home</MenuItem>
  </NextLink>
</SimpleMenu>;
```
_in this case would it mean that we render `OptionButton` instead of `OptionLink`?_

sorry I'm only pointing to the issue without offering a solution right now - I still need to think a little about it but I thought it would be worth sharing - maybe you or another team member would have an idea 🤔

ronronscelestes (commented):
hey, thanks again for working on this improvement @INNOVATIVEGAMER 
I commented with some issues I've seen but I don't have solutions in mind for some of them right now - I'll keep looking into it, let me know what you think 😊

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
INNOVATIVEGAMER: Ok, previously I didn't know about the `react-router-dom` removal in `v2`.
 I think the solution would be =>
1. `isExternal` should be the only prop which is deciding factor for external & internal links styling.
So this will update the `Link` component in v1 and v2 accordingly for styling.

2. For React use case, we will pass `href` and 'to' props as they are getting passed in `SimpleMenu` (my implementation)
3. For Nextjs links works outside our implementation and only styling is in our hands. So we tackled that using 1st point.

@ronronscelestes wdyt about this ?

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
INNOVATIVEGAMER: Ya, I have checked recently. It is written in storybook explanation but not implemented in code. I will add the implementation.

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
INNOVATIVEGAMER: Noted.

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
INNOVATIVEGAMER: Noted.

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
INNOVATIVEGAMER: Noted.

packages/strapi-design-system/src/Link/Link.js
```diff
@@ -43,7 +43,7 @@ const IconWrapper = styled(Box)`
   display: flex;
 `;
 
-export const Link = ({ href, to, children, disabled, startIcon, endIcon, ...props }) => {
+export const Link = ({ href, to, children, disabled, startIcon, endIcon, isExternal, ...props }) => {
```
ronronscelestes, line 46: Some comments on `Link`:
- Currently, it's not possible to tab to the `Link` in `SimpleMenu` because the ref isn't passed which needs to be fixed in SimpleMenu v1/v2 components, for Link v1 you will need to accept a `ref` with `forwardRef` and pass it to `LinkWrapper`
- I think we should remove `isExternal` here as we don't need it, this component use `react-router` so we only need to check on to or `href` prop `to` know if we are in an internal or external context. Plus it would be a breaking change since all `Link` used for external links without this prop in Strapi would lose their icon. Can you remove `isExternal` from prop and fix the condition for icon appearance below please?
```js
{href && (
   <IconWrapper as="span" aria-hidden={true} paddingLeft={2}>
       <ExternalLink />
   </IconWrapper>
)}
```
- the style of an external `Link` is broken when the element is focused in `SimpleMenu` because of `buttonFocusStyle` which I think we don't need here, we can rely on `ThemeProvider` `*:focus-visible` style and add the border to `LinkWrapper` instead of `buttonFocusStyle`
```js
const LinkWrapper = styled.a`
  ...
  ${buttonFocusStyle}; <== to remove

  &:focus-visible {
    border-radius: ${({ theme }) => theme.borderRadius};
  } <=== to add 
`;
```

packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -38,7 +48,7 @@ const StyledButtonSmall = styled(Button)`
   padding: ${({ theme }) => `${theme.spaces[1]} ${theme.spaces[3]}`};
 `;
 
-export const MenuItem = ({ children, onClick, to, isFocused, ...props }) => {
+export const MenuItem = ({ children, onClick, to, isFocused, href, isExternal, ...props }) => {
```
ronronscelestes, line 51: Some comments on `SimpleMenu` v1:
- padding/style is currently broken for external `Links` because the padding doesn't take the icon into account, to fix it I propose:
```js
const OptionExternalLink = styled(Link)`
  padding: ${({ theme }) => theme.spaces[2]};
  ...
`;

...

<OptionExternalLink href={href} {...menuItemProps}>
  <Typography>{children}</Typography>
</OptionExternalLink>
```
- sorry if I was unclear about `isExternal` in my last comments, I meant that we don't need `isExternal` in `Link` v1 or `SimpleMenu` v1, since we integrate react-router we only need to and `href` prop `to` know if it is an internal or external link that we want. Could you remove isExternal from `MenuItem` v1 prop + remove it from the condition when rendering?
```js
 else if (href)
    return (
      <OptionExternalLink href={href} {...menuItemProps}>
        <Typography>{children}</Typography>
      </OptionExternalLink>
    );
```
- I'm not sure why you added a condition to properly pass the ref to each items, we need it - without it focus/tabbing won't work
```js
  const menuItemProps = {
    tabIndex: isFocused ? 0 : -1,
    ref: menuItemRef,
    role: 'menuitem',
    ...props,
  };
```

packages/strapi-design-system/src/v2/Link/Link.js
```diff
@@ -33,9 +33,9 @@ const IconWrapper = styled(Box)`
   display: flex;
 `;
 
-export const Link = React.forwardRef(({ children, href, disabled, startIcon, endIcon, ...props }, ref) => {
+export const Link = React.forwardRef(({ children, href, disabled, startIcon, endIcon, isExternal, ...props }, ref) => {
```
ronronscelestes, line 36: Some comments on `Link` v2:
- could you please fix the outline + icon color issue we can see in SimpleMenu v2 with:
```js
const LinkWrapper = styled(BaseLink)`
  ...

  color: ${({ disabled, theme }) => (disabled ? theme.colors.neutral600 : theme.colors.primary600)};

  svg path {
    transition: fill 150ms ease-out;
    fill: currentColor;
  }

  ...

  ${buttonFocusStyle}; <=== to remove

  &:focus-visible {
    border-radius: ${({ theme }) => theme.borderRadius};
  }
`;
```

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
```diff
@@ -38,7 +47,7 @@ const StyledButtonSmall = styled(Button)`
   padding: ${({ theme }) => `${theme.spaces[1]} ${theme.spaces[3]}`};
 `;
 
-export const MenuItem = ({ as, children, onClick, isFocused, isLink, ...props }) => {
+export const MenuItem = ({ as, children, onClick, isFocused, to, href, isExternal, ...props }) => {
```
ronronscelestes, line 50: Some comments on `SimpleMenu` v2:
- same padding fix as for SimpleMenu v1 which doesn't take the icon into account
```js
const OptionExternalLink = styled(Link)`
  padding: ${({ theme }) => theme.spaces[2]};
  ...
`;

...

<OptionExternalLink href={href} isExternal={isExternal} {...menuItemProps}>
  <Typography>{children}</Typography>
</OptionExternalLink>
```
- same fix as for `SimpleMenu` v1, recover focusable element
```js
  const menuItemProps = {
    tabIndex: isFocused ? 0 : -1,
    ref: menuItemRef,  // <=== we need the ref
    role: 'menuitem',
    ...props,
  };
```

ronronscelestes (changes requested):
hey @INNOVATIVEGAMER 
thank you for your updates, I tested your code for a while - I tried some fixes about styling, accessibility and other issues, I left comments for each components 🙂
for `v2` not relying on `react-router` which makes it difficult to know if we are in a state of internal or external, I'll answer in my previous comment for follow up!

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
ronronscelestes: hi - sorry for the delay
1. for `v1` I think we can rely only on `to` and `href` prop to know which link to render since `react-router` makes it pretty straightforward to know in which context we are (if `to` we want `NavLink`, if `href` we want `Link` and pass `href` prop for the right rendering)

2. `v2` still sounds a bit tricky, I think the `isLink` prop Hicham integrated to let developer use whatever routing they want is still something we want for this component. 
Only solution I have in mind right now to keep the flexibility of external `Link` + not count on `to` to make the difference between internal and external link would be to have a `isInternal` and `isExternal` prop but I'm not sure it's a great solution 🤔 
@HichamELBSI could you give your opinion on that topic because I think support of external link in `SimpleMenu` was also something that you wanted and you're the one who implemented the `SimpleMenu` `v2` API? 👀

INNOVATIVEGAMER:
> hi @INNOVATIVEGAMER I was wondering if you would like us to handle this refactor or if you still want to handle it? 😊
Sorry @ronronscelestes But due to some personal issues I won't be able to do this refactor now. I hope you understand. 
I would like you guys to handle it.

ronronscelestes:
> Sorry @ronronscelestes But due to some personal issues I won't be able to do this refactor now. I hope you understand. I would like you guys to handle it.

of course, I understand, no problem at all 😊
we might need to discuss the best solution for this refactor and open a new PR once we decide to handle it in a sprint - I will let you know!

ronronscelestes:
@joshuaellis reworked `v1/SimpleMenu` and `v2/SimpleMenu` to support external links without breaking changes
also updated documentation for both, let me know what you think 👀

joshuaellis (approved):
LGTM, i'd like @udimberto to review this too please :)

udimberto:
> @joshuaellis I reworked `v1/SimpleMenu` and `v2/SimpleMenu` to support external links without breaking changes also updated documentation for both, let me know what you think 👀

### `1.4.0` documentation

Hey, team @ronronscelestes @joshuaellis, I've checked the documentation, but didn't find any reference or example to use External Links in `MenuItem`:
- https://design-system-git-main-strapijs.vercel.app/?path=/docs/design-system-components-simplemenu--base
- https://design-system-git-main-strapijs.vercel.app/?path=/docs/design-system-components-v2-simplemenu--base

### Experiment to validate the current behavior of `MenuItem` in `v1` and `v2`

I did an experiment and find out that `MenuItem v2` works as external link with `href` and `isLink` combined:
- https://codesandbox.io/s/strapi-ds--udi--checking-pr714-mnftt5?file=/src/App.tsx

In the other hand, `MenuItem v2` does not work with internal routes.

Usage example for External and Internal links:
```tsx
import { MenuItem as MenuItemV1 } from '@strapi/design-system' 
import { SimpleMenu, MenuItem } from '@strapi/design-system/v2'
 
<SimpleMenu label="menu">
  <MenuItem href="https://strapi.io" target="_blank" isLink>
    Visit Strapi Website
  </MenuItem>
  <MenuItemV1 to="/internal-route">
    Navigate to an internal react-router-route
  </MenuItem>
</SimpleMenu>
```

### Is this PR still needed?
Since the current 1.4.0 `MenuItem v2` works with external links, I believe we don't need the technical changes anymore.
Maybe the documentation, yes.

I'm affraid about possible breaking changes with the new conditionals for `MenuItem v1` in this PR.

I believe the decision here it's complicated.
I'm not approving this PR because I can't be sure if we are going to have possible breaking changes.

ronronscelestes:
hi @udimberto, thank you for your review!

> I've checked the documentation, but didn't find any reference or example to use External Links in MenuItem
- `SimpleMenu` v1: https://design-system-git-fork-innovativegamer-support-4d7768-strapijs.vercel.app/?path=/docs/design-system-components-simplemenu--with-links#with-links
- `SimpleMenu` v2: https://design-system-git-fork-innovativegamer-support-4d7768-strapijs.vercel.app/?path=/docs/design-system-components-v2-simplemenu--with-links#with-links


> In the other hand, MenuItem v2 does not work with internal routes.

`SimpleMenu` v2 does not integrate `react-router-dom` - that was the all goal of having v2 components for links (e.g. removing opiniated routing system) 
there's a documentation to use `SimpleMenu` v2 with `react-router-dom` [there](https://design-system-git-fork-innovativegamer-support-4d7768-strapijs.vercel.app/?path=/docs/design-system-components-v2-simplemenu--base#with-other-routing-libraries), let me know if it's clearer for you!

> I'm afraid about possible breaking changes with the new conditionals for MenuItem v1 in this PR.

I'm not sure I see what would be the risks there, could you explain further please? 😊
`SimpleMenu` v1 should always use `MenuItem` v1 (same for v2), I don't think we should take into account mixing one version with the other as a solution, wdyt?

with the changes done here, `MenuItem` v1 is still supporting the `to` prop integrating `react-router-dom` component underneath 
only change is that if you use `href` instead of `to` you get a proper external link support (without having to add things like target blank yourself) 👀

> Is this PR still needed?

good question! I think it does for 2 reasons:
- `SimpleMenu` v1 doesn't support external link at all, and if we want to support it for v2 I think it's better to support it for v1
- `SimpleMenu` 2 supported external link in a sense since you could pass any prop to make it work but
  - then it did not give the UI asked by design in the case of external link (end icon)

I don't want to decide the relevancy by myself though, wdyt @joshuaellis @gu-stav?

joshuaellis:
I think this PR is still needed, take your example:

```ts
import { MenuItem as MenuItemV1 } from '@strapi/design-system' 
import { SimpleMenu, MenuItem } from '@strapi/design-system/v2'
 
<SimpleMenu label="menu">
  <MenuItem href="https://strapi.io/" target="_blank" isLink>
    Visit Strapi Website
  </MenuItem>
  <MenuItemV1 to="/internal-route">
    Navigate to an internal react-router-route
  </MenuItem>
</SimpleMenu>
```

Particularly on your usage of `<MenuItem href="https://strapi.io/" target="_blank" isLink>` if you fail to pass `isLink` this then assumes it's a button still passing the props – weird accessibility issue. If anything we should be checking that there is `href` and `isExternal` to really tighten down the usage of the component.

The same behaviour goes for `v1` components, whilst I see that you're talking about adding conditionals which _could_ be considered breaking changes @udimberto, realistically this would be a misuse of the components API (it's one reason i'm personally against prop spreading – different discussion / different day). So in reality we're clamping the usage of both component APIs.

Since it's not removing anything, I don't expect anything to break in either product because no one should have been applying `href / to` to a `button` element 😄

udimberto:
> hi @udimberto, thank you for your review!
> 
> > I've checked the documentation, but didn't find any reference or example to use External Links in MenuItem
> 
> * `SimpleMenu` v1: https://design-system-git-fork-innovativegamer-support-4d7768-strapijs.vercel.app/?path=/docs/design-system-components-simplemenu--with-links#with-links
> * `SimpleMenu` v2: https://design-system-git-fork-innovativegamer-support-4d7768-strapijs.vercel.app/?path=/docs/design-system-components-v2-simplemenu--with-links#with-links
> 
> > In the other hand, MenuItem v2 does not work with internal routes.
> 
> `SimpleMenu` v2 does not integrate `react-router-dom` - that was the all goal of having v2 components for links (e.g. removing opiniated routing system) there's a documentation to use `SimpleMenu` v2 with `react-router-dom` [there](https://design-system-git-fork-innovativegamer-support-4d7768-strapijs.vercel.app/?path=/docs/design-system-components-v2-simplemenu--base#with-other-routing-libraries), let me know if it's clearer for you!
> 
> > I'm afraid about possible breaking changes with the new conditionals for MenuItem v1 in this PR.
> 
> I'm not sure I see what would be the risks there, could you explain further please? 😊 `SimpleMenu` v1 should always use `MenuItem` v1 (same for v2), I don't think we should take into account mixing one version with the other as a solution, wdyt?
> 
> with the changes done here, `MenuItem` v1 is still supporting the `to` prop integrating `react-router-dom` component underneath only change is that if you use `href` instead of `to` you get a proper external link support (without having to add things like target blank yourself) 👀
> 
> > Is this PR still needed?
> 
> good question! I think it does for 2 reasons:
> 
> * `SimpleMenu` v1 doesn't support external link at all, and if we want to support it for v2 I think it's better to support it for v1
> * `SimpleMenu` 2 supported external link in a sense since you could pass any prop to make it work but
>   
>   * then it did not give the UI asked by design in the case of external link (end icon)
> 
> I don't want to decide the relevancy by myself though, wdyt @joshuaellis @gu-stav?

Thanks for the explanation, @ronronscelestes, very clear now.

udimberto:
> I think this PR is still needed, take your example @udimberto:
> 
> ```ts
> import { MenuItem as MenuItemV1 } from '@strapi/design-system' 
> import { SimpleMenu, MenuItem } from '@strapi/design-system/v2'
>  
> <SimpleMenu label="menu">
>   <MenuItem href="https://strapi.io/" target="_blank" isLink>
>     Visit Strapi Website
>   </MenuItem>
>   <MenuItemV1 to="/internal-route">
>     Navigate to an internal react-router-route
>   </MenuItem>
> </SimpleMenu>
> ```
> 
> Particularly on your usage of `<MenuItem href="https://strapi.io/" target="_blank" isLink>` if you fail to pass `isLink` this then assumes it's a button still passing the props – weird accessibility issue. If anything we should be checking that there is `href` and `isExternal` to really tighten down the usage of the component.
> 
> The same behaviour goes for `v1` components, whilst I see that you're talking about adding conditionals which _could_ be considered breaking changes @udimberto, realistically this would be a misuse of the components API (it's one reason i'm personally against prop spreading – different discussion / different day). So in reality we're clamping the usage of both component APIs.
> 
> Since it's not removing anything, I don't expect anything to break in either product because no one should have been applying `href / to` to a `button` element 😄

Thanks for your points, @joshuaellis.

Indeed, I agree with you and @ronronscelestes.

The snippet which I put in here, was just to show how it works today, I was not that "contextualized" yet, but with your both explanations, makes all the sense.

Let's move forward then.

udimberto (approved):
## This PR is still needed?



Oh 👏
Yep 👍

Good job and thank you, @INNOVATIVEGAMER

PR #716, opened 2022-10-07
Cleaning and/or solving the frontend tests (warning/logs)

simotae14:
What does it do?
It tries to remove all the warnings in the unit tests and e2e tests

Why is it needed?
to have proper logs when running the tests

How to test it?
you can run both the tests with these commands
Try to remove the warnings, prevent errors and warnings from the:

unit tests, you can run them with the command
`yarn test:watch`
E2E tests, you can run them with the command
`yarn test:e2e`
Related issue(s)/PR(s)
It is related to this ticket https://strapi-inc.atlassian.net/jira/software/c/projects/DX/boards/9?modal=detail&selectedIssue=DX-465

### DIFF
--- a/packages/strapi-design-system/src/DatePicker/DatePickerCalendar.js
+++ b/packages/strapi-design-system/src/DatePicker/DatePickerCalendar.js
@@ -71,7 +71,7 @@ export const DatePickerCalendar = ({
           </Box>
           <RawTable colCount={7} rowCount={weeks.length + 1} initialCol={activeCol} initialRow={activeRow} role="grid">
             <RawThead>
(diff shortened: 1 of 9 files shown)


joshuaellis:
In the case of elements like `RawTable`, whilst we're using `prop-type` they will fail because of this: https://github.com/facebook/react/issues/6653#issuecomment-215793856

They're not necessarily fixable imo, I don't think certain props should have a default because then the warnings aren't flagged for users 🤔 not sure your thoughts on this?

joshuaellis (approved):
amazing work @simotae14 ⭐

PR #717, opened 2022-10-11
fix: set z-index on Label to avoid z-index breaking out of parent

joshuaellis:
### What does it do?

Set `z-index` to `0` on the `Label` of ToggleCheckbox.

### Why is it needed?

it's absolute children break out of the parent currently and will therefore show on top of popovers

### Related issue(s)/PR(s)

CONTENT-572


ronronscelestes (approved):
Nice catch thank you for the fix 💃

PR #720, opened 2022-10-12
[fix][ToggleCheckbox]: responsive behaviour

joshuaellis:
### What does it do?

* Sets the padding to be 12px (as per Design Team instruction)
* Ensures the component obeys it's containers layout properties (e.g grid)

### Why is it needed?

On the content manager it crashes into other components if in a grid view at reasonable breakpoints (e.g. not extremes like 720px)

### Related issue(s)/PR(s)

CONTENT-571

### DIFF
--- a/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
+++ b/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
@@ -51,8 +51,8 @@ const ValueBox = styled(Flex).attrs({
     the values are in a row to be 40px. But defining a height on the label
     would break the input when it wraps.
   */
-  padding-top: 6px;
-  padding-bottom: 6px;
+  padding-top: ${({ size }) => `${size === 'S' ? '2px' : '6px'}`};
+  padding-bottom: ${({ size }) => `${size === 'S' ? '2px' : '6px'}`};
 `;
 
 /**
@@ -87,7 +87,6 @@ export const ToggleCheckbox = React.forwardRef(
 
         <ToggleCheckboxWrapper
           hasRadius
-          size={size}
           disabled={disabled}
           padding={1}
           display="flex"
@@ -97,6 +96,7 @@ export const ToggleCheckbox = React.forwardRef(
           borderColor="neutral200"
         >
           <ValueBox
+            size={size}
             paddingLeft={3}
             paddingRight={3}
             justifyContent="center"
@@ -116,6 +116,7 @@ export const ToggleCheckbox = React.forwardRef(
           </ValueBox>
 
           <ValueBox
+            size={size}
             paddingLeft={3}
             paddingRight={3}
             justifyContent="center"
--- a/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
+++ b/packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
@@ -12,6 +12,7 @@ const Label = styled.label`
   position: relative;
   display: inline-block;
   z-index: 0;
+  width: 100%;
 `;
 
 const ToggleCheckboxWrapper = styled(Box)`
@@ -44,6 +45,7 @@ const ValueBox = styled(Flex).attrs({
   position: relative;
   user-select: none;
   z-index: 2;
+  flex: 1 1 50%;
 `;
 
 /**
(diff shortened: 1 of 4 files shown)


ronronscelestes (commented):
thanks for working on it 💅
I placed the dist of this branch to Strapi DS node modules: 

not the same issue as before but now the input grid cell seems to take more space than before while losing width 🤔 or maybe it's the input shrinking inside its cell?

image

in DS website width:
image

in CM width:
image

also weirdly the input seems to shrink a lot in DS when making the view narrow:
image

joshuaellis:
> not the same issue as before but now the input grid cell seems to take more space than before while losing width 🤔 or maybe it's the input shrinking inside its cell?

It should be filling the grid space, I will check the setup in the admin app as it fills the cell in the DS grid I've included.

> also weirdly the input seems to shrink a lot in DS when making the view narrow:

So this is sort of the issue with it not being fully responsive, IMO it would be better if the input broke instead of breaking the view (which is what it currently does), but equally it's not ideal, i'd probably expect it to wrap in these circumstances which would mean the options stack. I'm not sure if this is "okay" to add though wdyt? one for design?

maevalienard:
>  i'd probably expect it to wrap in these circumstances which would mean the options stack. I'm not sure if this is "okay" to add though wdyt? one for design?

What do you have in mind when you say "to wrap"?

Apart from what Julie has already mentioned, looks good to me!

joshuaellis:
> What do you have in mind when you say "to wrap"?

@maevalienard I would expect the options to stack when theres not enough space for the options to be next to one another.

packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.stories.mdx
```diff
@@ -34,9 +34,18 @@ Use a toggle whenever the users need to turn something on or off in a form. A to
     {() => {
       const [checked, setChecked] = useState(false);
       return (
-        <ToggleCheckbox onLabel="True" offLabel="False" checked={checked} onChange={() => setChecked((prev) => !prev)}>
-          The field is required?
-        </ToggleCheckbox>
+        <div style={{ display: 'grid', gridTemplateColumns: 'repeat(12,1fr)', gridGap: '16px' }}>
```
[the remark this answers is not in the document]
joshuaellis: > You ~could~

I should

ronronscelestes:
> i'd probably expect it to wrap in these circumstances which would mean the options stack. 

I would expect that as well - I don't think it should be allowed to shrink because it gives the impression that the input is not usable (e.g. not seeing the true label) but as you said it also shouldn't break the UI as it did before your fix 🤔

maevalienard:
> @maevalienard I would expect the options to stack when theres not enough space for the options to be next to one another.

That would work to me. It's your time to shine @joshuaellis ✨

joshuaellis:
So i've updated the component, it breaks at 62px (because that is the minimum width of the labels), but for me it's an extreme case. If this is something we want to tackle then lmk a solution and I can implement.

yanniskadiri:
If that's an extreme case, then no need to do it @joshuaellis

packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
```diff
@@ -44,6 +45,14 @@ const ValueBox = styled(Flex).attrs({
...
   user-select: none;
   z-index: 2;
+  flex: 1 1 50%;
+  /**
+    We declare the defined value because we want the height of the input when 
+    the values are in a row to be 40px. But defining a height on the label
+    would break the input when it wraps.
+  */
+  padding-top: ${({ size }) => `${size === 'S' ? '2px' : '6px'}`};
+  padding-bottom: ${({ size }) => `${size === 'S' ? '2px' : '6px'}`};
```
gu-stav, line 55: I think @ronronscelestes will comment that these should be defined as props on `ToggleCheckboxWrapper`, because `Box` provides `paddingTop` and `paddingBottom` already.

gu-stav (approved):
Added a small stylistic comment. Overall it works well and is a very good idea IMO!

packages/strapi-design-system/src/ToggleCheckbox/ToggleCheckbox.js
ronronscelestes: agree, I'm also worried about `6px` since it's not a theme spacing value but I guess we don't really have a choice here 🤔

gu-stav: > I'm also worried about 6px

What about adding a code-comment to reason about these magic numbers?

joshuaellis: you mean like this:

```js
  /**
    We declare the defined value because we want the height of the input when 
    the values are in a row to be 40px. But defining a height on the label
    would break the input when it wraps.
  */
```

Which already is there, but maybe it's not clear enough?

gu-stav: Ah, sorry - forgot about it, because the comment only refers to two lines. The comment works for me.

PR #729, opened 2022-10-25
Fix theme colors stories

HichamELBSI:
### What does it do?

New way to display theme color palettes.

### Why is it needed?

In order to show the `warningXXX` colors (this color is currently hidden), I had to add a white box to avoid color contrast warning.

Screenshot 2022-10-25 at 11 59 22

Screenshots

#### Before
Screenshot 2022-10-25 at 12 51 41

#### After
Screenshot 2022-10-25 at 12 51 30


gu-stav (approved):
Works for me. Very good idea - thank you!

PR #732, opened 2022-10-28
chore: Replace babel-jest with @swc/jest

gu-stav:
### What does it do?

Replaces `babel-jest` with `@swc/jest`.

### Why is it needed?

1. It follows the tooling in the strapi core and cloud repositories
2. It is much faster (1:50min -> ~30sec)


packages/strapi-design-system/test-bundler.js
```diff
@@ -1,42 +1,35 @@
 import IntlPolyfill from 'intl';
+import ResizeObserver from 'resize-observer-polyfill';
```
joshuaellis, line 2: How come this one and not https://github.com/juggle/resize-observer ? I think the latter is actually maintained regularly.

joshuaellis (changes requested):

packages/strapi-design-system/test-bundler.js
gu-stav: I don't mind which one to choose. If you think this one works better, I'm ok with that 👍🏼

joshuaellis (approved):

packages/strapi-design-system/test-bundler.js
joshuaellis: I just looked at when they were last updated 😅

PR #734, opened 2022-10-31
Status: Make preceding bullet configurable

gu-stav:
### What does it do?

It makes the preceding bullet of the [`Status`](https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-status--base) component configurable through a prop called `showBullet`.

### Why is it needed?

The bullet is not part of any `Status` component any longer. To stay visually consistent, I've added a new prop, which allows disabling the bullet and renders only `children` instead.

I think we could approach this differently if we'd want:

1. We could argue the bullet is decoration and, therefore, we can change it right away. This way, we'd have to keep the child `Flex` around.
2. We could create a new v2 version of the component which does not know about `Bullet` or `Flex`, but I feel like it adds too much unnecessary overhead.

In the spirit of composability, I think getting rid of any child wrappers is a sensible default for the future because it is less opinionated.

### How to test it?

The automated tests do that for you.

### DIFF
--- a/packages/strapi-design-system/src/Status/Status.js
+++ b/packages/strapi-design-system/src/Status/Status.js
@@ -19,7 +19,7 @@ const StatusWrapper = styled(Box)`
   }
 `;
 
-export const Status = ({ variant, children, ...props }) => {
+export const Status = ({ variant, showBullet, children, ...props }) => {
   const backgroundColor = `${variant}100`;
   const borderColor = `${variant}200`;
   const bulletColor = `${variant}600`;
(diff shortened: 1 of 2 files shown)


joshuaellis:
> We could create a new v2 version of the component which does not know about Bullet or Flex, but I feel like it adds too much unnecessary overhead.

I agree, this is unnecessary when we can provide a prop.

> In the spirit of composability, I think getting rid of any child wrappers is a sensible default for the future because it is less opinionated.

I also agree, I would expect this status component to style it's children not add decorations. It limits it's scope too much.

joshuaellis (approved):
Good solution for moving forward 👍🏼

PR #726, opened 2022-10-19
Icon: Update Strapi logo icon

gu-stav:
### What does it do?

Update the Strapi logo to the most recent version.

### Why is it needed?

To stay up-to-date with the company branding.

### Related issue(s)/PR(s)

- Fixes https://github.com/strapi/design-system/issues/725

PR #735, opened 2022-10-31
Theme: Remove default color from icons

gu-stav:
### What does it do?

It removes the default color assignment in the `Icon` documentation in

### Why is it needed?

> **Note**
> This is meant to be a discussion starter rather than a final implementation. Maybe I don't see the full picture, but I don't see how to not pass in a default color.

I found it quite confusing when working on https://github.com/strapi/design-system/pull/726 that my icon doesn't show up in the original color. I think not recoloring these icons will help us to spot problems easier in the future.

Looking it the current icons, I don't think it is necessary for many of them to be colored by default:

| Before | After |
|-|-|
| Screen Shot 2022-10-31 at 12 07 24 | Screen Shot 2022-10-31 at 11 11 53 |


udimberto (approved):
### Looks good to me

joshuaellis:
Closing due to inactivity – we can revisit this once we're ready to address it

PR #736, opened 2022-10-31
Add Rotate icon

gu-stav:
### What does it do?

Adds a new icon called `Rotate` (naming was stolen from font-awesome).

### Why is it needed?

This icon will be used in strapi in at least two places soon:

#### Status

`Pencil` and and `Check` are already part of the DS package.

Screenshot 2022-10-31 at 12 22 40

#### Locale switcher

We haven't finally decided how we could implement these icons. @maevalienard mentioned we can add them to the DS, but I'm a bit hesitant because they are basically `Pencil`, `Check` and `Rotate` with a background and border. I'd like to try first if we could replicate these using CSS in strapi.

Screenshot 2022-10-31 at 12 23 33

### How to test it?

tbd


joshuaellis (approved):

maevalienard:
LGTM (even if already merged 😅)

PR #738, opened 2022-10-31
GitHub: Use ubuntu as GitHub Action base image instead of macos

gu-stav:
### What does it do?

Changes the GitHub Actions base image from macOS to ubuntu.

### Why is it needed?

Running macOS is relatively slow (~ 40min down to ...) and expensive (minutes are multiplied by 10) compared to the ubuntu base image.

I don't see a specific reason we should use macOS to run our tests. This might change in case we enable visual regression tests one day.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   lint:
     name: "Lint"
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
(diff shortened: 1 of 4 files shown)


joshuaellis:
Reviewing this, maybe we need to await the update. I mean it would be a state update in jest so you would want to wait then to? 🤔

joshuaellis:
Okay so now we _know_ the intersection observer is not firing.

.github/workflows/tests.yml
```diff
@@ -5,24 +5,24 @@ on:
...
+  #         node-version: "14"
+  #         cache: "yarn"
+  #     - uses: ./.github/actions/install-modules
+  #     - name: Link locale packages and install their dependencies
+  #       run: yarn bootstrap
+  #     - name: Run lint
+  #       run: yarn lint
   unit_tests:
     name: "unit tests"
     needs: [lint]
```
joshuaellis, line 24: you need to omit this

joshuaellis:
closing due to inactivity

PR #740, opened 2022-11-01
feat: add vite to bundle

joshuaellis:
### What does it do?

This PR does a couple things:

* Adds vite as the bundler for the packages
* Consolidates deps / devDeps e.g. there were some devDeps in deps & some devDeps we weren't using in certain areas or at all

Ideally, I'd like to lift out storybook into it's own `docs` folder and make that isolated from the packages so we have to `import { Select } from '@strapi/design-system'` so that 

1. we can be sure the public facing package works as intended
2. remove unnecessary devDeps from the `design-system` package and keep it tidier in the long run.

This would require moving all the stories around and I want to get the opinion on it before actioning.

### Why is it needed?

Because moving to `vite` decreases bundle size by ~70-80% and therefore allows better tree shaking in other applications.
Keeping the packages tidy as possible to ensure we're not adding additional overheads where not required i.e. when people install our package imo they shouldn't be installing all the unneccessary devDeps for our storybook etc.

### Notes

~Because of the changes to `type:module` in the package.json, #741 needs to be merged first.~

It also might be worth releasing a `beta` of this prior to stable to ensure it correctly works with `cloud` and `cms`.
Therefore, we could set up a `beta` branch – thoughts?


joshuaellis:
Because of the changes to `type:module` in the package.json, #741 needs to be merged first.

joshuaellis:
NOTE: it also might be worth releasing a `beta` of this prior to stable to ensure it correctly works with `cloud` and `cms`.

Therefore, we could set up a `beta` branch – thoughts?

udimberto (approved):

packages/strapi-design-system/vite.config.js
```diff
@@ -0,0 +1,36 @@
+import { resolve } from 'path';
+import { defineConfig } from 'vite';
+import react from '@vitejs/plugin-react';
+import glob from 'tiny-glob';
+
+export default glob('./src/**/!(*.spec|*.e2e).{js,svg}').then(async (paths) => {
+  console.log(paths, resolve(__dirname, './src/index.js'));
```
gu-stav, line 7: Might be a leftover.

gu-stav (commented):
Overall this looks really tidy 💅🏼 

I found only one thing that I'm not sure of: `dist/index.cjs` contains only `require` statements and not the whole bundle. Is that intentional? Would e.g. `webpack` resolve these dependencies even if we do not transpile `node_modules` and consume cjs files? I think in the ESM case it does, but I'm not sure about cjs.

packages/strapi-design-system/vite.config.js
joshuaellis: > I found only one thing that I'm not sure of: dist/index.cjs contains only require statements and not the whole bundle. Is that intentional? Would e.g. webpack resolve these dependencies even if we do not transpile node_modules and consume cjs files? I think in the ESM case it does, but I'm not sure about cjs.

Yes, I've re-linked it with the `admin` app and it's does still build. The `exports` part of the `package.json` helps bundlers know which file is for `require` and which for `import` from there it's just a matter of following the trail.

Realistically it would be simpler to just bundle everything into one file because the direct import is unnecessary now (we could action removing `@strapi/design-system/Select` in the admin app when we release this version if you agree) because components aren't duplicated, but if we did that here then it would break plugins 😄

packages/strapi-design-system/package.json
```diff
@@ -57,15 +59,16 @@
...
+    "dev": "vite",
     "prebuild": "rimraf dist",
-    "build": "yarn build:files && cross-env IS_V2=true yarn build:files",
-    "build:files": "yarn build:dev && yarn build:prod && yarn copy:files && yarn create:index-files",
-    "build:dev": "cross-env NODE_ENV=development webpack-cli",
-    "build:prod": "cross-env NODE_ENV=production webpack-cli",
-    "copy:files": "node ../../tools/copy-files.js",
-    "create:index-files": "node ./createIndexFiles.js",
+    "build": "yarn build:prod",
+    "build:prod": "vite build",
```
gu-stav, line 71: Does it make a difference to set `NODE_ENV=production` here? Just to make sure the person who releases the DS publishes the most efficient bundles ...

joshuaellis: I don't believe so, it by default runs build in `production` mode – https://vitejs.dev/guide/env-and-mode.html#modes

packages/strapi-design-system/vite.config.js
gu-stav: > The exports part of the package.json helps bundlers know which file is for require and which for import from there it's just a matter of following the trail.

💡 Thanks! I knew it did in the case of esm (`module`), but it's good to know it does also for cjs (`main`).

> Realistically it would be simpler to just bundle everything into one file because the direct import is unnecessary now (we could action removing @strapi/design-system/Select in the admin app when we release this version if you agree) because components aren't duplicated but if we did that here then it would break plugins 😄

Yes, I agree and think we should settle for one direction in the future. It would also simplify the imports in e.g. strapi core a lot, if we could use one import statement.

joshuaellis: Cool, I am happy to action this when I update the admin app with the new DS 👍🏼

gu-stav (approved):
I'm good with these changes. Thanks again @joshuaellis for pushing it forward 🚀 

I think in terms of merge/ release I'd prefer if we'd release `1.26` asap followed by a `1.3` alpha/ beta release, so that both (cloud and cms) can pull it in to see if that works well.

What do you think?

joshuaellis:
> I think in terms of merge/ release I'd prefer if we'd release 1.26 asap followed by a 1.3 alpha/ beta release, so that both (cloud and cms) can pull it in to see if that works well.

> What do you think?

I agree, I'll try to get 1.2.6 released today as there's bits for `relations` that we need, then when people are happy with this PR we should do a `beta` release of 1.3 👍🏼 check it works across cloud and admin then release stable.

packages/strapi-design-system/.eslintrc
```diff
@@ -0,0 +1,34 @@
...
+    {
+      "files": "*.mdx",
+      "parser": "eslint-mdx",
+      "rules": {
+        "react/react-in-jsx-scope": "off",
+        "no-console": 0
+      }
+    }
+  ]
+}
```
HichamELBSI, line 34: It would be nice if we can extend the [@strapi/esling-config](https://github.com/strapi/eslint-config/blob/main/front.js). I think it would be better to have a more consistent front-end codebase between all the Strapi projects.

HichamELBSI: No need to do that move in your right now PR btw.

HichamELBSI (approved):
Good job @joshuaellis !

packages/strapi-design-system/.eslintrc
gu-stav: Agree - that's a nice task for another PR!

joshuaellis: Yeah i looked into that previously. But agreed, it would be easier to do that.

PR #741, opened 2022-11-01
chore: update playwright

joshuaellis:
### What does it do?

* Updates playwright (and removes unnecessary files)

### Why is it needed?

* Because it can be used across packages now and also does not require us to use the CJS form of imports.

### How to test it?

E2E tests will validate


package.json
```diff
@@ -17,17 +17,21 @@
...
     "storybook": "lerna run storybook --stream",
-    "test:e2e": "lerna run test:e2e --stream",
-    "test:e2e:ci": "lerna run test:e2e:ci",
-    "test:e2e:watch": "lerna --scope @strapi/design-system exec -- yarn test:e2e:watch",
     "test": "lerna run test --stream",
     "test:watch": "lerna --scope @strapi/design-system exec -- yarn test:watch",
+    "test:e2e": "playwright test",
+    "test:e2e:ci": "cross-env CI=true playwright test",
+    "test:e2e:watch": "chokidar '**/__tests__/*.e2e.js' -c 'npm run test:e2e -- --retries=0 {path}'",
+    "test:e2e:debug": "cross-env PWDEBUG=1 playwright test",
```
gu-stav, line 25: Why do you think it is better to have these at the repo root instead of only for the DS?

.github/workflows/tests.yml
```diff
@@ -57,7 +57,7 @@ jobs:
           globalPackages: http-server
 
       - name: Install Playwright
-        run: npx playwright@1.20.0 install --with-deps
+        run: npx playwright@1.27.1 install --with-deps
```
gu-stav, line 60: I think I've added something about this in the README or CONTRIBUTING file. Could you check if these need an update too?

package.json
joshuaellis: IMHO our packages should be clean as possible, when someone downloads the package it will also download the devDeps (which incl. playwright), this also opens the opportunity to share a config if we introduce more packages e.g. we do a separate v2 package or split the components up etc.

.github/workflows/tests.yml
joshuaellis: Updated 👍🏼

package.json
gu-stav: It makes sense, but in both cases (playwright at the root or scoped), developers must download dev dependencies if they want to interact with the repository. I think in the spirit of a monorepo, we should scope as much as possible and only move dependencies a level up once we run into a case where e.g. a configuration needs to be shared.

As long as we only update dependencies, we can move forward with this PR - otherwise, I think we need more opinions from e.g. Alex / the DS sync meeting.

joshuaellis: > developers must download dev dependencies if they want to interact with the repository

To be clear, if they download the package from npm they wouldn't need to download unnecessary devDeps in this scenario.

Okay for now i'll limit the scopes of the PRs to the specific changes and we can discuss shuffling things around in a DS sync 😄

gu-stav (approved):
Nice! Well done! 🚀

PR #746, opened 2022-11-02
[enhancement]: IconButton

joshuaellis:
### What does it do?

* Allows passing children as opposed to icon (although PropTypes aims to enforce we should pass one or the other)
* Ensure if you don't pass a label you pass an `aria-label` cause there's no excuse for no accessibility
* Set no-undef in `eslint` to `error` cause it's a better DX imo
* Adds extra accessibility flags to the `icon`/`children` via `cloneElement`
* Use's `VisuallyHidden` spans because iirc translators are more likely to translate those than aria-labels and it's clearer from a code POV

### Why is it needed?

* Because I'm adding drag and drop to relation items and the drag icon _needs_ to be accessible.

### DIFF
--- a/packages/strapi-design-system/src/IconButton/IconButton.js
+++ b/packages/strapi-design-system/src/IconButton/IconButton.js
@@ -144,18 +144,32 @@ IconButton.defaultProps = {
 /**
  * @type {(otherProps: string[]) => (props: Record<string, unknown>, propName: string) => Error | undefined}
  */
-const throwPropErrorIfNoneAreDefined = (otherProps) => (props, propName) => {
+const throwPropErrorIfNoneAreDefined = (otherProps, propType) => (props, propName) => {
   if (!props[propName] && otherProps.every((otherProp) => !props[otherProp])) {
     return new Error(`One of the following props is required: ${propName}, ${otherProps.join(', ')}`);
   }
+
+  PropTypes.checkPropTypes({ [propName]: PropTypes[propType] }, props, 'prop', 'IconButton');
 };
 
 IconButton.propTypes = {
-  ['aria-label']: throwPropErrorIfNoneAreDefined(['label']),
-  children: throwPropErrorIfNoneAreDefined(['icon']),
+  /**
+   * `PropTypes.string` – must be defined if `label` is not defined.
+   */
+  ['aria-label']: throwPropErrorIfNoneAreDefined(['label'], 'string'),
+  /**
+   * `PropTypes.node` – must be defined if `icon` is not defined.
+   */
+  children: throwPropErrorIfNoneAreDefined(['icon'], 'node'),
   disabled: PropTypes.bool,
-  icon: throwPropErrorIfNoneAreDefined(['children']),
-  label: throwPropErrorIfNoneAreDefined(['aria-label']),
+  /**
+   * `PropTypes.node` – must be defined if `children` is not defined.
+   */
+  icon: throwPropErrorIfNoneAreDefined(['children'], 'node'),
+  /**
+   * `PropTypes.string` – must be defined if `aria-label` is not defined.
+   */
+  label: throwPropErrorIfNoneAreDefined(['aria-label'], 'string'),
   noBorder: PropTypes.bool,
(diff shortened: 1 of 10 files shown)


packages/strapi-design-system/src/Switch/Switch.stories.mdx
```diff
@@ -36,7 +36,7 @@ Activated state is one of two of them available for a Switch.
 
 <Canvas>
   <Story name="activated">
-    <Switch label="Activate the microphone" selected={true} onChange={() => setChecked((s) => !s)} visibleLabels />
+    <Switch label="Activate the microphone" selected={true} onChange={(e) => console.log(e)} visibleLabels />
```
gu-stav, line 39: I guess these are a mistake?

packages/strapi-design-system/src/IconButton/IconButton.stories.mdx
```diff
@@ -95,6 +95,31 @@ IconButtons can be used within another component is a Group shape.
...
 
+### Icons as Children & aria-label
+
+IconButtons can take Icons as their children and can be fully accessible without a tooltip.
+
+<Canvas>
+  <Story name="children">
+    <Box padding={7}>
+      <Flex>
+        <IconButton onClick={() => console.log('edit')} aria-label="Edit">
```
gu-stav, line 106: What do you think about updating some visible text instead? It is easier to understand for readers of the docs, I think.

packages/strapi-design-system/src/IconButton/IconButton.js
```diff
@@ -126,10 +140,22 @@ IconButton.defaultProps = {
...
+};
+
 IconButton.propTypes = {
+  ['aria-label']: throwPropErrorIfNoneAreDefined(['label']),
+  children: throwPropErrorIfNoneAreDefined(['icon']),
   disabled: PropTypes.bool,
-  icon: PropTypes.element.isRequired,
-  label: PropTypes.string,
+  icon: throwPropErrorIfNoneAreDefined(['children']),
+  label: throwPropErrorIfNoneAreDefined(['aria-label']),
```
gu-stav, line 158: These helpers now show up in the user-facing documentation. Could you overwrite these with some better documentation (using code-comments)?

Screenshot 2022-11-02 at 15 07 08

packages/strapi-design-system/src/Switch/Switch.stories.mdx
joshuaellis: Well, they're technically `undefined` so the state didn't exist in the first place for the story. Could add it correctly if you prefer?

packages/strapi-design-system/src/IconButton/IconButton.js
joshuaellis: I didn't know that was possible, do you have an example / doc, but yes, I agree this would be better off resolved.

gu-stav: `BoxProps.js` contains examples. I didn't know either but storybook seems to parse code comments and use them to overwrite the defaults.

joshuaellis: nice, thats cool.

packages/strapi-design-system/src/IconButton/IconButton.stories.mdx
joshuaellis: I'm not sure i understand what you're proposing here.

gu-stav: I wanted to say that it could be nice to give visual feedback to somebody reading this documentation. Instead of logging these values, you could add them to a local store and render the value, to make it clear something is happening.

packages/strapi-design-system/src/Switch/Switch.stories.mdx
gu-stav: Yeah, I think I'd prefer if the stories would work correctly.

gu-stav (approved):
I'm good with that. Very good idea!

Issue #752, opened 2022-11-04
Combobox inside a ModalLayout results in a frustrated user


M15sy:
### Describe the bug

TLDR;  When selecting a ComboboxOption within a ModalLayout, the DismissibleLayer pointerdown event is triggered because the ComboboxOption is within a different portal to the modal’s portal so `!layerRef.current.contains(event.target) ` evaluates to true. This results in the modal closing, the Combobox onChange handler never being called and the user feeling frustrated as they believe the app to be broken.

### Additional context

Hi Strapi team 👋,

I recently upgraded a Strapi app to v4.4.5 and tried to create a new component type from the Content-Type Builder UI. However, when I tried to select the new component’s category, the ‘Create a component’ modal closed and I was not able to create my new component (see attached screenshot below).

After a bit of debugging, I figured out what the cause of this behaviour was (see bug description above) and I now know if I use my keyboard to select the category I am able to continue and create my new component type. 

I imagine the intention of the app is not to cause frustration so it would be great if this could be looked at. I did give it a little thought... and one possible solution could be to remove the pointerdown event from the DismissibleLayer and instead add an onClick handler to the ModalWrapper, but I’ll leave the choice up to you guys - though it would be great to validate the fix also works when selecting a ComboboxOption that overhangs the modal.

Thanks ✌️

### Required System information

- Npm version: 8.11.0
- NodeJS version: 16.16.0
- Strapi Design System version: 1.2.6
- Browser: Chrome 106.0.5249.119

### Steps to reproduce the behavior

1. Find a ModalLayout with a Combobox inside or use the code sandbox linked below.
2. Try to select a ComboboxOption using your mouse.
3. The modal closes and the Combobox value is never updated.

### Expected behavior

Combobox onChange to be triggered and the modal to stay open.

### Code snippets

https://codesandbox.io/embed/boring-chatterjee-gqhd7k?fontsize=14&hidenavigation=1&theme=dark

### Screenshots

<img width="1792" alt="image" src="https://user-images.githubusercontent.com/12456414/199942036-5ed760a3-47f7-40dc-b974-3d923683d78e.png">

joshuaellis:
Thanks for the report! I'm going to look at this now when I have an update I'll post here.

dennislo:
Hi @joshuaellis 

I'm also seeing this issue for the `Add new component to the dynamic zone` modal
[...]

PR #753, opened 2022-11-04
fix: dismissible layer closing prematurely

joshuaellis:
### What does it do?

* checks the closest react-portal to avoid other portalled elements closing the layer

### Why is it needed?

* because it breaks the CMS

### Related issue(s)/PR(s)

* #752


gu-stav (approved):
I've seen it working and like the solution. Thanks for fixing this so quickly!

HichamELBSI (approved):

PR #755, opened 2022-11-04
BaseCheckbox: Wrap component into Box to isolate style resets

gu-stav:
### What does it do?

Wraps the component into a `div` so that styles don't interfere with parent styles.

### Why is it needed?

`CheckboxInput` overwrites `margin`. In case of the settings pages for example, this creates a problem because it would overwrite spacing defined by `Stack`:

Screenshot 2022-11-04 at 14 43 21

With this fix the `margin` reset is properly isolated:

Screenshot 2022-11-04 at 14 45 04

### Related issue(s)/PR(s)

- Refs https://github.com/strapi/strapi/issues/14763


udimberto (approved):
### LGTM

joshuaellis (approved):

PR #756, opened 2022-11-07
Add: SimpleNav and SimpleNavItem components

udimberto:
### Feature
Adds two new components for simple navigation:
- `SimpleNav`, derivated from `Flex` and `Box`
- `SimpleNavItem`, derivated from `Box` as `BaseLink`

### Chore
Few modifications on `Box` component styling props:
- Accepts one more possible-prop `gap` 
- Re-group the `fontSize` prop with the Text properties
- Removes `cursor` duplicate from styles

## Why is it needed?
Will help on Strapi Cloud Dashboard.

### Considerations
Strapi Design System already have similar components, like `MainNav` and it's scope components.
But the existents ones of `MainNav` are little different from the designed to Strapi Cloud Dashboard.
Those differences maybe could be worked to `MainNav`, but I assume that is risky considering the Strapi CMS Dashboard.
Then I decided to add this one here in behalf of not `MainNav` not having an "error" state, which it's necessary on Strapi Cloud.

## Related issue(s)/PR(s)
Strapi Cloud it's a private repository, but will be used on Project Settings page as sidebar navigation and that will be available for our community and customers soon.

## Preview
| ~ | ~ |
| --- | --- |
| Screenshot 2022-11-06 at 10 06 54 PM | Screenshot 2022-11-06 at 10 07 17 PM |

### DIFF
--- a/packages/strapi-design-system/src/SimpleNav/SimpleNav.js
+++ b/packages/strapi-design-system/src/SimpleNav/SimpleNav.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import { Flex } from '../Flex';
+
+export const SimpleNav = (props) => {
+  return <Flex {...props} />;
+};
+
+SimpleNav.defaultProps = {
+  ...Flex.defaultProps,
+  as: 'nav',
+  display: 'flex',
+  direction: 'column',
+  alignItems: 'stretch',
+  position: 'sticky',
+  top: 4,
+  gap: 2,
+};
+
+SimpleNav.propTypes = {
+  ...Flex.propTypes,
+};
--- a/packages/strapi-design-system/src/SimpleNav/SimpleNavItem.js
+++ b/packages/strapi-design-system/src/SimpleNav/SimpleNavItem.js
@@ -1,0 +1,94 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { BaseLink } from '../BaseLink';
+import { Box } from '../Box';
+import { Flex } from '../Flex';
+import { Typography } from '../Typography';
+import { buttonFocusStyle } from '../themes/utils';
+
+const SimpleNavItemWrapper = styled(Box)`
+  text-decoration: none;
+  background: ${({ background, theme }) => theme.colors[background] || background};
+  color: ${({ textColor, theme, withError }) => theme.colors[withError ? 'danger600' : textColor]};
+
+  &.active,
+  &:active,
+  &[aria-current='page'] {
+    background: ${({ theme, withError }) => theme.colors[withError ? 'danger100' : 'primary100']};
+    color: ${({ theme, withError }) => theme.colors[withError ? 'danger600' : 'primary700']};
+  }
+
+  &:not(.active):not(:active):not([aria-current='page']) {
+    &:hover {
+      background: ${({ theme, withError }) => theme.colors[withError ? 'danger100' : 'neutral150']};
+      color: ${({ theme, withError }) => theme.colors[withError ? 'danger600' : 'neutral800']};
+    }
+  }
+
+  svg,
+  svg * {
+    fill: currentColor;
+  }
+
+  ${buttonFocusStyle}
+`;
+
+const SimpleNavItemIconWrapper = (props) => <Flex as="span" alignItems="center" aria-hidden="true" {...props} />;
+
+export const SimpleNavItem = ({
+  children,
+  className,
+  endIcon,
+  gap,
+  isActive,
+  startIcon,
+  textVariant,
+  withError,
+  ...rest
+}) => {
+  return (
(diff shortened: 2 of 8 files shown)


packages/strapi-design-system/src/SimpleNav/SimpleNav.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Flex } from '../Flex';
+
+export const SimpleNav = (props) => {
+  return <Flex {...props} />;
```
joshuaellis, line 5: Semantically I believe this should be a `nav` item then wrapping `ul` element because its a nav list.

packages/strapi-design-system/src/SimpleNav/SimpleNav.js
```diff
@@ -0,0 +1,21 @@
...
+};
+
+SimpleNav.defaultProps = {
+  ...Flex.defaultProps,
+  as: 'nav',
+  display: 'flex',
```
joshuaellis, line 11: IMO we shouldn't be giving users the options to change these styles _easily_ of course, they can wrap in their own `styled` components, but in the interest of a small API surface I think we should just keep it simple. 

I don't think we should spread all the props in the component so it could be as simple as:

```jsx
export const SimpleNav = (props) => {
	return <Flex>{props.children}</Flex>
}
```

Unless we explicitly _need_ users to add more props in which case they should be added.

packages/strapi-design-system/src/SimpleNav/SimpleNavItem.js
```diff
@@ -0,0 +1,94 @@
...
+  startIcon,
+  textVariant,
+  withError,
+  ...rest
+}) => {
+  return (
+    <SimpleNavItemWrapper className={`${className}${!isActive ? '' : ' active'}`} withError={withError} {...rest}>
+      <Typography variant={textVariant} textColor="currentColor">
+        <Flex as="span" display="flex" alignItems="center" justifyContent="space-around" gap={gap} width="100%">
+          {startIcon && <SimpleNavItemIconWrapper>{startIcon}</SimpleNavItemIconWrapper>}
```
joshuaellis, line 54: IMHO I really don't like the pattern of `startIcon` and `endIcon` and whilst I do appreciate we use it elsewhere in the design system I'm not sure it makes that much sense, you're then assuming that all Icons need to behave the same which we've found that they don't i.e. the sizing / spacing is different per use case according to the design.

I would just leave it as `children` and let the user do it – Composability.

packages/strapi-design-system/src/SimpleNav/SimpleNavItem.js
```diff
@@ -0,0 +1,94 @@
...
+  endIcon,
+  gap,
+  isActive,
+  startIcon,
+  textVariant,
+  withError,
+  ...rest
+}) => {
+  return (
+    <SimpleNavItemWrapper className={`${className}${!isActive ? '' : ' active'}`} withError={withError} {...rest}>
```
joshuaellis, line 51: I would just pass `isActive` as a prop because this string concatenation can get messy without a lib like `clsx`.

packages/strapi-design-system/src/SimpleNav/SimpleNavItem.js
```diff
@@ -0,0 +1,94 @@
...
+  hasRadius: true,
+  isActive: false,
+  kind: undefined,
+  paddingTop: 2,
+  paddingRight: 4,
+  paddingBottom: 2,
+  paddingLeft: 4,
+  startIcon: undefined,
+  textColor: 'neutral600',
+  textVariant: 'omega',
```
joshuaellis, line 82: Same rules apply to this styling props, I think we style the component and leave it at that, we're giving _too_ much flexibility and I think it could therefore be easily abused.

packages/strapi-design-system/src/SimpleNav/SimpleNavItem.js
```diff
@@ -0,0 +1,94 @@
...
+import { buttonFocusStyle } from '../themes/utils';
+
+const SimpleNavItemWrapper = styled(Box)`
+  text-decoration: none;
+  background: ${({ background, theme }) => theme.colors[background] || background};
+  color: ${({ textColor, theme, withError }) => theme.colors[withError ? 'danger600' : textColor]};
+
+  &.active,
+  &:active,
+  &[aria-current='page'] {
```
joshuaellis, line 17: where does `aria-current` come from in this instance?

joshuaellis (changes requested):
Thanks for the contribution Udi! and congratulations on making your first to the DS 🥳 

Don't worry about the size impact, this is largely down to the bundler at the moment but _should_ be resolved when we launch the `vite` version of the lib!

udimberto:
😅
I made a mistake, trying to put those components to DS, when actually it will be present only into Strapi Cloud particulary.

🫶
Thank you for all the constructive comments, @joshuaellis, I'm applying then to it on Strapi Cloud.

PR #758, opened 2022-11-08
Button: Unify component height with the UI Kit

gu-stav:
### What does it do?

Updates the buttons to use the heights defined in the [UI Kit](https://www.figma.com/file/PICiE8O4NLrHO1lhJftLdE/Strapi---UI-Kit-%F0%9F%A7%A9?node-id=9974%3A117504):

- `S`: 32px
- `M`: 36px
- `L`: 40px

### Why is it needed?

1. Currently the different sizes look messed up:

| Before | After |
|-|-|
| Screenshot 2022-11-08 at 17 05 29 | Screenshot 2022-11-08 at 17 20 57 |

➡️ [Preview](https://design-system-git-fix-button-height-strapijs.vercel.app/?path=/story/design-system-components-button--sizes)

I did my best to compose the height out of `line-height` and `padding` but neither combination worked and we always ended up using sub-pixel values. Therefore I've decided to make the heights fixed, which follows what has been done in `IconButton`.

2. The height of the `IconButton` doesn't align with the `Button[size=S]`: we have to shown `IconButton` in a row with `Buttons` and therefore they should have the same height.

### How to test it?

 Preview: https://design-system-git-fix-button-height-strapijs.vercel.app/?path=/story/design-system-components-button--sizes

### DIFF
--- a/packages/strapi-design-system/src/Button/Button.js
+++ b/packages/strapi-design-system/src/Button/Button.js
@@ -27,19 +27,33 @@ const BoxFullHeight = styled(Box)`
   height: 100%;
 `;
 
-// TODO: Check the L size button with Maeva
 export const ButtonWrapper = styled(BaseButton)`
   align-items: center;
-  padding: ${({ theme, size }) => `${size === 'S' ? theme.spaces[2] : '10px'} ${theme.spaces[4]}`};
-  background: ${({ theme }) => theme.colors.buttonPrimary600};
+  background-color: ${({ theme }) => theme.colors.buttonPrimary600};
   border: 1px solid ${({ theme }) => theme.colors.buttonPrimary600};
+  height: ${({ size }) => {
+    let height = 32;
+
+    if (size === 'M') {
+      height = 36;
+    } else if (size === 'L') {
+      height = 40;
+    }
+
+    return `${height / 16}rem`;
+  }};
+  padding-left: ${({ theme }) => theme.spaces[4]};
+  padding-right: ${({ theme }) => theme.spaces[4]};
+
   ${Box} {
     display: flex;
     align-items: center;
   }
+
   ${Typography} {
     color: ${({ theme }) => theme.colors.buttonNeutral0};
   }
+
   &[aria-disabled='true'] {
     ${getDisabledStyle}
     &:active {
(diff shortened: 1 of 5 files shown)


packages/strapi-design-system/src/Button/Button.js
```diff
@@ -27,19 +27,33 @@ const BoxFullHeight = styled(Box)`
...
 export const ButtonWrapper = styled(BaseButton)`
   align-items: center;
-  padding: ${({ theme, size }) => `${size === 'S' ? theme.spaces[2] : '10px'} ${theme.spaces[4]}`};
-  background: ${({ theme }) => theme.colors.buttonPrimary600};
+  background-color: ${({ theme }) => theme.colors.buttonPrimary600};
   border: 1px solid ${({ theme }) => theme.colors.buttonPrimary600};
+  height: ${({ size }) => {
+    let height = 32;
+
+    if (size === 'M') {
```
joshuaellis, line 37: Can we use `BUTTON_SIZES`?

packages/strapi-design-system/src/Button/Button.js
```diff
@@ -27,19 +27,33 @@ const BoxFullHeight = styled(Box)`
...
+      height = 36;
+    } else if (size === 'L') {
+      height = 40;
+    }
+
+    return `${height / 16}rem`;
+  }};
```
udimberto, line 44: #### Suggestion

IMO, will be good to put the button sizes at the file `src/themes/sizes.js`, like other components.
WDYT?

```suggestion
  height: ${({ size, theme }) => theme.sizes.button[size] || themes.sizes.button.S};
```

😂 _That snippet could not work, I'm not sure if will work properly_

maevalienard (approved):
Wooohooooo, looking great!!!! Thanks

udimberto (changes requested):
## A suggestion

Regarding the Button sizes "placement".



---

### Random comment

The way which we use the `Typography` component in almost everything, to control font-size, line-height and stuffs, it's weird, but I can be wrong to think that.

IMO, we should put the Typography styling-utilities available to all components we want, like `Box`, to support the styling-typography props without need to have the `Typograph` component rendering it's own `span` and messing with the parent-component height, by example.

Like:
```tsx
<Box variant="pi">
  Will render an ordinary div with the proper font-size and line-height for the variant "pi" of our DS, without a "span" wrapping the text-content.
</Box>
```
```

packages/strapi-design-system/src/Button/Button.js
gu-stav: Done in 875e6bc65b10f2beed1cac6deb09f1739922e415. Thanks - that's a very good idea!

packages/strapi-design-system/src/Button/Button.js
gu-stav: I went for @udimberto s suggestion to move these into the theme constants in 875e6bc65b10f2beed1cac6deb09f1739922e415 . I'm not sure how valuable that is, but it feels good to have this in one place with the other constants. Does that work for you?

joshuaellis: yep 👍🏼

joshuaellis (approved):

udimberto (approved):
### LGTM

PR #759, opened 2022-11-08
Typography: Fix attribute overwrites of variant styles

gu-stav:
### What does it do?

Re-orders the CSS for the `Typography` component. Custom attributes should come after `variantStyles`.

### Why is it needed?

Right now it is impossible to overwrite e.g. the `line-height` of a variant because of the way the CSS is ordered. Since there is no way to use `Typography` without a `variant` we could either:

1. Dis-allow overwriting font-size, line-height, font-weight and text-transform
2. Re-order the CSS in a way that makes overwrites possible

### Related issue(s)/PR(s)

- https://github.com/strapi/design-system/pull/758


joshuaellis (approved):

udimberto (approved):
### LGTM

gu-stav:
@joshuaellis Please leave the PR for now: there are too many snapshot updates that shouldn't happen and I want to investigate first why.

PR #764, opened 2022-11-10
[release]: 1.3.0

joshuaellis:
### What does it do?

* adds DateTimePicker

### Why is it needed?

* separates the vite bundler from DateTimePicker

### Notes

Any additional changes to the beta should be aimed at the `release/beta/1.3.0` branch, once we're happy it's stable then we can merge this branch and then release as stable.


joshuaellis:
the features in this branch will be released in 1.4.0

PR #766, opened 2022-11-10
Chore: Update WYSIWYG icons

gu-stav:
### What does it do?

Updates all WYSIWYG icons. Previously they included some padding, which caused them to render too small.

### Why is it needed?

This all started with the `More` icon that is supposed to look like that (used in an `IconButton`):

index-3

Guess what: it does not and looks much smaller instead:

| Current | Desired |
|-|-|
| index-2 | index |

It turned out that the problem came from the fact that all of these icons had some padding inline, which made them look smaller because e.g. the `IconButton` brings it's own padding.

Maeva also updated the UI Kit in Figma.

### How to test it?

Check the rich-diff in this PR.


gu-stav:
Good catch. I've made a mistake cleaning up the path props there 🤦🏼

ronronscelestes (approved):
LGTM 💅

joshuaellis:
I think these are duplicated in the `admin` app under `packages/core/admin/admin/src/icons`? They have the same names and aren't used. So maybe I should remove them there if they're here?

joshuaellis (approved):

gu-stav:
@joshuaellis Thanks - another good catch. I'll add a subtask and check!

PR #767, opened 2022-11-10
Fix: Required prop with inputs

### LINKED ISSUE #760: Required prop not working on TextInput component
### Describe the bug

Hello, i'm trying to use the prop required on TextInput, like the example below:

        <TextInput
          placeholder="FY23_"
          label="Identification"
          name="text"
          hint="Max 40 characters"
          error={getErrorName()}
          required={true}
          onChange={(e) => setName(e.target.value)}
          value={name}
        />

But it's not working well. The label is receiving the required prop, but the input field is not receiving.

image

This seems like a bug.

ronronscelestes:
## What does it do?

Fixes #760

- pass `required` prop to input
- stop passing `required` prop to `span` label 
- use `useField` context to pass `required` prop shared by `FieldInput` and `FieldLabel`
- clean and add stories for required state
- update a11y tests 

## Why is it needed?

`required` prop were only passed to `FieldLabel` to display a red asterisk and was wrongly given to the `label`
for [accessibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required) reasons `required` prop needs to be passed to input

## How to test it?

test that every input impacted by the changes passes down the `required` prop to the specific input


ronronscelestes:
curious to know if you see a way to handle `Select` and `Carousel` components, first one uses a `button` and the other one `div` 👀 @gu-stav @joshuaellis

gu-stav (changes requested):
We need to discuss adding a required attribute to `input`s first because the validation fires before the actual form `submit` event: https://codesandbox.io/s/dreamy-williams-edzj29?file=/index.html If used in a real form, we have no way I think to prevent the required validation and the native browser UI for it.

I'm not against this and believe it is a valuable a11y feature, but we need to discuss the consequences.

> curious to know if you see a way to handle Select and Carousel components, first one uses a button and the other one div 👀 @gu-stav @joshuaellis

Right now, it doesn't make sense to add this to both components and only keep the visual indicator. The fact that we can not apply it to them makes me think we should not add the attribute to some fields in the first place.

Good topic for the next DS sync.

joshuaellis:
> If used in a real form, we have no way I think to prevent the required validation and the native browser UI for it.

I haven't tested these solutions, but there's various articles that point to how we can still use the bedrock of HTML5 validation e.g. the required field but still implement our own error handling:

* https://www.aleksandrhovhannisyan.com/blog/html-input-validation-without-a-form/#reporting-input-validity-to-users
* https://adrianroselli.com/2019/02/avoid-default-field-validation.html
* https://oliverjam.es/blog/better-native-form-validation/

The last two are linked too in the first article anyway. So it might be possible to handle this without breaking changes & causing native browser error handling to occur 😄

ronronscelestes:
thanks for your feedbacks!

> the validation fires before the actual form submit event

nice catch @gu-stav - I forgot about this behavior and the default UI that comes with it

> we can still use the bedrock of HTML5 validation e.g. the required field but still implement our own error handling
thanks for the resources 

thank you for the resources @joshuaellis - interesting!

I was also wondering if we could only use `aria-required` as when I tested it it gave the same accessibility information on the input when using a voice reader, without blocking the event and imposing a default UI (tooltip) - wdyt?

ronronscelestes:
As discussed during the last DS sync, I replaced the `required` prop with `aria-required` for all inputs.
`required` caused the validation to fire before the submit.
`aria-required` brings the same accessibility feature without interfering with the validation.
(can be tested [here](https://codesandbox.io/s/bold-thunder-7skfes?file=/src/App.js))

https://user-images.githubusercontent.com/71838159/205682075-00c782a2-4cec-4e2f-b7f2-7785fc4741c8.mov



We will have to iterate on the Select and Carousel which contains no inputs to attach this accessibility attribute.

packages/strapi-design-system/src/Field/FieldLabel.js
```diff
@@ -17,19 +17,11 @@ const Action = styled(Flex)`
   }
 `;
 
-export const FieldLabel = ({ children, required, action, ...props }) => {
-  const { id } = useField();
+export const FieldLabel = ({ children, action, ...props }) => {
+  const { id, required } = useField();
```
gu-stav, line 21: I think we should keep the required prop, because otherwise we'd introduce a breaking change. What do you think about:

```suggestion
export const FieldLabel = ({ children, action, required: requiredDeprecatedProp, ...props }) => {
  const { id, required: requiredField } = useField();
  const required = requiredDeprecatedProp || requiredField;
```

packages/strapi-design-system/src/Combobox/Combobox.stories.mdx
```diff
@@ -75,9 +81,9 @@ This is the doc of the `Combobox` component
   </Story>
 </Canvas>
 
-## Error
+### Error
 
-This is the doc of the `Combobox` component
+`Combobox` can display error. The error message is customizable and replace the description line below the writing area.
```
gu-stav, line 86:
```suggestion
`Combobox` can display an error. The error message is customizable and replaces the description line below the input.
```

packages/strapi-design-system/src/Combobox/Combobox.stories.mdx
```diff
@@ -115,9 +141,9 @@ This is the doc of the `Combobox` component
   </Story>
 </Canvas>
 
-## Creatable
+### Creatable
 
-This is the doc of the `Combobox` component
+The creatable mode of `Combobox` allows to create a new option available as a value.
```
gu-stav, line 146:
```suggestion
The creatable mode of `Combobox` allows the creation of new options.
```

packages/strapi-design-system/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -89,7 +89,9 @@ TODO
   </Story>
 </Canvas>
 
-## Disabled
+### Disabled
+
+Depending on the permissions a user have or the status of an action, a `DatePicker` can be unreachable or unavailable yet.
```
gu-stav, line 94:
```suggestion
Depending on the permissions a user have or the status of an action, a `DatePicker` can be disabled.
```

packages/strapi-design-system/src/DatePicker/DatePicker.stories.mdx
```diff
@@ -118,9 +120,71 @@ TODO
   </Story>
 </Canvas>
 
-## Min and Max Dates
+### Error
+
+`DatePicker` can display error. The error message is customizable and replace the description line below the writing area.
```
gu-stav, line 125:
```suggestion
`DatePicker` can display an error. The error message is customizable and replaces the description line below the input.
```

packages/strapi-design-system/src/Field/FieldLabel.js
ronronscelestes: that's a good catch, thank you!

packages/strapi-design-system/src/NumberInput/NumberInput.stories.mdx
```diff
@@ -209,9 +211,34 @@ NumberInputs disabled field.
...
+        </Box>
+      );
+    }}
+  </Story>
+</Canvas>
+
+### With initial 0
 
-NumberInputs are always found in forms. They give users the possibility to add or edit numbers.
+`NumberInput` are always found in forms. They give users the possibility to add or edit numbers.
```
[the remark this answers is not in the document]
ronronscelestes: do you mean deleting the story description or the entire `initial 0` story? @gu-stav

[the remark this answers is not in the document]
gu-stav: Yes. Otherwise we'd have to repeat it everywhere. I don't think it adds much value in this case.

packages/strapi-design-system/src/Field/FieldLabel.js
```diff
@@ -17,19 +17,18 @@ const Action = styled(Flex)`
   }
 `;
 
-export const FieldLabel = ({ children, required, action, ...props }) => {
-  const { id } = useField();
+export const FieldLabel = ({ children, action, required: requiredDeprecatedProp, ...props }) => {
+  const { id, required: requiredField } = useField();
+  const required = requiredField || requiredDeprecatedProp;
+
+  if (requiredDeprecatedProp !== undefined) {
```
ronronscelestes, line 24: I added a console.warn here to inform about the deprecation, let me know what you think @gu-stav

packages/strapi-design-system/src/Field/FieldLabel.js
```diff
@@ -40,9 +39,10 @@ export const FieldLabel = ({ children, required, action, ...props }) => {
...
 
 FieldLabel.defaultProps = {
-  required: false,
   action: undefined,
+  required: undefined,
 };
+
 FieldLabel.propTypes = {
   action: PropTypes.element,
   children: PropTypes.node.isRequired,
```
joshuaellis, line 48: Can't comment on the line, but i would add jsdoc deprecated comment too :)

joshuaellis (approved):
the smallest change in the world, but not a blocker

packages/strapi-design-system/src/Field/FieldLabel.js
joshuaellis: below this line, above the required prop just to be clear.

ronronscelestes: ahh **regrets** I actually did before removing because I realized the component doesn't have its own storybook and it won't show but maybe I should add it anyway? 👀

joshuaellis: It should show in IDEs 🙏🏻

packages/strapi-design-system/src/CarouselInput/CarouselInput.js
```diff
@@ -25,13 +25,9 @@ export const CarouselInput = ({
...
   return (
-    <Field hint={hint} error={error} id={generatedId}>
+    <Field hint={hint} error={error} id={generatedId} required={required}>
       <Stack spacing={1}>
-        {label && (
-          <FieldLabel required={required} action={labelAction}>
-            {label}
-          </FieldLabel>
-        )}
+        {label && <FieldLabel action={labelAction}>{label}</FieldLabel>}
```
gu-stav, line 30: @ronronscelestes Is this label missing an id?

ronronscelestes: it gets it from `Field` 👀 @gu-stav 
image
image

gu-stav (approved):
Works for me. If you feel like you want to tackle my comment in another PR that would be fine. Well done!

packages/strapi-design-system/src/CarouselInput/CarouselInput.js
gu-stav: Yes, that's why we don't have to pass it on `<FieldLabel />`. There are a few where we pass it: `<FieldLabel id="..." />`

ronronscelestes: ouh will check that thanks!

ronronscelestes: done [here](https://github.com/strapi/design-system/pull/767/commits/a847144d9e83aedd1fd7d9a9acf9d9c634a2ff63) @gu-stav

gu-stav: Nice, thanks! Added a follow-up bug here: https://github.com/strapi/design-system/issues/800

PR #769, opened 2022-11-14
ProgressBar: Refactor component to use Box as base

gu-stav:
### What does it do?

- Refactors the component to use `Box` as base component
- Update progress-color from `neutral150` to `neutral0`

➡️ [Preview](https://design-system-git-enhancement-progressbar-strapijs.vercel.app/?path=/story/design-system-components-progressbar--m)

### Why is it needed?

- Follow the UI Kit

### Related issue(s)/PR(s)

- https://strapi-inc.atlassian.net/browse/CONTENT-220

### DIFF
--- a/packages/strapi-design-system/src/ProgressBar/ProgressBar.js
+++ b/packages/strapi-design-system/src/ProgressBar/ProgressBar.js
@@ -2,43 +2,37 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
-const ProgressBarSmall = styled.div`
-  background: ${({ theme }) => theme.colors.neutral600};
-  border-radius: ${({ theme }) => theme.borderRadius};
-  position: relative;
+import { Box } from '../Box';
 
-  width: 78px;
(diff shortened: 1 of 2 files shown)


ronronscelestes (commented):
I think the change of color makes it difficult to see the progress / might break accessibility?

before:
image
image


after:
image
image

I found neutral150 in the UI kit 👀
image

gu-stav:
@ronronscelestes The background-color on which this component will be used will change too. You can check the attached issue.

ronronscelestes:
> The background-color on which this component will be used will change too.

right I remember it now!
but still the UI kit is using neutral150, should we update it @maevalienard?
I guess it means we'll have to use a colored background behind the ProgressBar component every time?

gu-stav:
> I guess it means we'll have to use a colored background behind the ProgressBar component every time?

Yes. I've updated the story to reflect that, @ronronscelestes.

joshuaellis (approved):

maevalienard:
@ronronscelestes Thanks for the heads-up I'll add this to my to-do list :)

PR #770, opened 2022-11-14
Badge: Update colors and height according to the UI kit

gu-stav:
### What does it do?

- Updates the height of the `Badge` to match the UI Kit

| Before | After |
|-|-|
| Screenshot 2022-11-14 at 12 08 29 | Screenshot 2022-11-14 at 12 08 35 |

- Updates the colors
- Updates the `border-radius` from `4px` to `2px`

### Why is it needed?

Match the UI Kit

### Related issue(s)/PR(s)

- https://strapi-inc.atlassian.net/browse/CONTENT-220

### DIFF
--- a/packages/strapi-design-system/src/Badge/Badge.js
+++ b/packages/strapi-design-system/src/Badge/Badge.js
@@ -6,20 +6,23 @@ import { Flex } from '../Flex';
 import { Typography } from '../Typography';
 
 const Base = styled(Flex)`
-  border-radius: 2px;
-  height: ${16 / 16}rem;
+  border-radius: ${({ theme, size }) => (size === 'S' ? '2px' : theme.borderRadius)};
+  height: ${({ size, theme }) => theme.sizes.badge[size]};
 `;
 
-export const Badge = ({ active, textColor, backgroundColor, children, minWidth, ...props }) => {
+export const Badge = ({ active, size, textColor, backgroundColor, children, minWidth, ...props }) => {
+  const paddingX = size === 'S' ? 1 : 2;
+
   return (
     <Base
       inline
       alignItem="center"
       justifyContent="center"
       minWidth={minWidth}
-      paddingLeft={1}
-      paddingRight={1}
+      paddingLeft={paddingX}
+      paddingRight={paddingX}
       background={active ? 'primary200' : backgroundColor}
+      size={size}
       {...props}
     >
       <Typography variant="sigma" textColor={active ? 'primary600' : textColor}>
@@ -33,6 +36,7 @@ Badge.defaultProps = {
   active: false,
   backgroundColor: 'neutral150',
   minWidth: 5,
+  size: 'M',
   textColor: 'neutral600',
 };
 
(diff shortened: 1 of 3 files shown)


ronronscelestes (approved):
LGTM!
sad we don't have `2px` spacing or radius in the theme 🥲

packages/strapi-design-system/src/Badge/Badge.js
```diff
@@ -1,37 +1,44 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
 import { Flex } from '../Flex';
 import { Typography } from '../Typography';
 
+const Base = styled(Flex)`
+  border-radius: 2px;
+  height: ${16 / 16}rem;
```
joshuaellis, line 10: seem's unnecessary? couldn't this just be `1rem`? 🤔

joshuaellis (changes requested):

packages/strapi-design-system/src/Badge/Badge.js
gu-stav: I agree, but we always define values as px because that's what Figma displays. It avoids having the developer calculate px from rem values, when somebody needs to read this part.

joshuaellis: _hm_ wider discussion could be making `1rem === 10px` so it's actually a lot easier to work with....

joshuaellis (approved):

gu-stav:
@maevalienard What do you think about creating two variants for this component then? `size=S` and `size=M`?

maevalienard:
@gu-stav Usually, I'm not fan to introduce too many possibilities but I don't see the issue here. That would work to me :)

gu-stav:
@maevalienard Cool. I think in this case it could make sense. Would you like to go ahead and create both variants (S and M) in the UI Kit first?

maevalienard:
@gu-stav Is it just a change of border-radius or also a change of padding? If also padding could you please let me know about the two variants to have?

gu-stav:
@maevalienard Up to you. I imagine this might work:

S: height=16px, border-radius=2px
M: height=24px, border-radius=4px

Currently, the badge is height=24px and border-radius=4px.

maevalienard:
Sounds good to me. I'm changing the UI Kit accordingly right now

gu-stav:
@maevalienard I think it also might be useful to define:

S: padding-left/ padding-right: 1
M: padding-left/ padding-right: 2

Otherwise it looks a bit cramped. What do you think?

Preview: https://design-system-git-fix-badge-height-colors-strapijs.vercel.app/?path=/story/design-system-components-badge--size

maevalienard:
Updated ✅

gu-stav:
@maevalienard Just to avoid any misunderstanding: I'm looking at https://www.figma.com/file/zUx2bwEZId1RHZBM6zT14e/Strapi---Website-UI-Kit-%F0%9F%A7%A9?node-id=1431%3A31181&t=1y1oYB7XXIppy3b6-1 and can only see one variant which has once a height of 20px and once of 19px. In both cases the Typography doesn't match the `Badge` we have here.

Screenshot 2022-11-16 at 16 08 43

gu-stav:
Oh thanks. My bad :) 

The DS UI Kit looks much better: 

Screenshot 2022-11-18 at 10 41 10

:) However: the height of the medium badge is `25` which doesn't go well with our magic number of `4` doesn't it?

maevalienard:
I'll take that compliment 😂 

Indeed, I might have mistaken somewhere. I'll fix that right away, thanks :)

PR #771, opened 2022-11-16
[Fix] Remove alt as required in Avatar

ronronscelestes:
## What

Removing `required` for `alt` prop in Avatar component - decided after an alt fallback [discussion](https://github.com/strapi/strapi/pull/14880)
> Screenreaders will automatically read the filename if no alt attribute is provided.

### DIFF
--- a/packages/strapi-design-system/src/Avatar/Avatar.js
+++ b/packages/strapi-design-system/src/Avatar/Avatar.js
@@ -91,6 +91,7 @@ Initials.propTypes = {
 };
 
 Avatar.defaultProps = {
+  alt: undefined,
   preview: undefined,
 };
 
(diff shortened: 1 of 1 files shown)


gu-stav (approved):
Thanks!

joshuaellis (changes requested):
I think a better default would be `""` because iirc that informs screenreaders to ignore the image. wdyt?

ronronscelestes:
> I think a good default is the file's name because it reads at least the file's name, if not alt-text is set.

I agree - tested myself and I don't think having blank instead of filename by default would be great, user can still add `alt=""` if they think it's preferable in their case, wdyt?

image

joshuaellis (approved):
sure 👍🏼

PR #773, opened 2022-11-16
[NumberInput]: use useControllableState to handle number vs string inputs.

### LINKED ISSUE #658: Changing component's data with hook is not displayed on UI

- Node.js version: 16.14.2
- NPM version: 8.5.0
- Strapi version: 4.2.2
- Database: PostgreSQL 13  (Alpine)
- Operating system: Ubuntu 20.04.4

### Describe the bug

When updating component with beforeHook Strapi Admin UI is not changing value of number field. New value appears only after page refresh or interacting with field. (Put request is sending data)

1. Create collection/single
2. Create component with "Number" field
3. Add component to collection (tested as single component)
4. Add before update hook that somehow changes field value with entity service update (code will be below) 
5. Save and check network / field

### Expected behavior

Field value should refresh as it working with other field types inside component.

### Screenshots

After clicking save

Screenshot from 2022-07-05 03-12-43

Field focus

Screenshot from 2022-07-05 03-12-48

Remove focus

Screenshot from 2022-07-05 03-12-51

### Code snippets

```
module.exports = {
  beforeUpdate: async (event) => {
    console.log(event.params.data.test);

    if (event.params.data.test.id) {
      await strapi.entityService.update(
        'test.test',
        event.params.data.test.id,
        {
          data: { text: '1', number: 1, boolean: true },
        },
      );
    }
  },
};
```

### Additional context

Request:
```
{
      createdAt: "2022-07-04T23:58:43.896Z"
      createdBy: 1
      id: 1
      publishedAt: null
      sitemap_exclude: false
      test: {}
      updateField: "123345"
      updatedAt: "2022-07-05T00:10:28.626Z"
      updatedBy: 1
}
```
Response:
```
{
"id":1,
      "createdAt":"2022-07-04T23:58:43.896Z",
      "updatedAt":"2022-07-05T00:10:33.493Z",
      "publishedAt":null,
      "sitemap_exclude":false,
      "updateField":"123345",
      "test":{"id":4,"text":"1","number":1,"boolean":true}
}
```

### LINKED ISSUE #768: Moving components in dynamic zone

- Node.js version: 18.9.0
- NPM version: 8.19.1
- Strapi version: 4.3.9
- Database: Default (Sqlite)
- Operating system: Windows 10

### Describe the bug

I added a dynamic zone to a collection type and added 2 of the same component to the dynamic zone. The component consists of one text field and one number field.

The bug appears when you try to move the components up or down, depending on if a component is closed/minimized.

Here, both components are "open". I then try to move components up/down. See that the text field moves, but the number field doesn't?
<video src='https://user-images.githubusercontent.com/55314315/192141463-ce0c9cd3-b935-4950-9b4c-384c8821acd8.mp4'>

Here you can see, when one of the components are closed/minimized, it seems to work fine.
<video src='https://user-images.githubusercontent.com/55314315/192141596-6d3d40ce-8457-4de6-abf9-ad25eff646ab.mp4'>

Here is another interesting behavior:
<video src='https://user-images.githubusercontent.com/55314315/192141690-b74a5458-13ab-49d8-8cca-20404fe082cd.mp4'>

1. Create a collection type
2. Create a component with one short text field and one number (integer) field
3. Add a dynamic zone to the collection type
4. Add two of that same component to the dynamic zone
5. Fill in different values in the two components
6. Have both component views "open/maximized".
7. Try moving a component up or down and watch how the values in the fields behave.

### Expected behavior

I expect that the whole component, and not only one of the values, moves up/down.

joshuaellis:
### What does it do?

* uses useControllableState hook to manage the need to have a state be managed externally but also own it's own internal state

### Why is it needed?

* Mainly because if the component updates for an external source the internal state won't reflect this (e.g. dynamic zone component move)

### Related issue(s)/PR(s)

* closes #658 
* closes #768


tools/excludedFolders.js
```diff
@@ -1 +1,4 @@
-module.exports = process.env.IS_V2 === "true" ? ["helpers", ".DS_Store"] : ["helpers", ".DS_Store", "v2"];
+module.exports =
+  process.env.IS_V2 === "true"
+    ? ["helpers", "hooks", ".DS_Store"]
+    : ["helpers", ".DS_Store", "v2", "hooks"];
```
udimberto, line 4: Not blocking, just a suggestion to be more easier to read:
```suggestion
  process.env.IS_V2 === "true"
    ? [".DS_Store", "helpers", "hooks"]
    : [".DS_Store", "helpers", "hooks", "v2"];
```

udimberto (approved):
#### 😅 Looks complex to me 

But if it works, it works.

ronronscelestes (changes requested):
didn't fully understand the code yet but controlled state ❤️
mainly tested the components so far, found 2 strange behaviours:

decimals break when updating it with keyboard or arrows (plus or minus)

https://user-images.githubusercontent.com/71838159/203029644-80b965b1-0ae0-40f9-b6c0-1455a3ced370.mov



huge numbers give infinity when using minus with keyboard or arrows 
(tested on main to compare, seems it was broken there too since I couldn't update the value at all)

https://user-images.githubusercontent.com/71838159/203029463-496aba1f-4623-43f3-aef3-bf87d44caf71.mov

joshuaellis:
> decimals break when updating it with keyboard or arrows (plus or minus)

@ronronscelestes do you mean you're expecting the comma to stay instead of changing to a full stop?

ronronscelestes:
> @ronronscelestes do you mean you're expecting the comma to stay instead of changing to a full stop?

yes I would expect `1,1 -> 2,1 -> 3,1 -> 4,1` - etc, as it does in main branch 👀

joshuaellis:
> yes I would expect 1,1 -> 2,1 -> 3,1 -> 4,1 - etc, as it does in main branch 👀

you forget i'm British and i do `1.1 -> 2.1` 😭

joshuaellis:
> huge numbers give infinity when using minus with keyboard or arrows

Think this can be considered an enhancement to the input because we'd need to use something like `BigInt` as those numbers are too large for JS 🥲

ronronscelestes:
> Think this can be considered an enhancement to the input because we'd need to use something like `BigInt` as those numbers are too large for JS 🥲

true! I still wonder if having it as fixed number (in the video recording from `main` you can see that I cannot update the number with keyboard/arrows) is not better than the infinite bug, wdyt?
I'm not sure how it was done in the code though 🤔

joshuaellis:
IMO, it should be "broken", because at least then it's not changing the value. I also can't seem to get Infinity on either this branch or `main`. I think we should open a new issue to discuss handling of `BigInt`? Would that work for you? 😄

ronronscelestes:
> IMO, it should be "broken", because at least then it's not changing the value. I also can't seem to get Infinity on either this branch or `main`. I think we should open a new issue to discuss handling of `BigInt`? Would that work for you? 😄

works for me 😊

ronronscelestes (approved):
LGTM 💃

joshuaellis:
Opened #783 @ronronscelestes

PR #782, opened 2022-11-21
chore: Improve SimpleMenu component

mazzucchelli:
### What does it do?

Add the prop `onReachEnd` to the SimpleMenu popover

### Why is it needed?

Needed to append additional items when items are retrieved from a server.

### How to test it?

consume the prop from any `<SimpleMenu />` component in storybook

```
<SimpleMenu id="label" label={val} onReachEnd={() => console.log('END')}>
```

if enough items there should be a log in the console

### DIFF
--- a/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
+++ b/packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
@@ -98,6 +98,7 @@ export const SimpleMenu = ({
   onClose = () => {},
   size,
   popoverPlacement,
+  onReachEnd,
   ...props
 }) => {
   const menuButtonRef = useRef();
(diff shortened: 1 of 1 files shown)


packages/strapi-design-system/src/SimpleMenu/SimpleMenu.js
```diff
@@ -212,7 +220,14 @@ export const SimpleMenu = ({
         {label}
       </Component>
       {visible && (
-        <Popover onBlur={handleBlur} placement={popoverPlacement} source={menuButtonRef} spacing={4}>
+        <Popover
+          onBlur={handleBlur}
+          placement={popoverPlacement}
+          source={menuButtonRef}
+          onReachEnd={handleReachEnd}
+          intersectionId={shouldHandleReachEnd ? `popover-${menuId}` : undefined}
```
udimberto, line 228: Just an idea, not mandatory:
I was wondering if something like this it's good.
```tsx
const reactEndProps = typeof onReachEnd !== 'function' ? {} : {
  intersectionId: `popover-${menuId}`,
  onReachEnd,
};
```

```suggestion
          {...reachEndProps}
```

WDYT?

udimberto: As the team decided, during the DS Sync, we should disconsider my question/suggestion

udimberto (approved):
### LGTM

PR #787, opened 2022-11-24
Update import guidelines for components & use Strapi Eslint config

joshuaellis:
### What does it do?

* updates the story import guidelines as requested by @HichamELBSI 
* Uses the ESLINT config from strapi/strapi to keep things uniform
* Adds prettier checks to the CI


packages/strapi-design-system/src/Alert/Alert.js
```diff
@@ -140,5 +140,5 @@ Alert.propTypes = {
 };
 
 AlertIcon.propTypes = {
-  variant: Alert.propTypes.variant,
+  variant: Alert.propTypes.variant.isRequired,
```
ronronscelestes, line 143: shouldn't we leave the default variant defined by `defaultProps`? 👀

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.js
```diff
@@ -118,25 +122,30 @@ export const SimpleMenu = ({
...
     }
-  }, [didMount, visible]);
+  }, [didMount, visible, handleOpen, handleClose]);
 
   /* in case `label` is a custom react component, we know it is going to be
       a child of the menu button.
   */
   useEffect(() => {
-    if (React.isValidElement(label) && focusedItemIndex == -1) {
+    if (React.isValidElement(label) && focusedItemIndex === -1) {
```
ronronscelestes, line 148: wondering why it was written this way, we're sure `focusedItemIndex` is never a string?

joshuaellis: not sure, but yes it'll be a number 👍🏼

packages/strapi-design-system/src/Alert/Alert.js
joshuaellis: We don't export `AlertIcon` and the default is defined on it's parent, so i think it's probably fine to leave as is.

ronronscelestes (approved):
🔥

PR #788, opened 2022-11-25
[imports]: don't use duplicate imports as the application breaks

### LINKED ISSUE #781: [bug]: we have duplicate exports (due to the v2 folder)

all versions

### What's Wrong?

We currently export `Link` from `src/Link` and `src/v2/Link` therefore you can't just do `import { Link } from '@strapi/design-system'`. This isn't just isolated to `Link` but all v2 components.

### To Reproduce

Try `import { Link } from '@strapi/design-system'`.

### Expected Behaviour

To not crash and be able to use the components.

My solution would be to *not* export the `v2` components from `index` and instead _require_ a user to do a direct import of the component. 

I also would advocate we stop doing the `v2` system at all costs.

joshuaellis:
### What does it do?

* Removes `v2` from the exports list
* Doesn't export RadioGroup twice as well

### Why is it needed?

* As discussed in the DS sync we're removing the `v2` exports from the `index` as it breaks the package, they'll only be exported from `@strapi/design-system/v2`

### Related issue(s)/PR(s)

* resolves #781

### DIFF
--- a/packages/strapi-design-system/src/index.js
+++ b/packages/strapi-design-system/src/index.js
@@ -1,6 +1,3 @@
-/* eslint-disable import/export */
-// TODO: the import error will be fixed in a different PR
-
 export * from './Accordion';
 export * from './Alert';
 export * from './Avatar';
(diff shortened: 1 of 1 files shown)


ronronscelestes (approved):
LGTM 🙏

PR #789, opened 2022-11-27
Refactoring the DateTimePicker and the DatePicker and TimePicker state logic

simotae14:
### What does it do?

We want to move the Code of the DateTimePicker from the helper-plugin to the design system because it is better to divide the Design stuffs from the helper plugin.

### Decision
I think there is no reason to maintain the DateTimePicker inside the helper-plugin because it is a Design Component that can be used in other places. In addiction, in future we need to refactor it and the other field components to prevent a little bit the props drilling and transform them into controlled components.

### Consequences
I think we can have a more clear and correct division between components into the different repos. We need also to take care about the possible errors with the old versions of the Design System and strapi

Refactor the fields in order to be able to display an error state (red border) without displaying the error message.

Move the DateTimePicker in the Design System and have a beta components system that would display a warning to the user that this component is still under definition/development and that it is not recommended to use it because the code and the API will change soon.

### How to test it?
We need to test if we make some regressions in the code, in every section where we can use the DateTimePicker, how it works with content from another locale

### Why is it needed?
We need to move the logic from the helper-plugin to the Design System project


joshuaellis:
This PR isn't required, we'll merge `main` into the beta branch and release it as stable :)

PR #792, opened 2022-11-28
Convert codebase to Typescript

### LINKED ISSUE #633: Typescript Support or DefinitelyTyped
Hi there!

I was wondering if there are any plans for typescript support, or whether I should write type declarations and submit them to DefinitelyTyped. Which if I am not mistaken, is simply a matter of converting the propTypes and defaultTypes to typed functions and interfaces. 

Please guide me to the correct way of contributing regarding this.

joshuaellis:
### What does it do?

* Converts all files to typescript

### Why is it needed?

* types are good for you

### How to test it?

* The CI won't run fully until the full migration I think...

### Related issue(s)/PR(s)

* resolves partially #633


gu-stav:
@joshuaellis I've been thinking about this PR a bit: to move it forward, do you think it would be possible to publish the types separately so that e.g. the cloud-squad can opt-into it and we can adopt it in the CMS without a lot of breakages?

joshuaellis:
> @joshuaellis I've been thinking about this PR a bit: to move it forward, do you think it would be possible to publish the types separately so that e.g. the cloud-squad can opt-into it and we can adopt it in the CMS without a lot of breakages?

This is the purpose of DefinitelyTyped – we shouldn't publish our own types as this is the defacto method and is supported by many IDEs etc. IMO I wouldn't recommend it, it creates a dependancy on the DT org to approve things sometimes and it's commonly difficult to write types for code separately normally leading to issues (i've had a few with `threejs`).

gu-stav:
@joshuaellis I see your point, but I think we are not able to release the DS with TS support without a lot of errors and friction. I'd see this as an intermediate step, so that e.g. the cloud could start using it and in the monorepo we could experiment with it and see where the challenges lie.

joshuaellis:
What errors are you foreseeing? Maybe there's a way to mitigate.

My biggest concerns would be:
A. Relying on DT to publish (we have to write TS tests otherwise any PR requires their direct approval)
B. Contributors making the type changes on DT, with threejs it's basically null from the community and theirs is massive. This library is tweaked a lot and that speed just doesn't really work for DT imo

gu-stav:
@joshuaellis We have a lot of broken prop-types in the current codebase. These would become build-time errors and there is an unknown number of them. Same applies for plugins: we would introduce a whole new class of bugs if we'd release this without the user having to opt-in.

gu-stav (approved):

PR #795, opened 2022-11-29
[typescript]: Accordion, KeyboardNavigable, Stack, TextButton, Flex, Button, BaseButton

joshuaellis:
### What does it do?

Converts:
* Accordion
* KeyboardNavigable
* Stack
* TextButton
* Flex
* Button
* BaseButton

### Why is it needed?

* part of #792 

### Related issue(s)/PR(s)

* should be resolved *after* #794


packages/strapi-design-system/package.json
```diff
@@ -76,5 +76,9 @@
     "test:e2e:watch": "chokidar '**/__tests__/*.e2e.js' -c 'npm run test:e2e -- --retries=0 {path}'",
     "test:e2e:debug": "cross-env PWDEBUG=1 playwright test"
   },
+  "resolutions": {
+    "@types/react": "17.0.1",
```
gu-stav, line 80: Why didn't you use `17.0.2` here too?

packages/strapi-design-system/src/index.ts
```diff
@@ -44,7 +44,7 @@ export * from './Flex';
 export * from './Searchbar';
 export * from './Select';
 export * from './SimpleMenu';
-export * from './Stack';
+export * from './Stack/index.ts';
```
HichamELBSI, line 47: 🤔 Do you really need to precise `index.ts` ?

packages/strapi-design-system/src/Accordion/Accordion.tsx
```diff
@@ -53,95 +59,87 @@ const AccordionWrapper = styled(Box)`
...
    * If the `Accordion` is as child of an `AccordionGroup`, this prop will be set to `false` automatically.
    * The error message of the `AccordionGroup` will be shown below the group instead of the Accordion itself.
    */
-  hasErrorMessage: PropTypes.bool,
+  hasErrorMessage?: boolean;
   /**
    * The id of the component.
    */
-  id: PropTypes.string,
+  id?: string;
```
gu-stav, line 88: I know it wasn't set to required before, but isn't the id required in fact? Otherwise there would be conflicts when using multiple accordions at the same time? 🤔

packages/strapi-design-system/src/Accordion/Accordion.tsx
```diff
@@ -53,95 +59,87 @@ const AccordionWrapper = styled(Box)`
...
-        <Box paddingTop={1}>
-          <Typography variant="pi" textColor="danger600">
-            {error}
-          </Typography>
-        </Box>
-      )}
-    </AccordionContext.Provider>
-  );
-};
+export type AccordionSize = 'S' | 'M';
```
gu-stav, line 62: I'm wondering if we should centralize the sizing schema somewhere, because it will always be a subset of `S|M|L`? This way we can make it clear how it works, when new components are created using the `size` prop. Wdyt?

packages/strapi-design-system/src/Accordion/AccordionGroup.tsx
```diff
@@ -73,7 +81,7 @@ export const AccordionGroup = ({ children, footer, label, labelAction, error })
           {labelAction && <LabelAction paddingLeft={1}>{labelAction}</LabelAction>}
         </Flex>
       )}
-      <EnhancedGroup footer={footer}>{childrenArray}</EnhancedGroup>
+      <EnhancedGroup footer={Boolean(footer)}>{childrenArray}</EnhancedGroup>
```
gu-stav, line 84: Should we rename this prop (e.g. `hasFooter`) to make this clear?

packages/strapi-design-system/src/Flex/FlexProps.js
```diff
@@ -1,43 +0,0 @@
...
-  wrap: undefined,
-};
-
-export const flexPropTypes = {
-  alignItems: PropTypes.string,
-  basis: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-  direction: PropTypes.string,
-  /**
-   * Supports responsive values
-   */
```
HichamELBSI, line 24: These comments are useful for [ArgsTable](https://storybook.js.org/docs/react/writing-docs/doc-block-argstable). Can you make sure that the documentation is still displayed, please?

Screenshot 2022-12-05 at 13 17 03

HichamELBSI (commented):
Thanks @joshuaellis, good job! Can you just make sure we still have the documentation on the doc tab of the stories please? Also, the storybook build seems to fail 🤔

packages/strapi-design-system/src/Box/Box.tsx
```diff
@@ -1,5 +1,6 @@
 import { HTMLAttributes } from 'react';
 import styled, { CSSProperties, DefaultTheme } from 'styled-components';
+// import { ComponentType } from 'styled-components/node_modules/@types/react';
```
gu-stav, line 3: leftover

packages/strapi-design-system/src/Stack/Stack.tsx
```diff
@@ -1,42 +1,60 @@
...
+  shouldForwardProp: (prop, defPropValFN) => !transientProps[prop as keyof StackProps] && defPropValFN(prop),
 })`
   & > * {
     margin-top: 0;
     margin-bottom: 0;
   }
 
   & > * + * {
-    margin-top: ${({ theme, spacing }) => theme.spaces[spacing]};
+    margin-top: ${({ theme, spacing }) => extractStyleFromTheme(theme.spaces, spacing, undefined)};
```
gu-stav, line 23: I guess calling `extractStyleFromTheme()` is necessary to TS to infer the proper type, right? I'm wondering if it would be possible to implement this with getter methods or converting `theme.*` to proxy objects? This way we could keep the boilerplate out of styles and centralize the type inference?

packages/strapi-design-system/src/index.ts
gu-stav, line 47: Why is this one different than the others?

packages/strapi-design-system/yarn.lock
```diff
@@ -2802,11 +2802,6 @@
     regenerator-runtime "^0.13.7"
     resolve-from "^5.0.0"
 
-"@strapi/icons@^1.3.1":
-  version "1.3.1"
-  resolved "https://registry.yarnpkg.com/@strapi/icons/-/icons-1.3.1.tgz#475a234a40df8b7711e4739379f2667375f05bdc"
-  integrity sha512-whR/Yr09PuxkmklU3YL1iNyBvuqCTerep1+bLEVEUVijwQJiigqIC07uv/zGGdZKpTelstDGNqFWTF5JcdbHbg==
-
```
gu-stav, line 2809: Mmmh, why?

packages/strapi-design-system/src/index.ts
joshuaellis: I think my IDE did that, thanks for catching.

packages/strapi-design-system/src/Accordion/Accordion.tsx
joshuaellis: hmmmm we'd need a way for certain components to not use `L`.

packages/strapi-design-system/src/Stack/Stack.tsx
joshuaellis: Did you have an API in mind? I'm not super familiar with Proxy objects.

packages/strapi-design-system/src/Accordion/Accordion.tsx
joshuaellis: No the way the genId function works, we should be fine.

HichamELBSI (approved):
Good job @joshuaellis

packages/strapi-design-system/src/Accordion/Accordion.tsx
gu-stav: I guess that would be an enhancement anyways, so feel free to ignore it for now :)

packages/strapi-design-system/src/Stack/Stack.tsx
gu-stav: Something like:

```js
const spaces = [1,2,3];
const spacesProxy = {
  get(target, index) /* infer type */ {
    target[index];
  }
}

theme.spaces = new Proxy(spaces, spacesProxy);
```

This is just an idea: I have no idea if this could work ...

joshuaellis: I'm gonna open this a separate issue to not block the progression, I think it could be a neat idea and cleaner – you're totally right.

PR #799, opened 2022-12-06
Update LICENSE from MIT to Strapi MIT

gu-stav:
### What does it do?

In https://github.com/strapi/design-system/pull/636 I've made a mistake a added a standard MIT license instead of the Strapi MIT license. This PR reverts the changes in 5e54a86a08f1ce37409eac0ed990afcaae7a30a6 and bacc8686653e95fb116d9119f01fb52976772c50. 

### Why is it needed?

Legal requirements.

### DIFF
--- a/packages/strapi-design-system/LICENSE
+++ b/packages/strapi-design-system/LICENSE
@@ -1,20 +1,7 @@
-Copyright (c) 2012-2022 Scott Chacon and others
+Copyright (c) 2015-present Strapi Solutions SAS
 
-Permission is hereby granted, free of charge, to any person obtaining
-a copy of this software and associated documentation files (the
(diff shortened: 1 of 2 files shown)


alexandrebodin (commented):
We can use a basic MIT license with the right copyright holder I think. cc @Aurelsicoko

Aurelsicoko:
This license https://opensource.org/licenses/MIT with Strapi Solutions SAS as a copyright holder would be perfect

gu-stav:
Thanks! I've fixed + rebased the change.

alexandrebodin (approved):

Aurelsicoko (approved):

PR #814, opened 2022-12-19
[feature] Exp-432 new json input component

madhurisandbhor:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)


packages/strapi-design-system/src/InputJSON/InputJSON.js
```diff
@@ -0,0 +1,127 @@
...
+  const editorState = useRef(null);
+  const editorView = useRef(null);
+  const timerRef = useRef();
+
+  const getContentAtLine = (line) => {
+    return editorState.current?.doc?.line(line);
+  };
+
+  const markSelection = useCallback(({ message }) => {
+    let line = parseInt(message.split(':')[0].split('line ')[1], 10);
```
joshuaellis, line 28: I find this line really hard to read and understand what you're trying to achieve 🤔

joshuaellis (changes requested):
Quick first pass, overall there's a lot of code that isn't immediately understandable why you've done what you've done, i'd look at adding a lot of comments to really explain how things are working, it might lead to us being able to simplify / refactor the code to be clearer in the long run. But I don't see anything outrageous :)

packages/strapi-design-system/src/InputJSON/InputJSON.js
```diff
@@ -0,0 +1,127 @@
...
+  }, []);
+
+  const clearErrorHighlight = () => {
+    const docEnd = editorState.current?.doc?.length || 0;
+    editorView.current.dispatch({
+      effects: filterMarks.of((from, to) => to <= 0 || from >= docEnd),
+    });
+  };
+
+  const validateJSON = useCallback(
```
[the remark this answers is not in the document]
madhurisandbhor: Validation is not a part of codemirror here, but its taken from CMS InputJSON (jsonLint.js). and it's required to work with useEffect initially (dependency chain).

packages/strapi-design-system/src/InputJSON/InputJSON.js
```diff
@@ -0,0 +1,127 @@
...
+  );
+
+  const handleValidateJSON = useCallback(
+    ({ value, isOnChangeCallback }) => {
+      if (timerRef.current) {
+        clearTimeout(timerRef.current);
+      }
+
+      if (value) {
+        timerRef.current = setTimeout(() => validateJSON({ value, isOnChangeCallback }), WAIT);
```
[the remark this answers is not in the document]
madhurisandbhor: this serves 2 purpose :
1. It's mainly required for onChange handler, to avoid validating value for each and every input change. 
2. Also on load, if there is a wrong json string, we need validate it and highlight in the editor. So to get the editor state, we have createEditor callback which gets fired after useEffect. and so delay helps to get the editor state in the meantime.

packages/strapi-design-system/src/InputJSON/InputJSON.js
```diff
@@ -0,0 +1,127 @@
...
+    const docEnd = editorState.current?.doc?.length || 0;
+    editorView.current.dispatch({
+      effects: filterMarks.of((from, to) => to <= 0 || from >= docEnd),
+    });
+  };
+
+  const validateJSON = useCallback(
+    ({ value, isOnChangeCallback }) => {
+      try {
+        const formattedData = jsonlint.parse(value);
```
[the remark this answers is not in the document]
madhurisandbhor: Yes, validation is handled by jsonLint.js, as I mentioned above, but It does throw error if it's not able to parse the json string.

packages/strapi-design-system/src/InputJSON/InputJSON.js
madhurisandbhor: So parse error we get is of this format, and for highlighting we need to get the line number mentioned here. I know this is not readable, I will try to simplify by breaking it on separate lines. But if you have any other solution please suggest.
```
Error: Parse error on line 8:
...,      "b":6,""   }]
---------------------^
Expecting ':', got '}'
```

packages/strapi-design-system/src/InputJSON/InputJSON.js
```diff
@@ -0,0 +1,127 @@
...
+
+  const clearErrorHighlight = () => {
+    const docEnd = editorState.current?.doc?.length || 0;
+    editorView.current.dispatch({
+      effects: filterMarks.of((from, to) => to <= 0 || from >= docEnd),
+    });
+  };
+
+  const validateJSON = useCallback(
+    ({ value, isOnChangeCallback }) => {
```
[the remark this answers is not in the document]
madhurisandbhor: I have added onChange callback to update json object in parent component. This is being done after parsing the json string, which is also called on load. But there is no need to update json object on load, so skipping it with this flag.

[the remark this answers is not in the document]
joshuaellis: Okay these sorts of things should be documented in the code though, otherwise someone coming to refactor this in ~1 year, will not know these bits 😄 

Is passing a flag the best way? My concern is `onChange` will probably change it's identity _alot_ in the actual CMS codebase. It might be better to return a value from `validateJSON` and use that return to pass to the `onChange` handler. This means we keep the array dep smaller and more stable and we're not using a flag which i find confusing and muddying the water.

packages/strapi-design-system/src/InputJSON/InputJSON.js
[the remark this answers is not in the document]
joshuaellis: This should be written in a comment then to explain this, often code that throws is async

packages/strapi-design-system/src/InputJSON/InputJSON.js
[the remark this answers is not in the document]
madhurisandbhor: @joshuaellis can you please specify how you would do debounce in other way?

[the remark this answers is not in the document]
joshuaellis: i'd look at using a lib – perhaps https://github.com/biw/tiny-throttle which seems relatively harmless on bundle size :)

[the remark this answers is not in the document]
madhurisandbhor: Thanks I updated it. but I find [this](https://www.npmjs.com/package/throttle-debounce) more stable and popular except smaller bundle size than tiny-throttle. 
https://www.npmjs.com/package/throttle-debounce

packages/strapi-design-system/src/InputJSON/InputJSON.js
```diff
@@ -0,0 +1,140 @@
...
+    const parsedJson = await validateJSON(currentValue);
+
+    // onChange callback to update parsed json in parent
+    if (parsedJson) onChange(parsedJson);
+  };
+
+  /**
+   * debounceValidateJSON: to debounce onchange validating json.
+   */
+  const debounceValidateJSON = debounce(validateJSONOnChange, WAIT);
```
gu-stav, line 78: Do we really need to wait 500ms here? This seems to be rather long to me. I'd suggest something lower around 100ms.

packages/strapi-design-system/src/InputJSON/decorationExtension.js
```diff
@@ -0,0 +1,38 @@
+import { StateField, StateEffect } from '@codemirror/state';
+import { EditorView, Decoration } from '@codemirror/view';
+
+// Effects can be attached to transactions to communicate with the extension
+const addMarks = StateEffect.define();
+const filterMarks = StateEffect.define();
+
+const lineHighlightMark = Decoration.mark({
+  attributes: { style: 'background-color: yellow; color: black' },
+});
```
gu-stav, line 10: Now that we move this component into the DS: did you check these colors with Maeva? I feel like we could improve these a lot ...

markkaylor:
@gu-stav 

> create a JSON component (similar to JsonComponent) today and move it into the top-level of the repo.

Just to clarify, you would like me to move the current `JsonComponent` and rename it `JSON`? By top level do you mean `src/JsonComponent`?

Happy to make this change but just curious what's the reasoning? I'm also wondering if this component should even exist? Why not just have the codemirror setup directly in the `InputJson` component?

gu-stav:
@markkaylor Yeah we always have one component that is called `<Something>Input` which is a composition of label, description, error and the underlying component (e.g. `Field`) of things. If you need the plain JSON component (without label ...) somewhere, I'd suggest to move this to the top-level, otherwise just make it a private sub-component of `JSONInput`.

markkaylor:
@gustav I think I got most of your comments except I haven't added the tests for 

> move decorationExtension into utils/ and add tests

Technically I think we are testing this at a higher level since we have an e2e test that checks errors are highlighted

- I didn't really understand the need for the debouncing / timers. I removed them and everything seems to be working fine... @joshuaellis let me know if something went over my head there and I can add it back.
- I replaced the weird code for `jsonLint` with `jsonParseLinter` that comes with codemirror, it also gives us the line position so we don't have to parse the error message for that info.

I was able to link the DS in Strapi following Josh's instructions here: https://github.com/joshuaellis/strapi-design-system#connecting-to-strapi. I was also able to get it working with `yarn link`, I think we should add a section to the README like Josh has on his fork. Happy to open that PR.

Unfortunately after the yarn link I still haven't gotten the component working in the CMS. I've been struggling with this error:

```
Unrecognized extension value in extension set ([object Object]). This sometimes happens because multiple instances of @codemirror/state are loaded, breaking instanceof checks.
```

I found this issue https://github.com/uiwjs/react-codemirror/issues/216 (and a couple others) and I've tried a lot of different suggestions and worked on it a bit with @remidej, but still no luck, so if either of you have any ideas please let me know!

packages/strapi-design-system/src/Json/JsonInput.js
```diff
@@ -0,0 +1,139 @@
...
+import { Stack } from '../Stack';
+import { addMarks, filterMarks, lineHighlightMark } from './utils/decorationExtension';
+import { JsonInputContainer } from './JsonInputContainer';
+
+const StyledBox = styled(Box)`
+  outline: 1px solid ${({ theme, error }) => (error ? theme.colors.danger600 : 'transparent')};
+`;
+
+export const JsonInput = ({ id, label, value, error, hint, required, theme, onChange, disabled, labelAction }) => {
+  const [errorMessage, setErrorMessage] = useState(error);
```
markkaylor, line 17: ℹ️  Not sure about experience in the CMS yet but I thought it made sense to get more feedback about the error than just the highlighting.

packages/strapi-design-system/package.json
```diff
@@ -21,7 +21,7 @@
     "@floating-ui/react-dom": "^1.0.0",
     "@internationalized/number": "^3.1.1",
     "@radix-ui/react-use-callback-ref": "^1.0.0",
-    "@uiw/react-codemirror": "^4.19.1",
+    "@uiw/react-codemirror": "4.8.1",
```
markkaylor, line 24: ℹ️ Following this suggestion https://github.com/uiwjs/react-codemirror/issues/216#issuecomment-1172859107, this is the only solution I have so far for getting the input to work in the CMS.

packages/strapi-design-system/src/Json/JsonInput.js
markkaylor: Decided against this for now and instead just made the component behave as close to the current component as possible...

packages/strapi-design-system/package.json
gu-stav: Is that because we still have codemirror around in the CMS codebase or would that also happen if multiple JSON fields are used in one view?

packages/strapi-design-system/src/Json/JsonInput.js
```diff
@@ -0,0 +1,156 @@
...
+    },
+  });
+
+  useEffect(() => {
+    const currentEditor = editor.current;
+
+    if (currentEditor) {
+      setContainer(editor.current);
+    }
+  }, [setContainer]);
```
gu-stav, line 117: Should `editor.current` be a dependency here, so the effect runs once it changes?

packages/strapi-design-system/src/Json/JsonInput.js
```diff
@@ -0,0 +1,156 @@
...
+
+  return (
+    <Field error={error} hint={hint} required={required}>
+      <Stack spacing={1}>
+        {label && <FieldLabel action={labelAction}>{label}</FieldLabel>}
+        <JsonInputContainer id={id} ref={editor} error={error} {...boxProps} />
```
gu-stav, line 123: Could you also derive `error` from the field context (like `FieldError`) to not pass down the prop?

packages/strapi-design-system/package.json
markkaylor: I don't think it is related to multiple codemirror packages in the codebase. I tested removing the v5 codemirror in the CMS and upgrading it to v6 but still had the problem. This is error is also coming from the scoped @codemirror and not codemirror. Good question about multiple fields in one view I'll try that.

packages/strapi-design-system/src/Json/JsonInput.js
gu-stav, line 123: I think you could remove `id` here and just pass it through the restructured `boxProps`?

packages/strapi-design-system/src/Json/JsonInput.js
```diff
@@ -0,0 +1,156 @@
...
+import PropTypes from 'prop-types';
+import { jsonParseLinter, json } from '@codemirror/lang-json';
+import { useCodeMirror } from '@uiw/react-codemirror';
+
+import { Field, FieldLabel, FieldError, FieldHint } from '../Field';
+import { Stack } from '../Stack';
+import { JsonInputContainer } from './JsonInputContainer';
+import { markField, addMarks, filterMarks, lineHighlightMark } from './utils/decorationExtension';
+
+export const JsonInput = ({
```
gu-stav, line 11: It is really just a question: should we call it `JsonInput` or `JSONInput`?

packages/strapi-design-system/src/Json/JsonInput.js
markkaylor: I had it but the linter was telling me it shouldn't be there 🤔 

Screenshot 2023-01-09 at 11 02 13

packages/strapi-design-system/src/Json/JsonInput.js
markkaylor: `JsonInput` made more sense to me to stay consistent with the naming convention in the repo. What do you think, do you have a preference?

packages/strapi-design-system/src/Json/JsonInput.js
gu-stav: Ok, so better believe eslint than me!

packages/strapi-design-system/src/Json/JsonInput.js
gu-stav: I think I'd prefer "JSON", because it is an abbreviation. Also, [the spec uses uppercase letters only](https://datatracker.ietf.org/doc/html/rfc8259). @joshuaellis opinions?

joshuaellis: i'd make it `JSON` cause, it's `JSON` :)

packages/strapi-design-system/package.json
markkaylor: Mutliple fields in the same view seem to work fine

packages/strapi-design-system/src/Json/JsonInput.js
markkaylor: Sorry not sure I understand or I don't see how I can access the field context via `useField` from a styled component (`JSONInputContainer`)

joshuaellis (approved):
Great work, lot tidier and easier to understand. Looks like you've managed to "modernise" the code a lot ⭐

packages/strapi-design-system/src/Json/JsonInput.js
gu-stav: I see. 👍🏼

packages/strapi-design-system/src/JSON/JSONInput.js
```diff
@@ -0,0 +1,152 @@
...
+    if (lintErrors.length) {
+      highglightErrorAtLine(state.doc.lineAt(lintErrors[0].from).number);
+    }
+  };
+
+  const handleChange = (currentValue, viewUpdate) => {
+    validateJSON(viewUpdate);
+
+    // Call the parent's onChange handler
+    onChange(currentValue);
```
gu-stav, line 81: I'm wondering about one thing here: as far as I understand you'd call `onChange` every time, even if the JSON is not valid. This basically means we'd have to validate `currentValue` in the CM again.

It is a real question:

- should we export `validateJSON` so that we can re-use the same function outside?
- or should we only call `onChange` if the value is valid?
- or: (preferred I think) should have two callbacks: `onChange` and `onChangeValid` (good name to be found :D)

gu-stav (commented):
I agree with Josh: great work - well done! 👏🏼 Added one last comment/ question.

gu-stav:
@markkaylor Can you confirm this issue would be fixed by the new component? https://github.com/strapi/strapi/issues/15107

packages/strapi-design-system/src/JSON/JSONInput.js
markkaylor: Some confusion I think here is that although we are technically validating the JSON it is not a form validation but just _linting_ the input.

This implementation reproduces the current behavior of the JSON input in the CM where we:

lint the input:

https://github.com/strapi/strapi/blob/b9c09e19fc1be2c9eda438ca1539a998ca0a2d54/packages/core/admin/admin/src/content-manager/components/InputJSON/index.js#L146

and run the validation separately using `yup`:

https://github.com/strapi/strapi/blob/b9c09e19fc1be2c9eda438ca1539a998ca0a2d54/packages/core/admin/admin/src/content-manager/components/EditViewDataManagerProvider/utils/schema.js#L215

Form validation runs only `onSubmit` 

For your suggestions:
- I'm not sure we should export the `validateJSON` since it is very specific to codemirror. We could export the errors returned from `jsonParseLinter` But I'm not sure it is worth it since anyone can get the errors by calling `JSON.parse(value)`, that's what we do in the form validation above for the CM 

  _Side note: I think the error messages could be more explicit in the CM and execute onChange for this component, it's sort of a shame we just have a highlight for the error. I explored this a bit in this PR but I think we will have a translation issue and it would be the only input, as far as I know, that validates onChange. I see that as an improvement to consider for this component in another iteration_

- I think we always have to call the parent `onChange` otherwise the `value` prop we receive will never update when the json is invalid and that's the value we are passing to codemirror.
- Not sure I see what to do here actually but it seems like we might run into a similar problem as the second option.

I think I will rename `validateJSON` => `lintJSON`. Sorry for the long message and let me know if you see it differently or if you want to pair program.

markkaylor: > or a second parameter on the onChange callback indicating whether the value is valid or not. I think I'd prefer that.

Wouldn't that be the `error` prop?

markkaylor:
> Can you confirm this issue would be fixed by the new component?

It does fix the issue, you can tab into the component, but to lose focus you need to hit `esc`. https://codemirror.net/examples/tab/

packages/strapi-design-system/src/JSON/JSONInput.js
gu-stav: > I think I will rename validateJSON => lintJSON. Sorry for the long message and let me know if you see it differently or if you want to pair program.

That makes sense to me, thanks!

Let's leave the rest as it is then and keep the validation outside of component. 👍🏼

packages/strapi-design-system/src/JSON/JSONInput.js
```diff
@@ -0,0 +1,160 @@
...
+};
+
+JSONInput.defaultProps = {
+  label: undefined,
+  labelAction: undefined,
+  value: '',
+  error: undefined,
+  hint: undefined,
+  required: false,
+  theme: 'dark',
```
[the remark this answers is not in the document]
markkaylor: Great point and good catch, we still need to hard code it as dark on the `useCodeMirror` options object. updated.

gu-stav (approved):
🚀 Well done!

gu-stav (changes requested):
Ah no wait: I just saw in your PR you are importing:

```
import { JSONInput } from '@strapi/design-system/JSON';
```

I think this should become:

```
import { JSONInput } from '@strapi/design-system/JSONInput';
```

and therefore the folders needs to be renamed.

markkaylor:
@gu-stav updated the folder, btw I see this branch is targeting the feature branch. I guess we will merge into that branch first and do a beta release before merging into main?

gu-stav (approved):

gu-stav:
@markkaylor That sounds good to me.

PR #815, opened 2022-12-20
[fix] DateTimePicker and Select accessibility

ronronscelestes:
### DateTimePicker

- proposition to fix `DateTimePicker` accessibility issue:
  - when clicking on the label, it should focus on `DatePicker` component
  - for that we need `DatePicker` `input` to have an `id` corresponding to the `htmlFor` / `for` used by the `DateTimePicker` label
  - since by default `DatePicker` generates its own id, I passed a `parentFieldId` (⚠️ open to suggestion / ideas, not 100% comfortable with this solution)

https://user-images.githubusercontent.com/71838159/208694028-9ac33793-97a2-4ab6-bcea-0f903248516e.mov

### Select

- removing `as="span"` as it was breaking the link between the label and the button (e.g. clicking on the label wouldn't focus the button)
- proposition to improve `Select` accessibility by adding the `generatedId` used by the `Field` to button's aria-labelledby attribute
  -  allows to read the `Select` label when focusing on the `Select` button

https://user-images.githubusercontent.com/71838159/208694012-8549a6a6-50e4-4be1-87bd-c7c2b70ba34b.mov

## Questions

- would there be a better solution / refactor than passing a `parentFieldId`prop to `DatePicker` to ensure the link between the `DateTimePicker` label and the `DatePicker` input?
- any idea why `Select` label would be treated as a `span` and if removing that would cause any other issue?
- any issue possible that you could see/imagine with this fix?

### DIFF
--- a/packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
+++ b/packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
@@ -118,6 +118,8 @@ export const DateTimePicker = ({
         <Stack horizontal spacing={2}>
           <DatePicker
             data-testid="datetimepicker-date"
+            // DateTimePicker sharing its Field id to DatePicker component
+            parentFieldId={generatedId}
             name={name}
             ariaLabel={label || ariaLabel}
             error={typeof error === 'string'}
--- a/packages/strapi-design-system/src/DatePicker/DatePicker.js
+++ b/packages/strapi-design-system/src/DatePicker/DatePicker.js
@@ -12,6 +12,7 @@ import { getDefaultLocale } from '../helpers/getDefaultLocale';
 
 export const DatePicker = ({
   ariaLabel,
+  parentFieldId,
   initialDate,
   selectedDate,
   onChange,
@@ -24,7 +25,10 @@ export const DatePicker = ({
   id,
   ...props
 }) => {
-  const generatedId = useId('datepicker', id);
+  const internalGeneratedId = useId('datepicker', id);
+  // when used in a DateTimePicker, we want DatePicker input to use DateTimePicker id
+  // it allows accessibility feature as focusing the input when clicking on the label
+  const generatedId = parentFieldId ?? internalGeneratedId;
   const [visible, setVisible] = useState(false);
   const inputRef = useRef(null);
   const datePickerButtonRef = useRef(null);
@@ -109,6 +113,7 @@ DatePicker.defaultProps = {
   locale: undefined,
   initialDate: new Date(),
   onClear: undefined,
+  parentFieldId: undefined,
   placeholder: undefined,
   selectedDate: undefined,
   size: 'M',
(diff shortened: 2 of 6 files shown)


simotae14 (approved):
LGTM

christiancp100 (approved):
LGTM 🌻

packages/strapi-design-system/src/DateTimePicker/DateTimePicker.js
```diff
@@ -118,6 +118,8 @@ export const DateTimePicker = ({
         <Stack horizontal spacing={2}>
           <DatePicker
             data-testid="datetimepicker-date"
+            // DateTimePicker sharing its Field id to DatePicker component
+            parentFieldId={generatedId}
```
gu-stav, line 122: What do you think about just passing `id` instead? This way we could avoid adding a new prop entirely ... 🤔 

```
<Datepicker id={generatedId} />
```

ronronscelestes: mmh when `DatePicker` is used by itself, it needs to generate it's own `id` to pass to `TextInput`

if I pass the `DateTimePicker` `generatedId` to `DatePicker` through `id`, and then do:
```js
const generatedId = useId('datepicker', id);
```

`DateTimePicker` id and `DatePicker` would still be different 
- `DateTimePicker` id would be `datetime-label-1`
- `DatePicker` id would be `datepicker-datetime-label-1`

or I would need to remove the prefix `datepicker` when generating the id in `DatePicker` component - but would it mean risking id duplicates? 👀

gu-stav: Apologies. I didn't see this was waiting for me. I think it would be ok to remove the prefix from `DatePicker`, because `useId` uses a singleton pattern and therefore generates a new id for every call. Also the react@18 hook doesn't support / use prefixes so we should start to fade them out I think ...

joshuaellis (changes requested):
The hint isn't being assigned to the field in DateTimePicker....? I also agree we should avoid the parentId prop if we can...

ronronscelestes:
> The hint isn't being assigned to the field in DateTimePicker....?

I don't think it is for any field except if I'm missing something 🤔
do you mean that the hint should appear in the input `aria-label` / be read by a voice-over tool? @joshuaellis 
`TextInput`:
image

joshuaellis:
I think it should be reading `Description Line` too because that's "describing" the field.

ronronscelestes:
> I think it should be reading` Description Line` too because that's "describing" the field.

I agree!
I think it should be fixed for every inputs though (as I said same issue on `TextInput` and other inputs, I'm thinking of opening a Jira ticket and add it to the next sprint - wdyt?

joshuaellis:
Yep im fine with that.

joshuaellis (approved):

PR #817, opened 2022-12-21
Add 'Warning' variant to the 'Alert' component

ivanThePleasant:
### What does it do?

This PR adds `Warning` variant to the `Alert` component.

### Why is it needed?

Needed for product licensing business logic. 

### How to test it?

Pass the `warning` variant to the `Alert` component


gu-stav (approved):
🚀

PR #819, opened 2022-12-27
[fix] Tooltip misalignment 

ronronscelestes:
### Tooltip
Adding a `Tooltip` to one of the Table header would break the alignment 
Replacing the `span` ref wrapper with `Flex` seems to fix this issue
Open to suggestion for a better fix

### Cell
Remove unnecessary `div` if no `action` was passed

## Preview

|  before | after |  
|---|---|
| Screenshot 2022-12-27 at 09 41 56  | Screenshot 2022-12-27 at 09 42 07  |

### DIFF
--- a/packages/strapi-design-system/src/Table/Cell.js
+++ b/packages/strapi-design-system/src/Table/Cell.js
@@ -29,7 +29,7 @@ export const Th = ({ children, action, ...props }) => {
     <CellWrapper as={RawTh} {...props}>
       <Flex>
         {children}
-        <ActionWrapper>{action}</ActionWrapper>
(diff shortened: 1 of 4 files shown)


joshuaellis (approved):

gu-stav:
I'm not really sure we should do that, because: s a DS user I'd find it surprising the `Tooltip` component wraps my trigger element in a `Flex` (direction: row).

Imho the wrapping element should change as little as possible in terms of layout, otherwise, there might be unintended consequences. These could appear, if the tooltip is e.g. wrapping multiple elements, or if the wrapped text is very long because flex children have a different wrapping behavior than inline children.

I'm wondering: why does it fix the alignment issue for table headers?

PR #820, opened 2022-12-28
[enh] Avatar video support

ronronscelestes:
## What

- Adds video support for `Avatar`
- Allows `background` and `textColor` customization for `Initials`

## Test

- Use local video files and test `Avatar` component

### DIFF
--- a/packages/strapi-design-system/src/helpers/throwPropError.js
+++ b/packages/strapi-design-system/src/helpers/throwPropError.js
@@ -1,0 +1,25 @@
+import PropTypes from 'prop-types';
+
+/**
+ * @type {(otherProps: string[]) => (props: Record<string, unknown>, propName: string) => Error | undefined}
+ */
+
+export const throwPropErrorIfNoneAreDefined = (otherProps, propType) => (props, propName, componentName) => {
+  if (!props[propName] && otherProps.every((otherProp) => !props[otherProp])) {
+    return new Error(`One of the following props is required: ${propName}, ${otherProps.join(', ')}`);
+  }
+
+  return PropTypes.checkPropTypes({ [propName]: PropTypes[propType] }, props, 'prop', componentName);
+};
+
+/**
+ * @type {(otherProps: any{}) => (props: Record<string, unknown>, propName: string) => Error | undefined}
+ */
+
+export const throwPropErrorRequiredIf = (otherProps, propType) => (props, propName, componentName) => {
+  if (props[propName] === undefined && Object.entries(otherProps).every(([key, value]) => props[key] === value)) {
+    return new Error(`${propName} is required`);
+  }
+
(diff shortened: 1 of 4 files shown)


gu-stav (changes requested):
I think we need to discuss this approach. I'm not a big fan, because the behavior is very specific to the Strapi monorepo. Haven't thought about it much yet, but I think this is a good case to improve composability of the DS instead of passing it through a prop.

packages/strapi-design-system/src/helpers/throwPropError.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  return PropTypes.checkPropTypes({ [propName]: PropTypes[propType] }, props, 'prop', componentName);
+};
+
+/**
+ * @type {(otherProps: any{}) => (props: Record<string, unknown>, propName: string) => Error | undefined}
+ */
+
+export const throwPropErrorRequiredIf = (otherProps, propType) => (props, propName, componentName) => {
```
joshuaellis, line 19:
```suggestion
export const throwPropErrorIfRequired = (otherProps, propType) => (props, propName, componentName) => {
```

joshuaellis: found that a bit hard to read to begin with?

joshuaellis (changes requested):
Small request, rest of the work is lovely to see ⭐

packages/strapi-design-system/src/helpers/throwPropError.js
ronronscelestes: works for me 😊

gu-stav:
My main concern here is that we are overloading the `Avatar` component.

If we would say `Avatar`s can also be videos, I think that could work. The whole component doesn't make sense in the context of a list (because we are not rendering avatars, but assets).

Instead, I'd suggest we keep `Avatar` as limited as it is and create a new component - not sure what to call it, but it could be something about "assets" and "previews" because that's what it does in the end. An avatar is way more specific.

I think this could also be an interesting issue to discuss with design, because:

- this component is too small to render video player controls (so it would basically render a static image of the video)
- given that, how is the `preview` supposed to work? Shouldn't we do our best to preview a video player that can be interacted with? That would mean to take a different approach which doesn't have to be compatible with `Avatar`.
- I find that implementation very opinionated and specific to the CMS and I don't see how this would be useful e.g. in the context of the cloud

ronronscelestes:
I'll close this PR for now as we would need to discuss and refine this idea of an asset preview component

gu-stav:
@ronronscelestes Just curious: does that mean you agree? It's ok to disagree with me here :)

ronronscelestes:
> @ronronscelestes Just curious: does that mean you agree? It's ok to disagree with me here :)

I agree it feels like extending this component just for 1 CMS current need that is not really related to what an avatar means 
I also like the idea to rethink the hover/preview for the video - and that would make more sense to do that in a Preview component

PR #825, opened 2023-01-03
Update LICENSE from MIT to Strapi MIT for the monorepo

gu-stav:
### What does it do?

Updates the license of the monorepo from MIT to Strapi MIT.

### Why is it needed?

In https://github.com/strapi/design-system/pull/799 I forgot to update the license in the monorepo.


joshuaellis (approved):

joshuaellis:
To note, this License would never have been published to NPM as it's at the root of the monorepo.

Aurelsicoko (approved):
LGTM

PR #827, opened 2023-01-04
Fix/typescript building

joshuaellis:
### What does it do?

* fixes the build pipeline with Typescript support
* Add types to components that do not have types e.g. `CarouselInput` while it's not perfect, it does mean TS shouldn't complain about missing types, it produces a declaration file like this:

```ts
export function Carousel({ actions, children, label, nextLabel, onNext, onPrevious, previousLabel, secondaryLabel, selectedSlide, ...props }: {
    [x: string]: any;
    actions: any;
    children: any;
    label: any;
    nextLabel: any;
    onNext: any;
    onPrevious: any;
    previousLabel: any;
    secondaryLabel: any;
    selectedSlide: any;
}): JSX.Element;
```

### Why is it needed?

* So we can release an experimental version of the TS DS

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@babel/preset-react": "^7.18.6",
+    "@rollup/plugin-typescript": "^10.0.1",
     "@specifyapp/cli": "^0.0.51",
     "@strapi/eslint-config": "^0.1.2",
     "@tsconfig/node16": "^1.0.3",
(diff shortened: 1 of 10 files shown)


joshuaellis:
@gu-stav this is a really positive sign that declaration files are made automatically (although they're not accurate because they're `any` they do mean that in theory, TS shouldn't complain about missing types).

If an experimental release works I can ensure the testing suite also works before merging the main feature branch.

gu-stav (approved):
Well done 👏🏼 Vite starts to shine more an more.

PR #828, opened 2023-01-06
chore: use esnext

joshuaellis:
### What does it do?

* changes the target to `esnext`

### Why is it needed?

* Originally i had it at this or es2020 but then webpack caused issues in the admin app when trying to convert it down to 2015 🤔 Now this – https://github.com/strapi/strapi/pull/15369 is open, we can go back to being _super flexible_
* `esnext` performs as little transpilation as possible, this is perfect for a library because some people may want to support 2018 but if we compile to 2015, we're making life difficult for them – https://vitejs.dev/config/build-options.html#build-target

### DIFF
--- a/packages/strapi-design-system/vite.config.js
+++ b/packages/strapi-design-system/vite.config.js
@@ -11,7 +11,7 @@ export default glob('./src/**/!(*.spec|*.e2e).{js,svg}').then(async (paths) => {
       exclude: [],
     },
     build: {
-      target: 'es2015',
+      target: 'esnext',
       lib: {
         entry: {},
         formats: ['cjs', 'es'],
--- a/packages/strapi-icons/vite.config.js
+++ b/packages/strapi-icons/vite.config.js
@@ -9,7 +9,7 @@ export default defineConfig({
     exclude: [],
   },
   build: {
-    target: 'es2015',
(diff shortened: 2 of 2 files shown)


udimberto:
Sorry, team, I didn't have the bandwidth to validate it in the Cloud yet.

That it's my only concern: 
How this can impact/affects our Cloud project.

Hopefully nothing, probably in a good way.

But looks good to me

gu-stav:
@joshuaellis Can you confirm Vite/ webpack of the apps consuming the DS would take care to transpile everything inside of the DS according to their set ES target?

joshuaellis:
> @joshuaellis Can you confirm Vite/ webpack of the apps consuming the DS would take care to transpile everything inside of the DS according to their set ES target?

yeah so the DS publishes transpiled to ESNEXT and if you have a project that has a target of `2018` then all your code incl. NPM modules will be transpiled and bundled to 2018 – otherwise npm modules would cause browser issues e.g. if they used `?.` about a year or so ago (when it wasnt really a standard)

gu-stav (approved):
We set a transpilation target in the CMS and Vite (cloud) also has one, which means we should be save doing that. Thanks for the explanation!

PR #835, opened 2023-01-09
[enh] Initials customization

ronronscelestes:
## What

Allow background and textColor customization for `Initials` component 

### Snapshot

image

### DIFF
--- a/packages/strapi-design-system/src/Avatar/Avatar.js
+++ b/packages/strapi-design-system/src/Avatar/Avatar.js
@@ -70,24 +70,37 @@ const InitialsWrapper = styled(Flex)`
   }
 `;
 
-export const Initials = ({ children }) => {
+export const Initials = ({ children, background, textColor }) => {
   return (
     <InitialsWrapper
+      background={background}
       borderRadius="50%"
       width={`${avatarSize}px`}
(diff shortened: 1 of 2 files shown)


gu-stav:
@ronronscelestes Where does that come from? Does that fix an issue or is this linked to another PR?

ronronscelestes:
> @ronronscelestes Where does that come from? Does that fix an issue or is this linked to another PR?

it's a proposition
for the ML list view we create a brand new `Initials` like component just because it has a different background and text color

and other than a CMS issue, I think it would make sense to easily customize the colors of a text avatar with the DS color tokens - but it's just my opinion 🤷🏻‍♀️

gu-stav:
>  for the ML list view we create a brand new Initials like component just because it has a different background and text color

Where do we render the initials in the list-view?

ronronscelestes:
> Where do we render the initials in the list-view?

image

right now we are creating a custom component - that's why I thought it could make sense to allow some UI customization for `Initials` because it could be used there, only the color is different

gu-stav (approved):
That makes sense to me then!  🚀

joshuaellis (approved):

PR #842, opened 2023-01-13
[fix] MainNav scrollable

ronronscelestes:
## What 

As this [CMS community pull request](https://github.com/strapi/strapi/pull/15377) stated, the `MainNav` component does not resize properly 
This leads to a broken UI when e.g. having lots of elements in `NavSections`:
image

## Solution

**Changes applied to v1 & v2:**

Making `MainNav` wrapper Flex
Having `NavUser` not positioned absolutely anymore
Creating a `NavFooter` component to place the `NavCondense` component absolutely to `NavUser` 

### Discussion needed

I feel like `NavFooter` is inevitable to make a responsive/scrollable `MainNav` but would it be considered as a breaking change?
If yes do we have other options as I feel I've run out of ideas?


packages/strapi-design-system/src/Divider/Divider.js
```diff
@@ -6,6 +6,8 @@ import { Box } from '../Box';
 const DividerWrapper = styled(Box)`
   height: 1px;
   border: none;
+  /* If contained in a Flex parent we want to prevent the Divider to shink */
+  flex-shrink: 0;
```
ronronscelestes, line 10: without `flex-shrink: 0` here the `Divider` ends up being shrinked when there's not enough space (e.g. when the `MainNav` enters scrollable display)
is there another solution to fix that? 
image

(cc @joshuaellis @gu-stav)

packages/strapi-design-system/src/MainNav/MainNav.js
```diff
@@ -1,23 +1,26 @@
...
-const MainNavWrapper = styled(Grid)`
+const MainNavWrapper = styled(Flex)`
   width: ${({ condensed }) => (condensed ? 'max-content' : `${224 / 16}rem`)};
   background: ${({ theme }) => theme.colors.neutral0};
   position: sticky;
   top: 0;
   height: 100vh;
+  overflow-y: auto;
+  /* That doesn't work, overflow-y auto still hides the condensed toggle button when on condensed state */
+  overflow-x: visible;
```
ronronscelestes, line 15: I didn't find a fix for this yet, the `overflow-y: auto` or `scroll` (tried both) causes overflow-x to be hidden which we don't want for `NavCondense` and `notifications` UI:
Screenshot 2023-01-13 at 11 08 02

`overflow-x: visible;` do not fix this issue

(cc @joshuaellis @gu-stav)

joshuaellis:
> I feel like NavFooter is inevitable to make a responsive/scrollable MainNav but would it be considered as a breaking change?

No i don't consider it a breaking change. You haven't removed anything.

packages/strapi-design-system/src/Divider/Divider.js
joshuaellis: I think that's fine, alternatively (although maybe not possible) you could have wrapped the top bit with the divider in this circumstance.

packages/strapi-design-system/src/MainNav/MainNav.js
joshuaellis: Sorry I'm not familiar with the names you're using – you're saying the NEW flag is being cut off right?

ronronscelestes: oh yeah sorry, the "new" badge and also the button toggle next to the user name (it acts like the `MainNav` `overflow-y: auto` means `overflow hidden`)
image

packages/strapi-design-system/src/Divider/Divider.js
ronronscelestes: it would be possible but then you have to think about adding a `Box` when building your `MainNav` because you can choose to add a `Divider` or not, it's not integrated with the top "Strapi Dashboard" title:
image

joshuaellis: Yeah I thought so, it's fine as is 👍🏼

packages/strapi-design-system/src/MainNav/NavFooter.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Box } from '../Box';
+
+const BoxRelative = styled(Box)`
+  position: relative;
+`;
```
gu-stav, line 8: What do you think about adding a `position` prop to `Box`? I feel like this would help us in many places.

packages/strapi-design-system/src/MainNav/NavLink.js
```diff
@@ -62,13 +62,12 @@ const MainNavRow = styled(Flex)`
...
-  ${({ condensed }) =>
+  ${({ theme, condensed }) =>
     condensed &&
     `
 	  position: absolute;
-    transform: translate(35%, -50%);
-    top: 0;
-    right: 0;
+    top: -${theme.spaces[3]};
+    right:  -${theme.spaces[1]};
```
gu-stav, line 70: What are these numbers based on? Is that half the height/ width of the Badge?

packages/strapi-design-system/src/MainNav/NavSections.js
```diff
@@ -1,18 +1,24 @@
 import React, { Children } from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
 import { Stack } from '../Stack';
 import { Box } from '../Box';
 
+const BoxGrow = styled(Box)`
+  flex-grow: 1;
+  overflow-y: scroll;
+`;
```
gu-stav, line 10: Should this be a `Flex` then? Since `shrink` exists as a prop I think we should/ could add `grow` too.

gu-stav (commented):
Overall it looks good to me. Just as a word of caution: this won't have any effect on the CMS for now because it is using the v2 component. Therefore all these changes need to be implemented there too.

Are you planning to do that in a follow-up PR or should we keep these together?

packages/strapi-design-system/src/MainNav/NavFooter.js
ronronscelestes: I would love that - would do it in another PR though, I'll create a ticket to not forget!

packages/strapi-design-system/src/MainNav/NavLink.js
ronronscelestes: to be honest, they are mainly based on the visual aspect - checked and approved with @maevalienard - do you think we need another system?

gu-stav: No, but I think this deserves a comment :)

packages/strapi-design-system/src/MainNav/NavSections.js
ronronscelestes: adding grow to `Flex` sounds good 
but I don't think we need to use `Flex` here 
image
it causes other UI issues and demands adding other props to fix some of them (e.g. `align-items: start`) that I don't find useful there - wdyt?

maevalienard (approved):
Great job! Thank you :D

packages/strapi-design-system/src/MainNav/NavSections.js
gu-stav: Right, my bad!

ronronscelestes:
> this won't have any effect on the CMS for now because it is using the v2 component. Therefore all these changes need to be implemented there too.

thank you for the reminder! 
I'll add the changes to `MainNav` v2 in this PR!

joshuaellis (approved):

PR #848, opened 2023-01-19
[Storybook] Update & Move

joshuaellis:
### What does it do?

* moves storybook it its own dir for docs
* updates storybook & use webpack 5

### Why is it needed?

* means we are using the design-system and icons how our users would
* means we can set up a better dev env. – running `yarn develop` would set the watcher for the DS & Storybook dev
* you can run storybook in node 18 (finally)
* paves the way for a v7 upgrade with vite.

### NOTES

This is a POC – if we're happy with the direction i'll move all the stories over, imo it's the right thing to do.


udimberto (approved):
Great job, Josh.
So nice to see this constant improvements.
Big thank you.

---

**Not important observations**:

By installing `docs` dependencies, the `yarn.lock` had some changes.
I've tried with Node 16 and 18, they both modified the `yarn.lock`.

Taking this opportunity:
Do you think it's a good idea to add `.nvmrc` files and the property [`engines`](https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true#engines) to all of `package.json` in the project?
IMO, it can help.


```ts
// docs/.nvmrc
18
```

```ts
// docs/package.json
{
  ...
  "engines": {
    "node": "18.x",
    "yarn": "1.x"
  },
  ...
}
```

```ts
// packages/design-system/.nvmrc
16
```

```ts
// packages/design-system/package.json
{
  ...
  "engines": {
    "node": "16.x",
    "yarn": "1.x"
  },
  ...
}
```

joshuaellis:
> By installing docs dependencies, the yarn.lock had some changes.

This isn't finished yet so that could be why, but i'll keep an eye on it! :) 

> Do you think it's a good idea to add .nvmrc files and the property [engines](https://docs.npmjs.com/cli/v9/configuring-npm/package-json?v=true#engines) to all of package.json in the project?
IMO, it can help.

I don't have a problem with this and sure I think it's a fine idea, the engine should be a range iirc because we _can_ support node 14/16 too with the packages, not sure if the root needs min 18, but I agree we should also keep it as modern as possible 💪🏼

gu-stav:
I like the idea and would be fine with moving it to the top-level 👍🏼 Good thinking!

ronronscelestes (commented):
works for me too - would prevent some issues as forgetting to export new components (never happened to me 🙃)

gu-stav:
@joshuaellis I'm not a fan of having stories and components in different directories, because it is easier for them to be out of sync if changes are being made. Is there a way to keep them together even if we'd move the storybook installation?

joshuaellis:
> @joshuaellis I'm not a fan of having stories and components in different directories, because it is easier for them to be out of sync if changes are being made. Is there a way to keep them together even if we'd move the storybook installation?

🤔  arguably yes, but imo this goes against the point of the PR. I think having them separate is important because we're using the lib as users do to document it – e.g. there were random components that we were publishing to NPM 🤦🏼‍♀️ 

I see your concern about them going out of sync, but I think that's just got to be part of our commitment – similar to the contributors docs in the `strapi` app. Any MAJOR changes would fail the build of storybook :)

joshuaellis:
So assuming the E2E tests pass this will be good to merge. I need to fix vercel builds on there so I will set that up just before merging so when it goes into `main` it will all be fine.

PR #852, opened 2023-01-23
[chore] Box position 

ronronscelestes:
## What

Box `position` props was defined but not documented and often not used when it could have been
- documented with proptypes
- used as prop instead of custom style in numerous components
- larger cleanup by using `Box` props instead of custom

### DIFF
--- a/packages/strapi-design-system/src/Avatar/Avatar.tsx
+++ b/packages/strapi-design-system/src/Avatar/Avatar.tsx
@@ -70,8 +70,8 @@ export const Avatar = ({ src, alt, preview }: AvatarProps) => {
             background="neutral0"
             borderRadius="50%"
             position="absolute"
-            width={`${avatarSize / 16}px`}
-            height={`${avatarSize / 16}px`}
+            width={`${avatarSize}px`}
+            height={`${avatarSize}px`}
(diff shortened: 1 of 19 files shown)


packages/strapi-design-system/src/Avatar/Avatar.js
```diff
@@ -52,14 +41,26 @@ export const Avatar = ({ src, alt, preview }) => {
...
+        width={`${avatarSize}px`}
+        height={`${avatarSize}px`}
       >
-        {preview && previewVisible ? <Overlay /> : null}
+        {isHovering ? (
+          <Overlay
+            background="neutral0"
+            borderRadius="50%"
+            position="absolute"
+            width={`${avatarSize / 16}px`}
```
joshuaellis, line 57: That's very small! Did you mean rem?

joshuaellis (changes requested):

packages/strapi-design-system/src/Avatar/Avatar.js
ronronscelestes: thanks for catching it 😬

joshuaellis (approved):

PR #853, opened 2023-01-23
[chore] Remove prefix usage in useId

ronronscelestes:
## What

Remove prefix usage in `useId`:
- see the following discussion [here](https://github.com/strapi/design-system/pull/815#discussion_r1068107080)
-  prepare the future usage of React18 `useId` hook 

## Question

When an input is using the `Field` component: 
- We generate an id in the input component and `ref` it
- Then in the Field component we take the passed id and `ref` it again
Is that necessary?


joshuaellis:
Have you seen the radix implementation of useId? Could be good to copy that so r18 useId is used if there?

ronronscelestes:
> Have you seen the radix implementation of useId? Could be good to copy that so r18 useId is used if there?

does [that](https://github.com/strapi/design-system/pull/853/commits/fcba6f92b0138af5993456c1e3584c825e392d05) sound good to you? I didn't fully understand their implementation with `useState` and wasn't sure it was necessary for us, wdyt?

joshuaellis:
> does [that](https://github.com/strapi/design-system/pull/853/commits/fcba6f92b0138af5993456c1e3584c825e392d05) sound good to you? I didn't fully understand their implementation with useState and wasn't sure it was necessary for us, wdyt?

I think they'll have done it for a reason and i would be hesitant to change it 👀  also think we should bring in the `genId` func because we don't use it anywhere else i believe! :)

packages/strapi-design-system/src/Tooltip/Tooltip.js
```diff
@@ -16,8 +16,8 @@ const TooltipWrapper = styled(Box)`
 `;
 
 export const Tooltip = ({ children, label, description, delay, position, id, ...props }) => {
-  const tooltipId = useId('tooltip', id);
-  const descriptionId = useId('description');
+  const tooltipId = useId(id);
+  const descriptionId = useId(id);
```
joshuaellis, line 20: wouldn't this mean the tooltipID & descriptionIds are equal because we just pass back the `initialId`?

joshuaellis (changes requested):
Small Q

packages/strapi-design-system/src/Tooltip/Tooltip.js
gu-stav: Would it be possible to move away from `initialId`? It isn't supported in react@18 neither and I find it quite confusing :)

ronronscelestes: I'll check for the tooltip @joshuaellis 
I don't know if we should @gu-stav - I don't know potential consequences (people using their own ids when using our inputs), I didn't want to make the call and was not sure it needed to be done now 🤷🏻‍♀️

joshuaellis: > Would it be possible to move away from initialId? It isn't supported in react@18 neither and I find it quite confusing :)

Personally, i agree. _However_, it would break people's tests if they're looking for ids (imo bad practice though) and it would break if people use `ids` in the code – also imo bad practice not a react pattern. Do we want to deal with the potential fallout? I mean we can already see that it "broke" our tests already 😂

ronronscelestes: good catch @joshuaellis thanks! fixed [here](https://github.com/strapi/design-system/pull/853/commits/1e39e04569213ede53be27efaf4b12ef0dace41c)

joshuaellis (approved):
LGTM – i think lets see the impact of this PR before omitting passing ids 👍🏼

joshuaellis (approved):

gu-stav (approved):

PR #858, opened 2023-01-24
Accordion: forward shadow prop

gu-stav:
### What does it do?

It forwards the `shadow` prop on the `Accordion` component to the underlying `Box`.

### Why is it needed?

It allows us to render the Accordion with a shadow:

Screenshot 2023-01-24 at 16 20 01

Screenshot 2023-01-24 at 16 20 07

This is a design-requirement for review workflows. We can make use out of this for dynamic-zones too, which are currently using a style override:

```js
const StyledBox = styled(Box)`
  > div:first-child {
    box-shadow: ${({ theme }) => theme.shadows.tableShadow};
  }
`;
```

### DIFF
--- a/packages/strapi-design-system/src/Accordion/Accordion.tsx
+++ b/packages/strapi-design-system/src/Accordion/Accordion.tsx
@@ -91,6 +91,10 @@ export interface AccordionProps {
    */
   onToggle?: () => void;
   /**
+   * Shadow name (see `theme.shadows`)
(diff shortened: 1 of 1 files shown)


ronronscelestes (approved):

joshuaellis (approved):

maevalienard:
Merged but you also have my approval 😂 Thanks for the change!

PR #859, opened 2023-01-24
[Lerna]: handle parallel with storybook

joshuaellis:
### What does it do?

* removes storybook from `develop` scope
* uses workspaces in `lerna`

### Why is it needed?

* because deps were being confused
* the packages need to be built before `storybook` and with `parallel` (which is meant for long running tasks) it would fail – a lot.

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -17,7 +17,12 @@ jobs:
           node-version: 14
           cache: yarn
 
-      - uses: ./.github/actions/install-modules
+      - uses: actions/cache@v3
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-${{ matrix.node }}-${{ hashFiles('**/yarn.lock') }}
+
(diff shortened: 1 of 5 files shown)


gu-stav (approved):

gu-stav:
@joshuaellis As a note: while this fixes the storybook crashes for me, running `yarn develop` and `yarn storybook` at the same time leads to a "component not found" screen in storybook once a change has been made to a component. I'm not sure if this is introduced by this PR or whether this was already introduced when moving storybook out of the DS package?

joshuaellis:
> @joshuaellis As a note: while this fixes the storybook crashes for me, running yarn develop and yarn storybook at the same time leads to a "component not found" screen in storybook once a change has been made to a component. I'm not sure if this is introduced by this PR or whether this was already introduced when moving storybook out of the DS package?

This is at the moment expected behaviour (although not ideal), I need to into solutions in the meanwhile – my greatest piece of advice would be don't reload the webpage and it will correct itself.

PR #865, opened 2023-01-25
Add SingleSelect component

joshuaellis:
### What does it do?

* Introduces Radix-Select as the underlying component of a non-multii select.
* Moves `useIntersectionObserver` to `hook` and deprecates importing it from `helpers` (house cleaning)

### Why is it needed?

* It's a first step to a brighter world
* Completely lowers the amount of maintenance we have to do.

### How to test it?

* automated testing (I plan to add more once the CI actually passes)

### Related issue(s)/PR(s)

* resolves CONTENT-541


joshuaellis:
This is part of a bigger initiative hence why it's going into it's own branch, but in theory it could probably go to main.

I'm in the middle of adding a bunch of unit tests looking at different props and user behaviour but wanted to share the code so far. 

Had to refactor the DateTimePicker tests because they were a bit flakey :)

ronronscelestes (changes requested):
looks really good 💅
just a small UI issue when hovering on a Select option:
image

packages/strapi-design-system/src/Select/Option.tsx
```diff
@@ -1,107 +1,66 @@
...
+            zIndex={1}
+          />
+        </Box>
+        <Typography textColor={selected ? 'primary600' : 'neutral800'} fontWeight={selected ? 'bold' : undefined}>
           {children}
         </Typography>
       </Flex>
     </OptionBox>
+  ) : (
+    <SingleSelectOption {...{ children, value, selected, isChild, startIcon }} />
```
gu-stav, line 63: Does that mean the other branch is a `MultiSelectOption`? Would it make sense to improve readability to move it into it's own (local) component too?

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+import { useIntersection } from '../hooks/useIntersection';
+
+import { getThemeSize, inputFocusStyle } from '../themes/utils';
+
+export interface SingleSelectProps {
+  children: React.ReactNode;
+  /**
+   * @default "Clear"
+   */
+  clearLabel?: string;
```
gu-stav, line 23: I think you didn't change that but we should make it mandatory to overwrite this because otherwise the component won't be fully translateable.

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+  const clearRef = React.useRef(null);
+
+  /**
+   * These values are drawn out from the internals of the Radix component
+   * We can then use them to react to visual changes for the component
+   */
+  const [internalValue, setInternalValue] = React.useState('');
+  const [internalIsOpen, setInternalIsOpen] = React.useState(false);
+
+  const generatedId = useId('select', id);
```
gu-stav, line 86: Prefix should be removed.

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+    if (onChange) {
+      const shouldBeNumber = typeof passedValue === 'number';
+      onChange(shouldBeNumber ? Number(value) : value);
+    } else {
+      setInternalValue(value);
+    }
+  };
+
+  const handleClearClick: React.MouseEventHandler<HTMLDivElement> = (e) => {
+    console.log('calling clear');
```
gu-stav, line 105: Oopsi

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+   * We need to do this so the clear button can actually be
+   * used as opposed to the trigger swallowing the event.
+   *
+   * I'm not proud of it, but people are asking for it so maybe
+   * we can remove this in the future – https://github.com/radix-ui/primitives/issues/1569
+   */
+  const handleTriggerPointerDown: React.PointerEventHandler<HTMLButtonElement> = (e) => {
+    // @ts-ignore
+    if (clearRef.current && clearRef.current === e.target.closest('div')) {
+      console.log('preventing default');
```
gu-stav, line 125: Oopsi.

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+    } else {
+      setInternalValue(value);
+    }
+  };
+
+  const handleClearClick: React.MouseEventHandler<HTMLDivElement> = (e) => {
+    console.log('calling clear');
+    if (onClear) {
+      onClear(e);
+    }
```
gu-stav, line 108: You have a couple of these cases: should we just make them noop functions with a default assignment?

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+     * not be null. Because it uses a portal that (sensibly) is not mounted 24/7.
+     */
+    skipWhen: !internalIsOpen,
+  });
+
+  const value = typeof passedValue !== 'undefined' ? passedValue.toString() : internalValue;
+
+  return (
+    <Field hint={hint} error={error} id={generatedId} required={required}>
+      <Stack spacing={label || hint || typeof error === 'string' ? 1 : 0}>
```
gu-stav, line 151: `label` is a required prop so this will always be truthy.

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+        <FieldLabel action={labelAction}>{label}</FieldLabel>
+        <RadixSelect.Root
+          onOpenChange={handleOpenChange}
+          disabled={disabled}
+          required={required}
+          onValueChange={handleValueChange}
+          value={value}
+        >
+          <Trigger
+            aria-labelledby={`${generatedId} ${hintId} ${errorId}`}
```
gu-stav, line 161: Does it make sense to use `generatedId` here?

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+    width: 6px;
+
+    & > path {
+      fill: ${({ theme }) => theme.colors.neutral600};
+    }
+  }
+`;
+
+const Content = styled(RadixSelect.Content)`
+  z-index: 4;
```
gu-stav, line 253: Is there a reason this is not 2?

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+  background: ${({ theme }) => theme.colors.neutral0};
+  box-shadow: ${({ theme }) => theme.shadows.filterShadow};
+  border: 1px solid ${({ theme }) => theme.colors.neutral150};
+  border-radius: ${({ theme }) => theme.borderRadius};
+  padding: ${({ theme }) => theme.spaces[1]};
+  width: var(--radix-select-trigger-width);
```
gu-stav, line 259: Uff. Do you think we should hard-code this to avoid surprises?

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+  max-height: 15rem;
```
gu-stav, line 260: Would it be possible to explain in a comment what this number is based on?

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+    color: ${(props) => props.theme.colors.neutral600};
+    background: ${(props) => props.theme.colors.neutral150};
+  }
+
+  ${({ theme, $hasError }) => inputFocusStyle()({ theme, hasError: $hasError })};
+`;
+
+const DownIcon = styled(RadixSelect.Icon)`
+  & > svg {
+    width: 6px;
```
gu-stav, line 244: Should this be `rem`?

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,330 @@
...
+ */
+
+export interface SingleSelectOptionProps {
+  children: string | number;
+  startIcon?: React.ReactNode;
+  value: string | number;
+}
+
+export const SingleSelectOption = ({ children, startIcon, value }: SingleSelectOptionProps) => (
+  <SelectItem data-strapi-value={value} value={value.toString()}>
```
gu-stav, line 292: We we still need the data attribute?

packages/strapi-design-system/src/helpers/deprecations.ts
```diff
@@ -0,0 +1,17 @@
+export const prefix = '[@strapi/design-system]:';
+
+export const once = <TFunc extends (...args: any) => any>(fn: TFunc) => {
```
gu-stav, line 3: Could you add a test?

gu-stav (commented):
The lockfile changes look a bit weird to me. Should we maybe remove all lockfiles except the one at the root?

packages/strapi-design-system/src/Select/SingleSelect.tsx
gu-stav: Ah no ... it is not?! 🤔

gu-stav:
Another functional difference I can see: previously when using the arrow down key at the end of the list, the first option would be selected again.

packages/strapi-design-system/src/Select/Option.tsx
joshuaellis: Its the multi select yep, I thought about it but didn't want to bloat this PR with that change, the next ticket addresses combobox which is what a multi-select is 😄 so then it'll be moved out – wdyt? still needs moving?

packages/strapi-design-system/src/Select/SingleSelect.tsx
joshuaellis: this is a dynamic value so hardcoding it would defeat the point of it, when the select is rendered it's width is recorded and then we can make the dropdown adhear to it because it renders far away – https://www.radix-ui.com/docs/primitives/components/select#constrain-the-content-size

packages/strapi-design-system/src/helpers/deprecations.ts
joshuaellis: 🥲 sure

packages/strapi-design-system/src/Select/SingleSelect.tsx
joshuaellis: Don't think so, probably a hangover from how the old select worked.

packages/strapi-design-system/src/Select/Option.tsx
gu-stav: No all good then 👍🏼

packages/strapi-design-system/src/Select/SingleSelect.tsx
gu-stav: Are we sure radix will properly inject these properties into the parent application?

packages/strapi-design-system/src/Select/SingleSelect.tsx
joshuaellis: Actually i have made it required now... so yeah it'll always be truthy.

packages/strapi-design-system/src/Select/SingleSelect.tsx
joshuaellis: Think this is a hangover 👁️

packages/strapi-design-system/src/Select/SingleSelect.tsx
joshuaellis: I can't imagine why they wouldn't? If they don't, it's a bug in the library - I've not had any issue with radix.

We'll also be doing at least one beta incl. these changes anyway so we can revise if needs be.

packages/strapi-design-system/src/Select/SingleSelect.tsx
joshuaellis: Can do if you feel that strongly about it. I prefer this because it's a bit more explicit but idm - what do you wanna do?

gu-stav: Ok, keep it then 👍🏼

packages/strapi-design-system/src/Select/SingleSelect.tsx
joshuaellis: Isn't that a user of the library problem? 🤔 I'd rather give sensible defaults and make the component easier to integrate imo – but then yet again im probably biased because the default is always english 😅

joshuaellis:
>  previously the main select didn't have the focus outline after an option was selected (not saying this is better)

This is incorrect behaviour, we should always restore focus :) 

>  previously when using the arrow down key at the end of the list, the first option would be selected again.

The arrow keys also didn't correctly work when you opened the menu with a click and then use the arrow keys... This feature isn't part of the spec, it's something we've added and I don't think it's realistic for us to re-add it... IMO it's a loss im willing to live with?

joshuaellis:
Thanks for the first round of feedback @gu-stav & @ronronscelestes. Think it's ready for round 2 👀

ronronscelestes (commented):
second round:

> previously when using the arrow down key at the end of the list, the first option would be selected again.

> The arrow keys also didn't correctly work when you opened the menu with a click and then use the arrow keys... This feature isn't part of the spec, it's something we've added and I don't think it's realistic for us to re-add it... IMO it's a loss im willing to live with?

- I confirm that once you arrive at the last option, I have to use the up arrow to go back to the first one as it won't automatically bring me back to the first one as it did before. (see video)
- but as you say the keyboard focus is not lost when clicking first on the main Select now!

just to be sure do you mean that fixing the latter makes it difficult to not regress on the first one?  

- small UI details: after comparing to the previous version (see video), I think we now have a "double" outline to the main select, and we didn't have an outline on the option before - maybe we should confirm with design that it's okay?


https://user-images.githubusercontent.com/71838159/215044269-255e2617-be42-49f0-ab85-8571db37cb40.mov

joshuaellis:
> just to be sure do you mean that fixing the latter makes it difficult to not regress on the first one?

We don't have control over the logic/handling of how the list works without re-writing it as it's internal to the Radix component – imo I don't think it's a problem

> small UI details: after comparing to the previous version (see video), I think we now have a "double" outline to the main select, and we didn't have an outline on the option before - maybe we should confirm with design that it's okay?

Good eye 👁️, i'll take a look at that, it might be our focussing system!

gu-stav:
Except the one styling issue mentioned I'd only have on thing to add: I think when clicking "clear" the focus should also go back to the Select?

https://user-images.githubusercontent.com/2244375/215083570-a419c759-aca1-49ab-8b85-7790fb48f74c.mp4

joshuaellis:
Both comments addressed, also tidied up the focus because the select shouldn't be focused when you're focussed on clear imo :)

ronronscelestes (approved):
LGTM, well done!

packages/strapi-design-system/src/Select/SingleSelect.tsx
gu-stav: Ok, leave it then for now. Maybe it would be good to check and unify that behavior. I think if we want to push for not forget to translate it would be beneficial to make it mandatory.

packages/strapi-design-system/src/Select/Option.tsx
```diff
@@ -113,14 +72,62 @@ Option.defaultProps = {
...
+  indeterminate?: boolean;
+}
+
+const CheckMark = styled(Box)<CheckMarkProps>`
+  border: 1px solid
+    ${({ theme, selected, indeterminate }) =>
+      selected || indeterminate ? theme.colors.primary600 : theme.colors.neutral300};
+  height: 18px;
+  width: 18px;
+  background-color: ${({ theme, selected, indeterminate }) =>
```
gu-stav, line 88: I think you could make this a prop of the component instead to not customize too much CSS

packages/strapi-design-system/src/Select/Option.tsx
```diff
@@ -113,14 +72,62 @@ Option.defaultProps = {
...
+        background: url(${checkmarkIcon}) no-repeat no-repeat center center;
+        width: 100%;
+        height: 100%;
+        position: absolute;
+      }
+    `}
+`;
+
+const OptionBox = styled(Box)`
+  width: 100%;
```
gu-stav, line 122: Same for this one: could be a prop instead.

packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -0,0 +1,367 @@
...
+`;
+
+const Viewport = styled(RadixSelect.Viewport)`
+  padding: ${({ theme }) => theme.spaces[1]};
+`;
+
+const IconBox = styled(Box)`
+  background: transparent;
+  border: none;
+  border-radius: ${({ theme }) => theme.borderRadius};
```
gu-stav, line 303: Could this be just a `hasRadius` prop?

gu-stav (commented):
Minor suggestions.

The only think I've noticed functionally: when I click the label the component receives focus now, but it doesn't show the focus indicator. I think that is a bug.

packages/strapi-design-system/src/Select/Option.tsx
joshuaellis: This will be gone in the combobox refactor

packages/strapi-design-system/src/Select/Option.tsx
joshuaellis: Same as other, this will be redone in a later PR so will remember this going forward.

packages/strapi-design-system/src/Select/SingleSelect.tsx
joshuaellis: I'll make a GH issue for it as a `discussion`.

gu-stav (approved):
Well done! 🚀

PR #868, opened 2023-01-27
feat: add new package and shuffle deps to utilise workspaces

joshuaellis:
### What does it do?

* Adds a new `primitives` package for hosting our _super_ low level components (normally known as Raw in the DS storybook)
* Moves test suite base into the root of the package & react to both utilise workspaces (avoid dev conflicts) and mirror similar structure to the `strapi/strapi` repo

### Why is it needed?

* I'm in the process of rebuilding the combobox and to avoid a complicated component like we have in some spaces I plan to separate the primitive from the DS usage – in theory this will make it easier to integrate Radix's version when they add it
* This would reduce complexity at the DS level where we can focus on styling and our additional feature logic, components will then be simpler like seen in #865 

### How to test it?

* Test suite should still all pass


joshuaellis:
There's a broken test, but wanted to get the idea in front of you all before hand.

packages/primitives/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@strapi/primitives",
```
gu-stav, line 2: Should this be a private package for now?

package.json
```diff
@@ -37,20 +37,35 @@
...
+    "intl": "^1.2.5",
+    "jest": "^29.2.2",
+    "jest-environment-jsdom": "^29.2.2",
+    "jest-styled-components": "^7.1.1",
     "lerna": "^4.0.0",
     "lint-staged": ">=10",
     "plop": "^2.7.4",
     "prettier": "^2.8.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
```
gu-stav, line 68: Mh I don't think we should to that - only the packages that have a dependency on react should declare it. Why do want to have this here? What kind of dev conflicts are you running in?

> react to both utilise workspaces (avoid dev conflicts) and mirror similar structure to the strapi/strapi repo

In the stapi repo we also wanted to move it out of the global dependencies, because there it makes even less sense (the BE code doesn't know about react e.g.).

gu-stav (commented):
Overall I feel like a change of this magnitude should be discussed in a DS Sync first ...

package.json
joshuaellis: Because it's a dev dependency and my POV is there's no point declaring them per package when we're using workspaces, just like with jest.

Otherwise you have to update the dep in more than one place seems more of a chore imo.

I've seen it used in common practice on numerous repos - and it'll things such as "using more than one version of X" that crops up.

Strapi it makes sense but I think that's not comparable because of the blend of FS code.

packages/primitives/package.json
joshuaellis: I don't think it needs to be, this is going into a feature branch so it will need to be public to release betas.

joshuaellis:
> Overall I feel like a change of this magnitude should be discussed in a DS Sync first ... 

IDM, it's why I got everyone to review it here to avoid waiting. Again, it's going into a feature branch so it can be reverted. 

But if we'd prefer to discuss in a sync then sure 😊

udimberto (commented):
👏

markkaylor (commented):
- Are we expecting or already identified many components that we will have to create primitives for? (I just remember discussing combobox)
- Is a separate package a bit overkill if we only need to create a couple primitives that aren't provided by radix?

Otherwise I think the separate package is a cool idea so we can share our primitives 👍

joshuaellis:
> Are we expecting or already identified many components that we will have to create primitives for (I just remember discussing combobox)?

I think we'll need to "fork" the Select to enable `multiple` on the select because it's not implemented so theres another to be added and i'd definitely like to add RawTable bits

> Is a separate package a bit overkill if we only need to create a couple primitives that aren't provided by radix?

The amount of packages required to create a new primitive in a composable way (like radix) is very large so imo it's easier to keep it in it's own package so we can _clearly_ see the deps required just for the primitives stuff 👍🏼

PR #869, opened 2023-01-27
Bump lerna, fix HMR with Storybook & fix workflow running twice

joshuaellis:
### What does it do?

* Bumps lerna to most updated version
* Fixes HMR with storybook in dev mode
* Stops workflow running twice

### Why is it needed?

* Vite auto-clears dist which can make HMR in storybook BAD
* Lerna being updated gives us more features and auto-parallels / gives ability to create a pipeline / gives ability to local cache
* All the suites run twice 🥲


package.json
```diff
@@ -11,26 +11,22 @@
   ],
   "scripts": {
     "bootstrap": "lerna bootstrap",
-    "build-storybook": "lerna run build-storybook --stream",
-    "develop": "lerna run develop --parallel --ignore @strapi/design-system-docs",
-    "build": "lerna run build --stream",
+    "develop": "lerna run develop --stream",
+    "build": "lerna run build",
```
gu-stav, line 15: Does it make sense because this would always build everything (e.g. including the website, wouldn't it?)

ronronscelestes (approved):
works well on my side, thank you 🙏

packages/strapi-design-system/vite.config.ts
```diff
@@ -12,6 +12,7 @@ export default glob('./src/**/!(*.spec|*.e2e).{js,svg,ts,tsx}').then(async (path
       exclude: [],
     },
     build: {
+      emptyOutDir: false,
```
gu-stav, line 15: 💡

package.json
joshuaellis: not the website at the moment because it's not part of the `monorepo` but it's good for SB :)

joshuaellis: Which is what we're already doing

Issue #818, opened 2022-12-23
[bug]: Combobox sends you to bottom of page


Boegie19:
### What version of strapi-design-system are you using?

- Node version: v18.9.1
- Npm version: 8.19.2
- Strapi design version: 1.4.1
- Browser Google Chrome

### What's Wrong?

when I open a Combobox it sends me to the bottom of the page 
![Alt Text](https://i.gyazo.com/363ea37292dbbf885da239e43f15dbfd.gif)

### To Reproduce

Add Combobox
Add ComboboxOptions
make the body height 200% 
open the Combobox
get send to the bottom of your screen

### Expected Behaviour

It should stay near the Combobox  so that I don't have to scroll back to the top of the page

deanmchugh:
Im getting the same behaviour. Has anyone found a solution yet?

joshuaellis:
If either of you can provide a codesandbox demonstrating the issue that'd be really helpful to get the ball rolling 😄

Boegie19:
https://codesandbox.io/s/strapi-combobox-end-of-page-example-u6llvn
the it is caused by the css 
```css
#root {
  height: 200vh;
}
```
if I remote the root height it fixes itself if eve if the root height is only 100% it still goose down a little bit

Boegie19:
@joshuaellis Anything else you need?

joshuaellis:
> @joshuaellis Anything else you need?

No I haven't had the time to look yet and neither has another dev. I'll let you know when I have an update.

PR #871, opened 2023-01-30
[fix] Remove component with unnecessary computeScrollIntoView

markkaylor:
### What does it do?

- Removes a call to [maintainScrollVisibility](https://github.com/strapi/design-system/blob/803fbfde0bd7049a9a24dbf08a7a311ef9513e4a/packages/strapi-design-system/src/Combobox/utils.js#L92-L93) which is causing a bug when no option is selected.

### Why is it needed?

- When the combobox is in a container with a scrollable height clicking the combobox input causes a scroll down the page

### How to test it?
- Do the following on main to see the bug then on this branch to see the fix
- In the Combobox story, wrap the Combobox in a container with a massive height like 
```
  <div style={{height: '200vh'}}><Combobox/></div>
```
- Click the input to see the options, it should not scroll down the page.

### Related issue(s)/PR(s)

resolves: https://github.com/strapi/design-system/issues/818

https://github.com/strapi/strapi/pull/15504


packages/strapi-design-system/src/Combobox/ComboboxList.js
```diff
@@ -1,22 +0,0 @@
...
-
-export const ComboboxList = ({ options, activeOptionRef }) => {
-  useEffect(() => {
-    if (activeOptionRef.current) {
-      maintainScrollVisibility(activeOptionRef.current);
-    }
-  }, [activeOptionRef]);
-
-  return options;
-};
```
markkaylor, line 14: ℹ️  I didn't really understand the need for this component other than to call the useEffect and maintainScrollVisibility when there is an activeOptionRef. Removing this resolves the bug and the component seems to function fine without but let me know if there is something I'm missing...

joshuaellis: Might have been a hangover from when we weren't using `floating-ui` 🤔

joshuaellis (commented):
Not sure we can safely delete the file? Might be worth just deprecating it instead... I have a func we can cherry pick wdyt @gu-stav?

packages/strapi-design-system/src/Combobox/ComboboxList.js
```diff
@@ -1,8 +1,12 @@
 import { useEffect } from 'react';
 import PropTypes from 'prop-types';
-
 import { maintainScrollVisibility } from './utils';
 
+/**
+ * @deprecated
+ * This component was deprecated in 1.6.3. It will be removed in the next major release 2.0.0.
+ */
```
markkaylor, line 8: @joshuaellis not sure if this is all I need to do to deprecate the file 🤔. You mentioned you had a function to cherry pick, but not sure I understood exactly what that would be for.

joshuaellis: This will be done later, so thanks for leaving the note! 😄 

What do you think about removing the effect so anyone using the component has it fixed?

joshuaellis (approved):

PR #836, opened 2023-01-10
[chore] Add connecting to Strapi section to CONTRIBUTING

markkaylor:
### What does it do?

Adds a section describing how the design system can be connected to Strapi for testing in the CMS

### Why is it needed?

Testing in isolation using Storybook is great but you might also want to test your work in the CMS

### How to test it?

Make a change somewhere in the design system, then follow the steps to link the design system to your Strapi application. Verify the local version of the DS is being used.

--- 

More or less the same as https://github.com/joshuaellis/strapi-design-system#connecting-to-strapi

PR #872, opened 2023-01-31
workflows not running but are required

joshuaellis:
### What does it do?

* removes the path filtering for PR workflows

### Why is it needed?

* tests are "required" but then they don't run so PRs are left in limbo

### Related issue(s)/PR(s)

* #836

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,12 +3,11 @@ on:
   push:
     branches:
       - main
-  pull_request:
-    types: [opened, reopened]
-    paths:
-      - 'packages/**'
-      - '!docs/**'
-      - '!website/**'
+  pull_request: {}
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
 
 jobs:
   typecheck:


ronronscelestes (approved):

.github/workflows/tests.yml
```diff
@@ -3,12 +3,11 @@ on:
   push:
     branches:
       - main
-  pull_request:
-    types: [opened, reopened]
-    paths:
-      - 'packages/**'
-      - '!docs/**'
-      - '!website/**'
+  pull_request: {}
```
gu-stav, line 6: Do you maybe want to reverse the behavior to be in sync with the monorepo?

```suggestion
  pull_request:
        paths-ignore:
          - 'docs/**'
          - 'website/**'
```

joshuaellis: there's actually a problem in the monorepo where if you add something to docs in the same PR as other code, all the tests skip... So i think i'd rather be over-interested in the test suite running – with that context, wdyt?

gu-stav: Yeah we'd have to introduce the same skip actions here too then I suppose. Whatever you prefer: just wanted to throw the idea out :)

PR #874, opened 2023-01-31
Remove templates

joshuaellis:
### What does it do?

* Removes templates
* Removes `plop`
* Removes `.specifyrc.json` because it should have been removed previously

### Why is it needed?

* We don't believe they're used by anyone so there's no value in keeping them.


gu-stav (approved):

simotae14 (approved):

joshuaellis:
I've shared this in `engineering` and i'll give it a couple of days to see if anyone objects :)

ronronscelestes (approved):

udimberto (approved):

markkaylor (approved):

zelumine (approved):

jhoward1994 (approved):

package.json
```diff
@@ -45,7 +45,6 @@
     "husky": "=4",
     "lerna": "^6.4.1",
     "lint-staged": ">=10",
-    "plop": "^2.7.4",
```
markkaylor, line 48: I think you can also remove the generate command https://github.com/strapi/design-system/blob/014fa87b3c4917b00ddeb5afccb856f8ed66bc5a/package.json#L17

PR #876, opened 2023-02-01
[fix] borderRadius on json input

markkaylor:
### What does it do?

- Fixes a small gap between the editor and the border

before

json-before

after

Screenshot 2023-02-01 at 16 09 39

### How to test it?

Check if there is a gap


remidej (approved):
Tested locally ✅

PR #878, opened 2023-02-02
BaseButton: Extend Flex instead of plain button

gu-stav:
### What does it do?

Improves `BaseButton` so that instead of inheriting from `styled.button` it inherits from `Flex`. It also refactors components building on top of it, such as `LinkButton`, `Button` and `IconButton`.

This means all of these components can now be styled with props the `Box` and `Flex` components can handle.

- Button: https://design-system-git-enhancement-base-button-flex-strapijs.vercel.app/?path=/story/design-system-components-button--base
- LinkButton: https://design-system-git-enhancement-base-button-flex-strapijs.vercel.app/?path=/story/design-system-components-linkbutton--base
- IconButton: https://design-system-git-enhancement-base-button-flex-strapijs.vercel.app/?path=/story/design-system-components-iconbutton--base

### Why is it needed?

This will allow using all props defined for `Flex` and `Box` to be used on `BaseButton`. I found it quite confusing I can e.g. not use `background` or `border` props.


ronronscelestes (commented):
I like the idea and I didn't find any UI regression!
I'll let Josh help on the TS side 😊

packages/strapi-design-system/src/Button/Button.tsx
```diff
@@ -23,49 +24,28 @@ const LoadingWrapper = styled.div`
   will-change: transform;
 `;
 
-const BoxFullHeight = styled(Box)`
-  height: 100%;
-`;
-
```
gu-stav, line 29: Assigning a height of 100% actually leads to a problem which is visible e.g. here https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-button--full-width

joshuaellis (commented):
Just wondering if there was a particular reason not take this opportunity to convert those three files to TS? 🤔

joshuaellis:
you have a regression with button variants
CleanShot 2023-02-02 at 14 17 27@2x

gu-stav:
@joshuaellis Why not. 👍🏼  I think I'd need your help at some point and would do it after this PR is ready to avoid more regressions but then I think it is a good learning opportunity.

Just to be clear: are you ok with the direction of this PR? I've seen a few regressions and will sort them out - just wanted to start exploring where refactoring the BaseButton would lead me to and I think it is quite promising.

joshuaellis:
I think surface level it looks fine, i'd be interested to see if it works as anticipated with something like `LinkWrapper` where you can pass an `as` does it override the correct settings / styles? 🤔

joshuaellis:
> The TypeScript parts probably needs some more love. I didn't understand why disabled wasn't part of HTMLButtonElement already

_Probably_ because the generic extends `HTMLElement` and `disabled` doesn't exist on every element so you could extend the props with `HTMLImageElement` instead of `HTMLButtonElement` which iirc will not have `disabled` so it can't guarantee its safe.

gu-stav:
@joshuaellis I think this is ready for review 🙏🏼 I haven't found any visual regressions anymore and merged your previous TS work 🚀

udimberto (approved):
### LGTM

joshuaellis (approved):
If you feel like it, LinkButton v2 isnt TSX 👀

gu-stav:
@joshuaellis I'm happy to continue this journey in a follow-up PR ✨

PR #881, opened 2023-02-02
Box: Extend transient props

gu-stav:
### What does it do?

Ensures the `Box` component does not render the following props as HTML attributes:

- `width`
- `height`
- `cursor`
- `color`

### Why is it needed?

Became evident in https://github.com/strapi/design-system/pull/878.

### How to test it?

Follow the 🐇 automated tests.

### DIFF
--- a/packages/strapi-design-system/src/Box/__tests__/Box.spec.tsx
+++ b/packages/strapi-design-system/src/Box/__tests__/Box.spec.tsx
@@ -24,9 +24,11 @@ describe('Box', () => {
     },
   );
 
-  it.each(['color', 'cursor'])('does not render color or cursor props as HTML attributes', (prop) => {
-    const { container } = setup({ [prop]: 'something' });
-    expect(container).not.toHaveAttribute('color');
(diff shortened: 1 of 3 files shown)


packages/strapi-design-system/src/Box/__tests__/Box.spec.tsx
```diff
@@ -1,36 +1,35 @@
...
+    },
+  );
 
-      const el = screen.getByText('Hello world');
-      expect(el).toHaveStyle(`${spacingProp}: 16px`);
+  it.each(['color', 'cursor', 'height', 'width'])(
+    'does not render color or cursor props as HTML attributes',
+    (prop) => {
+      const { container } = setup({ [prop]: 'something' });
+      expect(container).not.toHaveAttribute('color');
```
joshuaellis, line 31: Should we be using `prop` here?

gu-stav: 🤦🏼

joshuaellis (approved):

PR #887, opened 2023-02-06
[Popover] expose primitives via composition

joshuaellis:
### What does it do?

* Refactors the popover to be more composable
* exposes the layers as components (Content / Scrolling)
* Converts to TS

### Why is it needed?

* Some instances of the popover do not want a scrollable container nor do they want the default padding

### How to test it?

* Unit testing


joshuaellis:
@ronronscelestes you'll then need to do something like this:

```ts
import { Portal, PopoverPrimitives } from '@strapi/design-system'

const MyComponent = () => {
	return ( 
		<Portal>
			<PopoverPrimitives.Content>
				<h1>hey, listen!</h1>
			</PopoverPrimitives.Content>
		</Portal>
	)
}
```

ronronscelestes (approved):
LGTM 🙏

PR #888, opened 2023-02-06
[fix] Visually center Play icon

ronronscelestes:
## What

`Play` icon looked not centered because of an optical illusion

- [x] Visually center the icon directly into the path

| Before | After |
|-|-|
|  Screenshot 2023-02-06 at 12 25 49 | Screenshot 2023-02-06 at 12 24 06 |


joshuaellis (approved):

joshuaellis:
Should be pointed at the correct release branch :)

packages/strapi-icons/assets/icons/Play.svg
```diff
@@ -1,3 +1,3 @@
-<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M4 20.2516V3.74842C4 2.975 4.84039 2.49435 5.50702 2.88648L19.5347 11.1381C20.192 11.5247 20.192 12.4753 19.5347 12.8619L5.50702 21.1135C4.84039 21.5057 4 21.025 4 20.2516Z" fill="#212134"/>
-</svg>
+<svg width="24" height="24" viewBox="0 0 8 8" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M2 0.934258C2 0.534911 2.44507 0.296715 2.77735 0.518233L7.37596 3.58398C7.67283 3.78189 7.67283 4.21811 7.37596 4.41603L2.77735 7.48177C2.44507 7.70329 2 7.46509 2 7.06574V0.934258Z" fill="#212134"/>
+</svg>
```
gu-stav, line 3: You could also drastically reduce the size of the SVG by using:

```html
<svg width="24" height="24" viewBox="0 0 8 8" fill="none" xmlns="http://www.w3.org/2000/svg">
  <path d="M2 .93c0-.4.45-.63.78-.41l4.6 3.06c.3.2.3.64 0 .84l-4.6 3.06A.5.5 0 0 1 2 7.07V.93Z" fill="#212134"/>
</svg>
```

ronronscelestes: extremely curious: how do you know how to do that? 👀

gu-stav:
I'm wondering a bit if we should do that: in that particular case (rendering the play icon in a circle) it might be the solution, but overall I think to center the icon like any other is a good idea because we don't know the context in which it would be used. 🤔 

How would that look like e.g. in an `IconButton`?

ronronscelestes:
> How would that look like e.g. in an `IconButton`?

I think it's better centered too, at least visually
image




the only situation we could have a design issue would be when multiple icons are placed on the left side? @maevalienard 
image

packages/strapi-icons/assets/icons/Play.svg
gu-stav: Using https://jakearchibald.github.io/svgomg/ and reducing the precision as long as there is no visual difference.

maevalienard:
Honestly I think it looks good as such, it's well optically centered and I'm glad we take the time to do something like this 🥰

PR #892, opened 2023-02-07
Chore: Update svgr to v6, fix width/ height attributes

gu-stav:
### What does it do?

- Updates `@svgr/cli` to v6
- Move cli args into a config file
- Disable `icons` option

### Why is it needed?

- Keep dependencies up-to-date
- I'd argue the config file is easier to read and maintain

### Options

https://react-svgr.com/docs/options/

- `dimensions: true`: Keeps the viewBox attribute
- `icon: 1rem` Makes sure the default icon size is 1rem instead of 1em
- `typescript`: Generates .tsx files with [TypeScript](https://www.typescriptlang.org/) typings.


packages/strapi-icons/package.json
```diff
@@ -29,8 +29,9 @@
     "react-dom": "^17.0.2"
   },
   "scripts": {
-    "build": "yarn generate:icons && vite build && yarn generate:types",
-    "generate:icons": "svgr --typescript --icon -d ./src ./assets/icons",
+    "build": "yarn generate:icons && yarn generate:vite:build && yarn generate:types",
+    "generate:vite:build": "vite build",
```
joshuaellis, line 33:
```suggestion
    "build:vite": "vite build",
```

IMO not necessary since its such a short command, but it's definitely not a generate script 😅 

or just `build:prod` which is what i do sometimes...

packages/strapi-icons/package.json
```diff
@@ -29,8 +29,9 @@
...
+    "generate:icons": "svgr -- ./assets/icons",
```
joshuaellis, line 34: can't have the input in the config? :(

joshuaellis (changes requested):

packages/strapi-icons/package.json
gu-stav: I went for `build:prod` because that is how it is setup in the DS package.

packages/strapi-icons/package.json
gu-stav: I don't think there is.

joshuaellis (approved):

ronronscelestes (approved):

PR #897, opened 2023-02-10
Refactor Select and Combobox

### LINKED ISSUE #634: Refactor Select / Combobox - proof of concept
Currently the `Select` and `Combobox` components are very complex and complicated to understand. It is hard to extend them, because the way events are handled get in the way.

This leads to a problem, that we still have `react-select` in the codebase of `@strapi/strapi`. In the content-squad we will need to extend both components in the near future. Before-hand we'd like to explore, if we can make both components easier to maintain, by using a framework like: https://github.com/downshift-js/downshift or https://www.radix-ui.com/docs/primitives/components/select .

joshuaellis:
### What does it do?

* Adds new primitive package
* Refactors Select to use RadixUI
* Adds Combobox Primitive
* Refactors Combobox to use new primitive

### How to test it?

* Automated testing

### Related issue(s)/PR(s)

* resolves #634


gu-stav (approved):
Since this PR does not contain any new changes that haven't been reviewed + tested in isolated PR I'm just going to approve it.

We've discussed in Slack Josh will create a beta / experimental release for the 1.7 branch because there are more changes to test.

PR #898, opened 2023-02-14
[fix] IconButton no border

ronronscelestes:
## What

fixing no border application on `IconButton` 

| Before | After |
|-|-|
| Screenshot 2023-02-14 at 10 05 15  | Screenshot 2023-02-14 at 10 04 29 |

### DIFF
--- a/packages/strapi-design-system/src/IconButton/IconButton.tsx
+++ b/packages/strapi-design-system/src/IconButton/IconButton.tsx
@@ -54,7 +54,7 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(
       <IconButtonWrapper
         aria-disabled={disabled}
         background={disabled ? 'neutral150' : undefined}
-        borderWidth={noBorder ? '0px' : undefined}
+        borderWidth={noBorder ? 0 : undefined}
         justifyContent="center"
         height={`${32 / 16}rem`}
         width={`${32 / 16}rem`}
--- a/packages/strapi-design-system/src/IconButton/IconButton.tsx
+++ b/packages/strapi-design-system/src/IconButton/IconButton.tsx
@@ -54,7 +54,7 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(
       <IconButtonWrapper
         aria-disabled={disabled}
         background={disabled ? 'neutral150' : undefined}
-        borderWidth={noBorder ? 0 : undefined}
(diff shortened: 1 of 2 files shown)


packages/strapi-design-system/src/IconButton/IconButton.tsx
```diff
@@ -54,7 +54,7 @@ export const IconButton = React.forwardRef<HTMLButtonElement, IconButtonProps>(
       <IconButtonWrapper
         aria-disabled={disabled}
         background={disabled ? 'neutral150' : undefined}
-        borderWidth={noBorder ? 0 : undefined}
+        borderWidth={noBorder ? '0px' : undefined}
```
joshuaellis, line 57: do you know why we have to pass a string? is it because we're just checking for a false value?

joshuaellis (changes requested):

markkaylor (commented):
Looks good to me now in the CMS, thanks for the quick fix.

packages/strapi-design-system/src/IconButton/IconButton.tsx
ronronscelestes: after checking again I think it comes from the `BaseButton`, checking `borderWidth` as a falsy value
```js
if (borderColor && !borderStyle && !borderWidth) {
  return `1px solid ${theme.colors[borderColor]}`;
}
```

joshuaellis: What do you think about changing that to be 

```js
if (borderColor && !borderStyle && typeof borderWidth === "undefined") {
  return `1px solid ${theme.colors[borderColor]}`;
}
```

Which _should_ also solve the issue for us and therefore allow us to pass numerical values like `0`?

ronronscelestes: sounds better indeed

ronronscelestes:
image

@joshuaellis any idea why `LinkButton` would be so unhappy (ran the failing tests 4-5times) - didn't make any change on this component here and `yarn lint` is doing well locally 🧐

joshuaellis:
> image
> 
> @joshuaellis any idea why `LinkButton` would be so unhappy (ran the failing tests 4-5times) - didn't make any change on this component here and `yarn lint` is doing well locally 🧐

Not sure either because that file doesn't exist... i've rebased the PR so the field is now included 🤞🏼

PR #899, opened 2023-02-14
fix(Layout/HeaderLayout): Don't set height to 0

Torsten85:
### What does it do?

Avoids setting header layout container height to 0px by using undefined (so no height inline style) as fallback

### Why is it needed?

When the container renders while `isVisible` is `false`,  the div will get a height of `0px`.
When `isVisible` turns `true`, the inner `<BaseHeaderLayout />` renders, but the container cannot grow,
resulting in a missing header and a broken layout.

This mostly occures when a custom field gets rendered and needs some time to load asynchronous 

### How to test it?

* Create a customField that needs some time to load (in my case, a monaco editor)
* Go on the create page
* Keep hitting "Reload" while the dev tools are open and "disable cache" is checked

### DIFF
--- a/packages/strapi-design-system/src/Layout/HeaderLayout.js
+++ b/packages/strapi-design-system/src/Layout/HeaderLayout.js
@@ -42,7 +42,7 @@ export const HeaderLayout = (props) => {
 
   return (
     <>
-      <div style={{ height: headerSize?.height }} ref={containerRef}>
+      <div style={{ height: headerSize?.height || undefined }} ref={containerRef}>
         {isVisible && <BaseHeaderLayout ref={baseHeaderLayoutRef} {...props} />}
       </div>
       {!isVisible && <BaseHeaderLayout {...props} sticky width={headerSize?.width} />}


joshuaellis (changes requested):
I'm not entirely sure this solves the problem you're facing – would a better solution not be changing this effect:

```ts
useEffect(() => {
    if (baseHeaderLayoutRef.current) {
      setHeaderSize(baseHeaderLayoutRef.current.getBoundingClientRect());
    }
  }, [baseHeaderLayoutRef]);
```

to be

```ts
useEffect(() => {
    if (baseHeaderLayoutRef.current) {
      setHeaderSize(baseHeaderLayoutRef.current.getBoundingClientRect());
    }
  }, [isVisible]);
```

So that _if_ the ref is defined, we update the size when `isVisible` changes?

Torsten85:
> I'm not entirely sure this solves the problem you're facing – would a better solution not be changing this effect:
> 
> ```ts
> useEffect(() => {
>     if (baseHeaderLayoutRef.current) {
>       setHeaderSize(baseHeaderLayoutRef.current.getBoundingClientRect());
>     }
>   }, [baseHeaderLayoutRef]);
> ```
> 
> to be
> 
> ```ts
> useEffect(() => {
>     if (baseHeaderLayoutRef.current) {
>       setHeaderSize(baseHeaderLayoutRef.current.getBoundingClientRect());
>     }
>   }, [isVisible]);
> ```
> 
> So that _if_ the ref is defined, we update the size when `isVisible` changes?

_if_ the ref is defined and `isVisible` changes, the `headerHeight` already be set by the `useResizeObserver`.
And once this value is set and the inline css rule is applied, the `baseHeaderLayoutRef` cannot grow beyond the `headerHeight`. So this effect would not work.

I simply avoid setting an initial `height: 0px` style.

But actually I'm not really sure why the `useHeaderSize()` hook exists in the first place. Why not use a normal css layout here?

joshuaellis:
> But actually I'm not really sure why the useHeaderSize() hook exists in the first place. Why not use a normal css layout here?

I'm not entirely sure either. @gu-stav  or @ronronscelestes do you have any historical knowledge about this component?

ronronscelestes:
> do you have any historical knowledge about this component?

from what I remember `HeaderLayout` was first not built with a sticky behavior - and the `useHeaderSize` logic was implemented once it was asked for it to appear sticky with a different design once not visible on the screen (after scrolling)

@Torsten85 I'm not entirely sure I understand what UI issue you are facing, would you be able to share a reproducible setup? 🙏

gu-stav:
@Torsten85 @joshuaellis I think this hook exists, because of the way `sticky` is implemented. In case the element becomes sticky, it get's position `fixed`. This would lead to a jump of the following content because the element is not taken out of the render tree for the size calculations. From what I understand in this case we measure the height of the header and reserve the space, to prevent the jump.

I think there are better solutions in 2022 that we could explore as a follow-up because we are running into this problem elsewhere in the CMS too.

PR #900, opened 2023-02-15
Add deprecate notice to v1 components

joshuaellis:
### What does it do?

* Adds deprecation notice in storybook about V1 components

### Why is it needed?

* We have V2 and supporting two of the same component is on overhead we don't need.

### DIFF
--- a/docs/components/DeprecationNotice.js
+++ b/docs/components/DeprecationNotice.js
@@ -1,0 +1,27 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+
+import { Flex, Typography, ThemeProvider, lightTheme } from '@strapi/design-system';
+
+export const DeprecationNotice = ({ children, href }) => (
+  <ThemeProvider theme={lightTheme}>
+    <Flex padding={5} background="danger200" justifyContent="center" marginTop={4} marginBottom={4}>
+      <Typography fontSize={4} fontWeight="bold" as="p">
+        ⛔️
+        <strong>
+          {' This component has been deprecated. Please use '}
+          <a style={{ color: 'inherit', textDecoration: 'underline' }} href={href}>
+            {children}
+          </a>
+          {' instead. '}
+        </strong>
+        ⛔️
+      </Typography>
+    </Flex>
+  </ThemeProvider>
+);
+
+DeprecationNotice.propTypes = {
+  children: PropTypes.node.isRequired,
+  href: PropTypes.string.isRequired,
+};
--- a/docs/stories/Stack.stories.mdx
+++ b/docs/stories/Stack.stories.mdx
@@ -4,13 +4,13 @@ import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs';
 
 import { Stack, Box } from '@strapi/design-system';
 
+import { DeprecationNotice } from '../components/DeprecationNotice';
+
 <Meta title="Design System/Technical Components/Stack" component={Stack} />
 
 # Stack
 
-This is the doc of the `Stack` component
-
-[View source](https://github.com/strapi/design-system/tree/main/packages/strapi-design-system/src/Stack)
+<DeprecationNotice href="/?path=/docs/design-system-technical-components-flex--base">Flex</DeprecationNotice>
 
 ## Imports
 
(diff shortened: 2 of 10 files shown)


docs/components/DeprecationNotice.js
```diff
@@ -0,0 +1,27 @@
...
+import PropTypes from 'prop-types';
+
+import { Flex, Typography, ThemeProvider, lightTheme } from '@strapi/design-system';
+
+export const DeprecationNotice = ({ children, href }) => (
+  <ThemeProvider theme={lightTheme}>
+    <Flex padding={5} background="danger200" justifyContent="center" marginTop={4} marginBottom={4}>
+      <Typography fontSize={4} fontWeight="bold" variant="beta">
+        ⛔️
+        <strong>
```
gu-stav, line 11: I think this is already bold from the Typography component?

docs/stories/Stack.stories.mdx
```diff
@@ -4,13 +4,13 @@ import { Meta, Story, Canvas, ArgsTable } from '@storybook/addon-docs';
...
+import { DeprecationNotice } from '../components/DeprecationNotice';
+
 <Meta title="Design System/Technical Components/Stack" component={Stack} />
 
 # Stack
 
-This is the doc of the `Stack` component
-
-[View source](https://github.com/strapi/design-system/tree/main/packages/strapi-design-system/src/Stack)
+<DeprecationNotice href="/?path=/docs/design-system-technical-components-flex--base">Flex</DeprecationNotice>
```
gu-stav, line 13: Should we also mention that it can be replaced by using the `gap` prop?

ronronscelestes: random thought about `gap` I had this morning, if we pass a single value to gap it'll set `row-gap` and `column-gap` with this value - if we pass an array then it's still handled as one single value but adding responsive capabilities
I don't think it can properly replace `spacing` of `Stack` right now as it doesn't allow to customize row and column values
if I'm not making a mistake there, should we consider improving that? (for `Grid` and `Flex` btw)

ronronscelestes (approved):
other than the comments from Gustav, LGTM! 🙏

docs/stories/Stack.stories.mdx
gu-stav: Doesn't gap follow the flex-direction? If set to row it will add a gap between columns and if set to column it will add a gap between the rows?

The fact that it is not capable of handling responsive values wouldn't worry me too much, because basically nothing is responsive ...

ronronscelestes: mmh for flex you're right - for Grid though it will apply both directions
not really worried about responsive on this one, as you said nothing is responsive rn 😬 just that we don't offer granularity on gap (at least for Grid)

gu-stav: I agree, but we are not deprecating `Grid`, because it is way more complex than Flex and can handle different cases :)

docs/components/DeprecationNotice.js
joshuaellis: The font-size isn't inherited, so you end up having to do:

```jsx
      <Typography fontSize={4} as="p">
        ⛔️
        <Typography fontSize={4} as="strong" fontWeight="bold">
          {' This component has been deprecated. Please use '}
          <a style={{ color: 'inherit', textDecoration: 'underline' }} href={href}>
            {children}
          </a>
          {' instead. '}
        </Typography>
        ⛔️
      </Typography>
```

Which i don't think is better? Although i will add `as="p"` in, i'm surprised that isn't the default 👀

docs/stories/Stack.stories.mdx
joshuaellis: >Should we also mention that it can be replaced by using the gap prop?

Maybe i'm being too literal, but using the `gap` prop does not change it to `flex`? We can add migration guides for each component if you'd like.

joshuaellis: > Doesn't gap follow the flex-direction? If set to row it will add a gap between columns and if set to column it will add a gap between the rows?

But you can pass two values as the CSS properties so it's something we should think about in the future:

https://codesandbox.io/s/vigilant-sanderson-c4vihv?file=/src/styles.css

PR #901, opened 2023-02-18
Adds declaration to design system

quiloos39:
In the current state of the design system, the distribution folder doesn't include the index.d.ts file. While this could be either intentional or forgotten in either case this pull request adds declaration creation.


joshuaellis:
Hey @quiloos39 thanks for this, this is an intentional move to not export them yet, we're still working on them and we don't want to do partial support for our internal projects.

PR #902, opened 2023-02-21
Reintroduce eslint for typescript

joshuaellis:
### What does it do?

* Adds eslint & prettier to docs
* Adds eslint to design-system package (which very closely mirrors the config for cloud)

### Why is it needed?

* Eslint is necessary for life.


packages/strapi-design-system/package.json
```diff
@@ -61,6 +61,7 @@
     "build": "yarn build:prod",
     "build:prod": "vite build",
     "generate:types": "tsc --noEmit false --emitDeclarationOnly --declarationDir dist",
+    "lint": "eslint . --ext .js,.jsx,.tsx,.ts --fix",
```
gu-stav, line 64: Should we be consistent with the CMS here and split `linting` from `fixing`, by setting up two tasks (e.g. for CI)?

```suggestion
    "lint": "eslint . --ext .js,.jsx,.tsx,.ts --fix",
    "format": "yarn lint -- --fix",
```

gu-stav (commented):
Well done & thank you for closing items from my todo-list. 🪀

packages/strapi-design-system/package.json
joshuaellis: i only kept fix cause it was there, to be honest i don't mind omitting it and making the developer run the fix command lol. Your suggestion seems off so I can't commit it by the way.

gu-stav: What do you mean by "off"?

joshuaellis: You didn't remove `--fix` 👀  sneaky

gu-stav: Yes, I'm guilty of editing the message afterward. Sorry 🙈

gu-stav (approved):

PR #905, opened 2023-02-28
No length limit on the name for the API token or transfer token 

simotae14:
### What does it do?

As a developer managing transfer tokens of API tokens, I want to have a limit on the maximum number of characters per tone name.

### Why is it needed?

Because we have some UI issues with long names
The activity must not be breaking change and so it must only affect the frontend validation

### How to test it?

 - As a developer with the permissions useful to create or edit a API tokens and transfer tokens
   - when I create an existing API token
     - then I want the validation on the Name field to limit the length of the field up to 100 characters
   - when I access the API tokens list
     - then I want the name trimmed as the description field
   - when I create an existing transfer token
     - then I want the validation on the Name field to limit the length of the field up to 100 characters
   - when I access the transfer tokens list
     - then I want the name trimmed as the description field
   - when I edit and save the edits on an existing token and the token has a length on the name longer than 100 characters
     - then I want the validation to highlight it and prevent the saving of the edits

### Related issue(s)/PR(s)

The ticket related to it is https://strapi-inc.atlassian.net/browse/DX-613 and the strapi Pr related to this is https://github.com/strapi/strapi/pull/15895

### DIFF
--- a/packages/strapi-design-system/src/Layout/HeaderLayout.js
+++ b/packages/strapi-design-system/src/Layout/HeaderLayout.js
@@ -58,6 +58,10 @@ const StickyBox = styled(Box)`
   width: ${(props) => props.width}px;
 `;
 
+const TitleFlex = styled(Flex)`
+  min-width: 0;
+`;
+
 export const BaseHeaderLayout = React.forwardRef(
   ({ navigationAction, primaryAction, secondaryAction, subtitle, title, sticky, width, ...props }, ref) => {
     const isSubtitleString = typeof subtitle === 'string';
@@ -101,6 +105,8 @@ export const BaseHeaderLayout = React.forwardRef(
       );
     }
 
+    const FlexElement = props?.ellipsis ? TitleFlex : Flex;
+
     return (
       <Box
         ref={ref}
@@ -113,12 +119,12 @@ export const BaseHeaderLayout = React.forwardRef(
       >
         {navigationAction ? <Box paddingBottom={2}>{navigationAction}</Box> : null}
         <Flex justifyContent="space-between">
-          <Flex>
+          <FlexElement>
             <Typography as="h1" variant="alpha" {...props}>
               {title}
             </Typography>
             {secondaryAction ? <Box paddingLeft={4}>{secondaryAction}</Box> : null}
-          </Flex>
+          </FlexElement>
           {primaryAction}
         </Flex>
         {isSubtitleString ? (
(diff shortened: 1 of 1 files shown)


packages/strapi-design-system/src/Layout/HeaderLayout.js
```diff
@@ -58,6 +58,10 @@ const StickyBox = styled(Box)`
   width: ${(props) => props.width}px;
 `;
 
+const TitleFlex = styled(Flex)`
+  min-width: 0;
+`;
+
```
gu-stav, line 64: `Flex` inherits from `Box` and it has a [`minWidth`](https://github.com/strapi/design-system/blob/6c4710e399055cb8831c50d3b4a02bfbcc57aa04/packages/strapi-design-system/src/Box/Box.tsx#L105) prop that should be used instead.

simotae14: done

packages/strapi-design-system/src/Layout/HeaderLayout.js
```diff
@@ -105,7 +101,11 @@ export const BaseHeaderLayout = React.forwardRef(
       );
     }
 
-    const FlexElement = props?.ellipsis ? TitleFlex : Flex;
+    let flexProps = {};
+
+    if (props?.ellipsis) {
+      flexProps.minWidth = 0;
+    }
```
joshuaellis, line 108: Is there a reason you opted to do it this way? As opposed to `minWidth={ ellipsis ? 0 : undefined }`?

Also do we need to conditionally add it? Does it actually cause any issues if we leave as is? 🤔 Feels very flakey to have to do it this way.

simotae14: I have added this option just to prevent any weird behavior in the other places where we use HeaderLayout but yes, I can add the prop every time and in the QA phase check every HeaderLayout in Strapi

simotae14: I did it, no regression errors, maybe, in a future a suggestion is to add the ellipsis prop even in other places where we use the HeaderLayout because can prevent some weird behaviours with long titles or subtitles

joshuaellis: I think in the future it would be nicer to use composition so you have more control over what you put in where e.g. 

```jsx
<HeaderLayout.Root>
	<HeaderLayout.Action><MyIconButton /></HeaderLayout.Action>
	<HeaderLayout.Title>Hello</HeaderLayout.Title>
	<HeaderLayout.Subtitle>there</HeaderLayout.Subtitle>
</HeaderLayout.Root>
```

joshuaellis (approved):

PR #910, opened 2023-03-06
Convert Field component to TS

joshuaellis:
### What does it do?

* converts Field to TS
* Adds a unit test suite (lmk if you think i've missed something)
* Revisits the documentation of Field

### How to test it?

* automated testing


markkaylor (approved):
looks good to me

packages/strapi-design-system/src/Field/FieldAction.tsx
```diff
@@ -0,0 +1,34 @@
+import { cloneElement, forwardRef, ReactNode } from 'react';
+
+import styled from 'styled-components';
+
+import { Flex, FlexProps } from '../Flex';
+import { VisuallyHidden } from '../VisuallyHidden';
+
+const FieldActionWrapper = styled(Flex)`
+  font-size: 1.6rem;
```
gu-stav, line 9: Why is this here?

joshuaellis: That was here before

gu-stav: Oh apologies. Missed it!

PR #915, opened 2023-03-08
Add experimental release workflow

joshuaellis:
### What does it do?

* Adds Github action for people to make their own experimental releases exactly the same as Strapi

### Why is it needed?

* Will give more autonomy to squads

### Note

@alexandrebodin what's the best way to handle the NPM token? Should I add my own or would it make more sense to add a more generic user? I'm not sure what we use for the main Strapi repo.

### DIFF
--- a/.github/workflows/experimental.yml
+++ b/.github/workflows/experimental.yml
@@ -1,0 +1,25 @@
+name: 'Experimental Releases'
+
+on:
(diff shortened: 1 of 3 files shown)


joshuaellis:
Needs a bit more work before doing!

gu-stav (approved):

PR #918, opened 2023-03-09
Remove field hint type check for string

MonsterDeveloper:
### What does it do?

Removed `typeof hint !== 'string'` check in `FieldHint.tsx`

### Why is it needed?

ReactNodes won't go through the check. This input
```javascript
<TextInput
  label="Key"
  hint={
    <>
      Do this{" "}
      <Link href="https://strapi.io/" isExternal>
        here
      </Link>
    </>
  }
  required
/>
```

wouldn't display a hint due to it not being a string.

### How to test it?

Not sure, whether additional tests are needed.

### Related issue(s)/PR(s)

—

### DIFF
--- a/packages/strapi-design-system/src/Field/FieldHint.tsx
+++ b/packages/strapi-design-system/src/Field/FieldHint.tsx
@@ -4,7 +4,7 @@ import { Typography } from '../Typography';
 export const FieldHint = () => {
   const { id, hint, error } = useField();
 
-  if (!hint || typeof hint !== 'string' || error) {
+  if (!hint || error) {
     return null;
   }
 
--- a/packages/strapi-design-system/src/Field/__tests__/Field.spec.tsx
+++ b/packages/strapi-design-system/src/Field/__tests__/Field.spec.tsx
@@ -48,6 +48,13 @@ describe('Field', () => {
       expect(getByRole('textbox')).toHaveAttribute('aria-describedby', expect.stringContaining('hint'));
     });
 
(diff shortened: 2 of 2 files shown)


joshuaellis:
I think it would be good to add a test to ensure the react node renders 😊

MonsterDeveloper:
@joshuaellis seems ok now. What do you think?

gu-stav:
Thank you for opening the PR. I think from a DS perspective this is a good idea, but we had discussions with design previously if we want to open up that hint. At the time we've decided to restrict it.

@maevalienard I'd like to have your opinion here too. The suggestion is to allow passing any component to the hint of inputs, so that e.g. plugins can render their own components instead of only plain text.

joshuaellis:
> Thank you for opening the PR. I think from a DS perspective this is a good idea, but we had discussions with design previously if we want to open up that hint. At the time we've decided to restrict it.
> 
> @maevalienard I'd like to have your opinion here too. The suggestion is to allow passing any component to the hint of inputs, so that e.g. plugins can render their own components instead of only plain text.

My opinion on the matter is while sure they can render w/e they want, more importantly they can render accurate links either using `ReactRouterDom` links and get the benefits or link out to external documentation.

maevalienard:
Sounds good to me 👌 (thanks for the ping!)

gu-stav (approved):
I'm good with that then :) Thanks for adding tests! 👏🏼

joshuaellis (approved):

PR #920, opened 2023-03-10
Add nesting to Simple Menu (v2)

joshuaellis:
### What does it do (in this order)?

* [x] Adds test suite to current `SimpleMenu v2`
* [x] Converts `SimpleMenu v2` to typescript
* [x] Refactor to use `radix-ui` underneath
* [x] Exposes the styled primitives of `Trigger` and `Content` to make your own customisable Menus (see docs)
* [x] Exposes parts to make nested menus
* [x] Adds more tests to cover new functionality

### Subtle Changes

Worked with Maeva on these:

* `MenuItem` now extends `Box` so all `BoxProps` are valid, however, due to `styled-components` limitations this won't work if you pass the `as` prop.
* Both `MenuItem` types of `link` now extends `v2/Link` so `LinkProps` are also valid e.g. `startIcon`.
* `Typography` in `MenuItem` now has `color: inherit` and `MenuItem` provides the default `neutral800`
* Individual `MenuItem` components can be `disabled`
* `MenuItem` will not have the blue outline because it doesn't need it, hovering items focusses them as such the background color is enough

### Why is it needed?

* Part of updating dynamic zones to add components above and below in situ
* More tests – safer development
* Radix is good.

### How to test it?

* Automated testing


joshuaellis:
@mazzucchelli for your `max-height` issue I'm thinking we could expose a composable API along side the current API so you would do:

```jsx
import { SimpleMenuRoot, SimpleMenuTrigger, SimpleMenuContent } from "@strapi/design-system/v2"

const Component = () => {
 return (
  <SimpleMenuRoot>
   <SimpleMenuTrigger>Notifications</SimpleMenuTrigger>
    <SimpleMenuContent maxHeight={320}>{/* ... */}</SimpleMenuContent>
  </SimpleMenuRoot>
 )
}
```

as opposed to importing `SimpleMenu`, IMO it's the direction I prefer for all components anyway... wdyt?

joshuaellis:
Need to refactor test nuances & add a couple around the submenu, but the code imo is good to review.

packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+  svg > path,
+  &:focus-visible svg > path {
+    fill: currentColor;
+  }
+
+  ${getOptionStyle}
+`;
+
+const IconWrapper = styled.span`
```
gu-stav, line 227: What do you think about making this Flex?

packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
```diff
@@ -0,0 +1,333 @@
...
+  display: flex;
+  align-items: center;
+
+  svg {
+    height: 4px;
+    width: 6px;
```
gu-stav, line 233: px -> rem

packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
```diff
@@ -0,0 +1,333 @@
...
+
+const MenuLabel = forwardRef<HTMLSpanElement, LabelProps>((props, ref) => (
+  <DropdownMenu.Label asChild>
+    <StyledLabel ref={ref} variant="sigma" {...props} />
+  </DropdownMenu.Label>
+));
+
+const StyledLabel = styled(Typography)`
+  padding: ${({ theme }) => theme.spaces[2]} ${({ theme }) => theme.spaces[4]};
+  color: ${({ theme }) => theme.colors.neutral600};
```
gu-stav, line 251: Why don't you use it as a prop?

gu-stav (commented):
Cool stuff!

packages/strapi-design-system/src/BaseLink/BaseLink.tsx
```diff
@@ -18,6 +20,7 @@ export const BaseLink = React.forwardRef<HTMLAnchorElement, BaseLinkProps>(
   ({ href, rel = 'noreferrer noopener', target = '_self', disabled = false, isExternal = false, ...props }, ref) => {
     return (
       <A
+        as="a"
```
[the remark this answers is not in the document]
joshuaellis: Yeah it'll overwrite this

packages/strapi-design-system/src/v2/Link/Link.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import { ExternalLink } from '@strapi/icons';
+import styled from 'styled-components';
+
+import { BaseLink, BaseLinkProps } from '../../BaseLink';
+import { Box } from '../../Box';
+import { buttonFocusStyle } from '../../themes/utils';
+import { Typography } from '../../Typography';
+
+const LinkWrapper = styled(BaseLink)`
```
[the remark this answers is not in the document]
joshuaellis: another time 😛

packages/strapi-design-system/src/v2/Link/Link.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  &:active {
+    color: ${({ theme }) => theme.colors.primary700};
+  }
+
+  ${buttonFocusStyle};
+`;
+
+const IconWrapper = styled(Box)`
+  display: flex;
```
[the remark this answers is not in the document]
joshuaellis: why not just get rid of it?

maevalienard (approved):
🤩 wow!

packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
```diff
@@ -0,0 +1,316 @@
...
+      </SubmenuTrigger>
+    </DropdownMenu.SubTrigger>
+  );
+});
+
+const SubmenuTrigger = styled(OptionButton)`
+  display: flex;
+  align-items: center;
+  justify-content: flex-start;
+  gap: ${({ theme }) => theme.spaces[5]};
```
gu-stav, line 255: Since `OptionButton` is a Box, all of these can become props.

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.tsx
```diff
@@ -0,0 +1,84 @@
...
+      onReachEnd(entry);
+    }
+  };
+
+  const handleOpenChange = (isOpen: boolean) => {
+    if (isOpen) {
+      onOpen();
+    } else {
+      onClose();
+    }
```
gu-stav, line 52: Should we do a typecheck here because both props are optional?

yarn.lock
```diff
@@ -1310,18 +1310,38 @@
...
+    "@floating-ui/core" "^0.7.3"
+
 "@floating-ui/dom@^1.2.1":
   version "1.2.1"
   resolved "https://registry.yarnpkg.com/@floating-ui/dom/-/dom-1.2.1.tgz#8f93906e1a3b9f606ce78afb058e874344dcbe07"
   integrity sha512-Rt45SmRiV8eU+xXSB9t0uMYiQ/ZWGE/jumse2o3i5RGlyvcbqOF4q+1qBnzLE2kZ5JGhq0iMkcGXUKbFe7MpTA==
   dependencies:
     "@floating-ui/core" "^1.2.1"
 
+"@floating-ui/react-dom@0.7.2":
```
gu-stav, line 1337: Could we potentially run into a conflict with the other 1.x version?

packages/strapi-design-system/src/v2/SimpleMenu/SimpleMenu.tsx
joshuaellis: no because there are `noop` defaults declared.

gu-stav: Kind of: you recently commented on one of my PRs that onClose could also be something different, e.g. `null` and in this case it would throw. I think you had a good point there it is a good idea to wrap functions always in a type-check.

joshuaellis: aha great point, to be honest i prefer stricter checking than defaults anyway 😛

yarn.lock
joshuaellis: I'd assume not as it's not unusual to have different deps versions, its just the strapi app that handles deps in such a peculiar manor.

gu-stav (approved):
👏🏼

PR #922, opened 2023-03-15
BaseButton: Merge background prop with falsy values

gu-stav:
### What does it do?

- Updates the `IconButton` stories to use a background-color, which is never used in the button itself to make the background-color of the button itself visible. Before they were the same which made me miss they are now transparent instead of white.
- Merges the `background` prop in `BaseButton`

### Why is it needed?

Components inheriting from `BaseButton` may set `background=undefined` if they don't want to set a background-color explicitly (e.g. for a ternary). Merging props with `background=undefined` leads to the original value being overwritten. This PR prevents that by checking for a fasly value.

### How to test it?

### Related issue(s)/PR(s)

- Broke in https://github.com/strapi/design-system/pull/878

### DIFF
--- a/packages/strapi-design-system/src/BaseButton/BaseButton.tsx
+++ b/packages/strapi-design-system/src/BaseButton/BaseButton.tsx
@@ -28,7 +28,7 @@ export interface BaseButtonProps<TElement extends HTMLElement = HTMLButtonElemen
 }
 
 export const BaseButton = React.forwardRef<HTMLButtonElement, BaseButtonProps>(
-  ({ disabled, children, ...props }, ref) => {
+  ({ disabled, children, background = 'neutral0', ...props }, ref) => {
     return (
       <BaseButtonWrapper
         ref={ref}
@@ -38,7 +38,7 @@ export const BaseButton = React.forwardRef<HTMLButtonElement, BaseButtonProps>(
         disabled={disabled}
         padding={2}
         hasRadius
-        background="neutral0"
+        background={background}
         borderColor="neutral200"
         cursor="pointer"
         {...props}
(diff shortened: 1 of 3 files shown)


packages/strapi-design-system/src/BaseButton/BaseButton.tsx
```diff
@@ -28,7 +28,7 @@ export interface BaseButtonProps<TElement extends HTMLElement = HTMLButtonElemen
 }
 
 export const BaseButton = React.forwardRef<HTMLButtonElement, BaseButtonProps>(
-  ({ disabled, children, ...props }, ref) => {
+  ({ disabled, children, background, ...props }, ref) => {
```
joshuaellis, line 31:
```suggestion
  ({ disabled, children, background = 'neutral0', ...props }, ref) => {
```

easier?

packages/strapi-design-system/src/BaseButton/BaseButton.tsx
```diff
@@ -38,7 +38,10 @@ export const BaseButton = React.forwardRef<HTMLButtonElement, BaseButtonProps>(
         disabled={disabled}
         padding={2}
         hasRadius
-        background="neutral0"
+        // components inheriting from BaseButton set background
+        // to "undefined", to avoid setting any explicit and duplicate
+        // fallback color
+        background={background ?? 'neutral0'}
```
joshuaellis, line 44:
```suggestion
        background={background}
```

joshuaellis (commented):
what do you think about that?

joshuaellis (approved):
✌🏼

PR #923, opened 2023-03-15
TextButton: Cleanup component

gu-stav:
### What does it do?

- Replaces the inline-paddings with gap
- Splits out a separate icon story
- Simplify markup

### Why is it needed?

The component was a bit messy.

### DIFF
--- a/packages/strapi-design-system/src/TextButton/TextButton.tsx
+++ b/packages/strapi-design-system/src/TextButton/TextButton.tsx
@@ -66,10 +66,11 @@ export const TextButton = React.forwardRef<HTMLButtonElement, TextButtonProps>(
         onClick={handleClick}
         as="button"
         type="button"
+        gap={2}
         {...props}
       >
         {(startIcon || loading) && (
-          <Box as="span" paddingRight={2} aria-hidden>
+          <Box as="span" aria-hidden>
             {loading ? (
               <LoadingWrapper>
                 <Loader />
@@ -79,11 +80,13 @@ export const TextButton = React.forwardRef<HTMLButtonElement, TextButtonProps>(
             )}
           </Box>
         )}
+
         <Typography variant="pi" textColor={isDisabled ? 'neutral600' : 'primary600'}>
           {children}
         </Typography>
+
         {endIcon && (
-          <Box as="span" paddingLeft={2} aria-hidden>
+          <Box as="span" aria-hidden>
             {endIcon}
           </Box>
         )}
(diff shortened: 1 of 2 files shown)


joshuaellis:
How come you think a default of null is more useful that undefined? 🤔

packages/strapi-design-system/src/TextButton/TextButton.tsx
```diff
@@ -66,27 +59,23 @@ export const TextButton = React.forwardRef<HTMLButtonElement, TextButtonProps>(
...
-            {loading ? (
-              <LoadingWrapper>
-                <Loader />
-              </LoadingWrapper>
-            ) : (
-              startIcon
-            )}
-          </Box>
+        {loading ? (
+          <LoadingWrapper aria-hidden>
```
gu-stav, line 67: Actually: I don't think this should have `aria-hidden`. Wdyt @joshuaellis ?

gu-stav:
> How come you think a default of null is more useful that undefined? 🤔

Sorry that was a leftover - it's gone now.

packages/strapi-design-system/src/TextButton/TextButton.tsx
joshuaellis: Interesting question, is it just an icon? If so then I think it's fine, but if it supposed to tell people "this is loading" then probably not, but I don't know if adding to its label is correct either...

gu-stav: I think if this component is in a loading state, which is only indicated by the icon, it shouldn't be hidden. On the other hand `Loader` defines `role="alert" aria-live="assertive"` - I could imagine these don't relate to whether the element is hidden or nor and fire anyway, which would then inform the user. However: if this is the case why define it as hidden at all?

joshuaellis: > I think if this component is in a loading state, which is only indicated by the icon, it shouldn't be hidden

But if it's just an SVG it doesn't mean anything to a screen-reader without a label you can see that [here](https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-textbutton--loading)

Also it only uses the icon `Loader` not the component so there's no aria role // announcement which i think is okay, in reality you'd expect the label for button to change, `save` -> `saving`. We could add `aria-busy`? but im not sure on that one either....

This issue looks like it could be useful to read through? https://github.com/Shopify/polaris/issues/1212

gu-stav: Let's leave it as it is and create a follow-up for this. I could imagine this also applies to all other buttons and it might make sense to work out what we want globally. Would that work for you?

joshuaellis: Yeah i think is a topic that could benefit from some research 🔍

joshuaellis (approved):

PR #929, opened 2023-03-22
Add Dependabot

joshuaellis:
### What does it do?

* Adds dependabot yaml

### Why is it needed?

* Even though we don't have loads of dependencies, it's nice to keep our dev deps up to date!


gu-stav (approved):
💅🏼

PR #937, opened 2023-03-22
Fix/add ts declarations

csechrist:
### What does it do?

This PR builds the design-system and icons ts declaration files into a separate folder and adds them to the package.json

### Why is it needed?

Whenever creating a new strapi plugin, I am having to work around the missing types by declaring the module, removing the benefits of using typescript for a consumer of the package

### How to test it?

You can verify that the ./types folder exists after running `yarn build` within the repository

### Related issue(s)/PR(s)

Related to #633

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ test-results
 playwright-report
 packages/strapi-icons/src
 coverage
+types
--- a/packages/strapi-design-system/.eslintignore
+++ b/packages/strapi-design-system/.eslintignore
@@ -1,2 +1,3 @@
 **/node_modules/**
 **/dist/**
+**/types/**
--- a/packages/strapi-design-system/package.json
+++ b/packages/strapi-design-system/package.json
@@ -62,7 +62,7 @@
     "develop": "vite build --watch",
     "build": "yarn build:prod && yarn generate:types",
     "build:prod": "vite build",
-    "clean": "rm -rf dist node_modules",
+    "clean": "rm -rf dist node_modules types",
     "generate:types": "tsc --noEmit false --emitDeclarationOnly --declaration --declarationDir types",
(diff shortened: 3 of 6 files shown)


csechrist:
I am not sure this is the right branch, if not let me know and I apologize about that!

joshuaellis:
Hey thanks for this I believe the icons package already publishes it's types. With the rest of the design system I don't think we're quite there to be publishing them as they've not been testing in our cloud repo.

Last time I tested publishing partial types (not every component has been converted to TS yet) it also still threw errors because certain components had no types and therefore it assumed all the props were required 🤔

csechrist:
> Hey thanks for this I believe the icons package already publishes it's types. With the rest of the design system I don't think we're quite there to be publishing them as they've not been testing in our cloud repo.
> 
> Last time I tested publishing partial types (not every component has been converted to TS yet) it also still threw errors because certain components had no types and therefore it assumed all the props were required 🤔

Ah ok @joshuaellis  -- is there anything that I can do to help get it to that point?  Types would be a super helpful thing for us!

joshuaellis:
> > Hey thanks for this I believe the icons package already publishes it's types. With the rest of the design system I don't think we're quite there to be publishing them as they've not been testing in our cloud repo.
> > Last time I tested publishing partial types (not every component has been converted to TS yet) it also still threw errors because certain components had no types and therefore it assumed all the props were required 🤔
> 
> Ah ok @joshuaellis -- is there anything that I can do to help get it to that point? Types would be a super helpful thing for us!

You're more than welcome to convert our components to TS! I would either group a couple together that are linked e.g. a TextButton could rely on Button, so it makes sense to change them at the same time or just do one at a time if they don't have unconverted dependent components.

When we have full TS coverage of all components we'll them be looking to do experimental releases of the DS to test on our cloud repo :)

PR #945, opened 2023-03-29
Fix decimal and number i18n formatting issue

marob:
### What does it do?

Fix https://github.com/strapi/strapi/issues/15965

### Why is it needed?

Without that fix, decimal numbers are first displayed without locale formatting (ie: `123456.789`), which:
- is not consistent with auto-formatting when editing the field (being `123,456.789` in English or `123 456,789` in French, for example)
- creates an issue if field is clicked+blurred in some languages: in French, `123456.789` is transformed in `123 456 789`

### How to test it?

Just use some decimal/float type.

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/15965


package.json
```diff
@@ -43,7 +43,7 @@
     "chokidar-cli": "^3.0.0",
     "cross-env": "^7.0.3",
     "eslint": "^8.36.0",
-    "fs-extra": "^10.0.0",
+    "fs-extra": "^11.1.1",
```
joshuaellis, line 46: Can you revert the dep changes please? While it's useful, I think we should do it in a separate PR :)

marob: Fixed. It was due to the fact that I created the MR from main branch instead of release one, then targeted the release branch.

joshuaellis (approved):

simotae14 (approved):

PR #948, opened 2023-04-06
Improve DS Website Colors page

udimberto:
### What does it do?

Transforming, from images to HTML/selectable content, the Colors page of Strapi DS Website.

### Why is it needed?

Will help the community to select/use/copy the Strapi DS color tokens more easier.

### DIFF
--- a/website/components/MarkdownComponents/ColorShades.jsx
+++ b/website/components/MarkdownComponents/ColorShades.jsx
@@ -40,11 +40,11 @@ const ColorCardInfoContrast = ({ backgroundColor = '', isLighter = false, isSmal
         color="neutral0"
         padding="4px 2px"
         textAlign="center"
-        style={{ fontSize: '12px' }}
+        style={{ fontSize: '12px', textTransform: 'uppercase' }}
       >
         {tinycolor2.isReadable(textColor, backgroundColor, { level: 'AAA', size: isSmall ? 'small' : 'large' })
-          ? 'PASS'
-          : 'FAIL'}
+          ? 'Pass'
+          : 'Fail'}
       </Box>
     </Flex>
   );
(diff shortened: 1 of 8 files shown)


Eventyret:
LGTM @udimberto 🥳

joshuaellis (dismissed):
Nice changes all around @udimberto <3

gu-stav (commented):
That looks very good indeed! Thanks!


What do you think about improving the accessibility of these cards before we merge it? Right now there are only `div`s which makes them very much not accessible.

Screenshot 2023-04-11 at 12 53 05

Markup-wise I could think of:

- each card is an `article`
- each `article` has a headline with the name of the color (hidden)
- all other key-value pairs are a `dl/dt` structure
- the font-size icons could get a hidden aria label indicating "Large font" / "Small font" for example and be a `dl/dt` too?

HichamELBSI (approved):

udimberto:
> 

I think that's great, @gu-stav 😄.

Awesome suggestions, thank you so much.
I've already applied them (_not very sure about what I did_ 😅).

Would you mind to give a second look, when you have a free time, please?

udimberto:
> That looks very good indeed! Thanks!
> 
> What do you think about improving the accessibility of these cards before we merge it? Right now there are only `div`s which makes them very much not accessible.
> 
> Screenshot 2023-04-11 at 12 53 05
> 
> Markup-wise I could think of:
> 
> * each card is an `article`
> * each `article` has a headline with the name of the color (hidden)
> * all other key-value pairs are a `dl/dt` structure
> * the font-size icons could get a hidden aria label indicating "Large font" / "Small font" for example and be a `dl/dt` too?

I think that's great, @gu-stav 😄.

Awesome suggestions, thank you so much.
I've already applied them (_not very sure if I've applied them correctly_ 😅).

Would you mind to give a second look, when you have a free time, please?

website/components/MarkdownComponents/ColorShades.jsx
```diff
@@ -0,0 +1,154 @@
...
+  label: PropTypes.node,
+};
+
+const ColorCardInfoContrast = ({ backgroundColor = '', isLighter = false, isSmall = false }) => {
+  const textColor = isLighter ? '#FFF' : '#000';
+
+  return (
+    <Box textAlign="center">
+      <Box
+        as="dd"
```
joshuaellis, line 29:
```suggestion
        as="dt"
```
?

joshuaellis (changes requested):

website/components/MarkdownComponents/ColorShades.jsx
```diff
@@ -0,0 +1,154 @@
...
+        padding={4}
+        minHeight="104px"
+      >
+        <Grid
+          as="dl"
+          aria-label={`Contrast accessibility checks for ${colorName} ${colorShade}`}
+          gap={2}
+          gridCols={4}
+          style={{ alignItems: 'end' }}
+        >
```
joshuaellis, line 88: I think because you're not doing a grid here you're just doing a row, we should use `Flex`, it'll reduce the HTML too and get rid of `GridItem` 👍🏼

udimberto: Yes, it worked.
Changes applied.
Thank you for that.
🤝

gu-stav (approved):
Great work, @udimberto. I think this is a very good start!

I've tried using the page using voiceover today and could think of a few improvements, but would leave it up to you if you want to implement those:

- The colors at the top are focusable (probably because they have a tooltip). I was wondering if it would make sense to link them to the headlines below (with an anchor) or remove them at all? I don't see how they improve the information on the page.

Screenshot 2023-04-14 at 08 42 52

- You could wrap each `article` in an ordered-list item, since the color shades are a list of things. This would allow screen-reader users more easily to discover how many shades each color contains and to skip over them. I don't think it is major because the page is already grouped by headlines and screenreader users could skip to the next one.

- The labels "FAIL" and "PASS" are currently written in uppercase in the markup. I think it would be better to use `text-transform: uppercase` for it.

Again: good work! 🚀

joshuaellis (approved):

Bassel17 (approved):
ohhh that's a great improvement love this !

PR #953, opened 2023-04-13
create the useInternationalizedNumber hook

simotae14:
### What does it do?

Add a new hook `useInternationalizedNumber` to create a number formatter and a number parser using the `@internationalized/number` library.
And also we have replaced all the code with the new hook inside the NumberInput component

### Why is it needed?

Becuase it is useful if we want in future to use again the library, we need just to add the hook to configure the library, and also can be useful in the CMS where we have to show the number with a specific locale format

### How to test it?

Just take a look at the NumberInput component

### DIFF
--- a/packages/strapi-design-system/src/hooks/useInternationalizedNumber.ts
+++ b/packages/strapi-design-system/src/hooks/useInternationalizedNumber.ts
@@ -1,0 +1,17 @@
+import { useRef } from 'react';
+
+import { NumberFormatter, NumberParser } from '@internationalized/number';
+
+import { getDefaultLocale } from '../helpers/getDefaultLocale';
+
+export const useInternationalizedNumber = (locale, { parserParams, formatParams }) => {
+  // define the locale or use the default one
+  const finalLocale = locale || getDefaultLocale();
+  const numberParserRef = useRef(new NumberParser(finalLocale, parserParams));
+  const numberFormaterRef = useRef(new NumberFormatter(finalLocale, formatParams));
+
+  return {
+    parser: numberParserRef.current,
+    formatter: numberFormaterRef.current,
+  };
+};
--- a/packages/strapi-design-system/src/NumberInput/NumberInput.js
+++ b/packages/strapi-design-system/src/NumberInput/NumberInput.js
@@ -1,17 +1,16 @@
 /* eslint-disable no-restricted-globals */
-import React, { useRef } from 'react';
+import React from 'react';
 
-import { NumberFormatter, NumberParser } from '@internationalized/number';
 import { CarretDown } from '@strapi/icons';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
 import { Field, FieldLabel, FieldHint, FieldError, FieldInput } from '../Field';
 import { Flex } from '../Flex';
-import { getDefaultLocale } from '../helpers/getDefaultLocale';
 import { KeyboardKeys } from '../helpers/keyboardKeys';
 import { useControllableState } from '../hooks/useControllableState';
 import { useId } from '../hooks/useId';
+import { useInternationalizedNumber } from '../hooks/useInternationalizedNumber';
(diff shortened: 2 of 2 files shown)


packages/strapi-design-system/src/hooks/useInternationalizedNumber.ts
```diff
@@ -0,0 +1,17 @@
+import { useRef } from 'react';
+
+import { NumberFormatter, NumberParser } from '@internationalized/number';
+
+import { getDefaultLocale } from '../helpers/getDefaultLocale';
+
+export const useInternationalizedNumber = (locale, { parserParams, formatParams }) => {
```
joshuaellis, line 7: There's no type annotation for what `locale` or the parms are, it'd also be good to add JSdoc to explain what this is used for etc.

joshuaellis (changes requested):
Implementation looks good, we're not exporting it from `@strapi/design-system` though so no one will be able to use it, since we're exporting it i'd expect a storybook documentation page too :)

gu-stav (commented):
Good idea! :) Please add some tests.

Overall I am not sure if we should share this hook from the design-system. It seems more of a coincidence to me we already have everything in place in terms of dependencies. I'm find with using it internally, but would share the hook itself from the helper-plugin in case you need it in the CMS.

What do you think?

joshuaellis:
> would share the hook itself from the helper-plugin in case you need it in the CMS

I'd disagree with this, it feels like we'd just be re-writing the hook for no reason? 🤔

gu-stav:
> I'd disagree with this, it feels like we'd just be re-writing the hook for no reason? 🤔

Well the reason I see is that the design-system is focused on components/ tools that provide building blocks for the UI. I don't see the library as a utility package and if we'd start to share a hook that does number formatting I couldn't say how to differentiate the helper-plugin from the design-system.

I don't think it is a big issue but I think we should be careful with this. This is mostly opinion, but I think the NumberInput in general does too much and is too opinionated for a design-system already (when it comes to parsing) and I'd be careful to share the tools from the design-system rather than moving the complexity out into e.g. the helper-plugin.

simotae14:
@gu-stav and @joshuaellis I understand both your points, I think it will be useful to have this new hook but maybe it is better to have it in the helper-plugin. I have just one question, if we expose the hook inside the helper plugin we need to install `@internationalized/number` on it and then `strapi/helper-plugin` in the design-system, are you ok with it?

gu-stav:
@simotae14 The helper-plugin can not be a dependency of the design-system, because it is very CMS specific. I'd propose to duplicate the hook in the design-system and the helper-plugin - it is not a big chunk of work. The hook can be used internally in the design-system. I don't see a problem in installing the dependency in the helper-plugin.

joshuaellis:
> I'd propose to duplicate the hook in the design-system and the helper-plugin - it is not a big chunk of work

What would be the difference between duplicating it and re-exporting it? 🤔 the design-system is already a dependency of the helper-plugin so it's not a big deal & this hook doesn't impart defaults.

gu-stav:
If we start to export this hook from the design-system, we will have to support this until we decide to break it. We can create a hook for it and use internally in the design-system. We can then duplicate the hook and share it from the helper-plugin. We can discuss this in a sync meeting if you want, but I don't see the point of sharing a number formatting utility function from the design-system.

joshuaellis:
> If we start to export this hook from the design-system, we will have to support this until we decide to break it. We can create a hook for it and use internally in the design-system. We can then duplicate the hook and share it from the helper-plugin. We can discuss this in a sync meeting if you want, but I don't see the point of sharing a number formatting utility function from the design-system.

Lets discuss in the next sync because i think there a few scenarios this could apply too and we should all agree on the approach with pros/cons 👍🏼

simotae14:
ok, in the meantime I can just try to implement correctly the code and add a test, in the sync meeting we can decide where to place the hook...thanks guys for your comments

PR #954, opened 2023-04-13
Fixes issue with date picker icon not being visible in dark mode

PritamSangani:
### What does it do?

Fixes issue with date picker icon not being visible in dark mode

### Why is it needed?

The icon for date picker was only visible in light mode

### How to test it?

1. Go to storybook for design system and go to DatePicker or DateTimePicker components. 
2. Toggle between light and dark mode
3. The icon on date picker should be visible on both modes and should be the same colour as the icon for time picker

### Related issue(s)/PR(s)

Haven't created an issue for this. Let me know if I should create one


joshuaellis (changes requested):

joshuaellis (approved):
Nice catch

PR #957, opened 2023-04-19
refactor multi-select to use primitives

joshuaellis:
### What does it do?

* Adds fork of radix/select
* Adds more tests for the select component in the DS
* Refactors the multi-select options to use the primitive

### Why is it needed?

* RadixUI does not support `multi` on the `Select` component so we needed to implement it ourselves, also because a select can have tags there were some event changes we had to do to support this, all in all, we needed to split off from the library on this one.

### How to test it?

* automated testing


packages/strapi-design-system/src/TimePicker/TimePicker.tsx
```diff
@@ -53,7 +52,7 @@ export const TimePicker = ({ id, value, step = 15, onChange, ...props }: TimePic
     return `${hours}:${minutes}`;
   };
 
-  const handleChange = (value: string | number) => {
+  const handleChange = (value: string | number | string[]) => {
```
gu-stav, line 55: TIL: `Array.toString`

gu-stav (commented):
- [x] Clicking the label does not focus the component (might just be missing mapping between both in the example): https://design-system-git-feat-multi-select-strapijs.vercel.app/?path=/story/design-system-components-select--multi
- [ ] Clear does not work (might also just be an incomplete example)
- [ ] Hitting "Enter" on a checkbox also toggles the element: is that what is wanted according to the spec?
- [x] Previously it was not possible to tab through the tags (multi with tags): when you select tags, tab out of the component and in again each tab is not tabbable (with a broken focus outline). Previously I was not able to do so. Do you know what is expected? I guess it makes sense to be able to remove each tag individually, but hitting "Enter" on a Tag does not work and I am not sure if this would introduce additional complexity. (this might also be a bug on the Tag component - in this case should we just create a follow-up ticket?)
- [ ] Multi-Nested: 1) Select a group 2) Deselect one child 3) Go back and toggle the group does now toggle each children individually instead of either to on of off
- [x] In the [primitive example](https://design-system-git-feat-multi-select-strapijs.vercel.app/?path=/story/design-system-primitives-select--multiple-items) once my focus gets to the last item, the popover is pinned to the bottom of the screen (which might be totally ok - just find it weird from a docs pov)
- [x] in the [size s](https://design-system-git-feat-multi-select-strapijs.vercel.app/?path=/story/design-system-components-select--size-s) example: one I select a values it displays "9 values selected"
- [ ] Multi nested: if nothing is selected the sub-items break into two lines; once selected they are in one again. Is that a CSS problem or is that because the Select does not have enough width? Before that did not happen because the value always was "0 items selected" https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-select--multi-nested

## Radix fork

~I think it would be good to highlight (e.g. via comments) which are the parts that you had to change, so that we have a chance one day to get rid of the fork again once implemented upstream. You could also link the open issue for that. What do you think?~

I see on slack you already did this.

packages/strapi-design-system/src/TimePicker/TimePicker.tsx
joshuaellis: 😎

docs/stories/primitives/Select.stories.mdx
```diff
@@ -0,0 +1,239 @@
...
+- `Select.TextInput`
+- `Select.Icon`
+- `Select.Portal`
+- `Select.Content`
+- `Select.Viewport`
+- `Select.Item`
+- `Select.ItemText`
+- `Select.ItemIndicator`
+- `Select.NoValueFound`
+- `Select.CreateItem`
```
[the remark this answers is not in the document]
joshuaellis: Yeah maybe, I want to write more documentation in a follow up PR as well, it'd be the last checkbox on the list.

packages/strapi-design-system/src/hooks/useComposeRefs.ts
```diff
@@ -0,0 +1,50 @@
...
+ *  const ref = useComposedRefs(internalRef, forwardedRef);
+ *
+ *  React.useEffect(() => {
+ *   ref.current.focus();
+ *  }, [ref]);
+ *
+ *  return <input ref={ref} />
+ * }
+ * ```
+ */
```
joshuaellis, line 44: @gu-stav let me know if this is what you had in mind 👍🏼

joshuaellis:
>  Clear does not work (might also just be an incomplete example)

Can you elaborate on this one please? It works for me, the only scenario I can see it doing this is if the popover is open, but this is to be expected because of the `FocusTrap` that's used – which you can see the same happening [here](https://design-system-f6ly9oivw-strapijs.vercel.app/?path=/story/design-system-components-select--base) which is from the main feat branch.

>  Hitting "Enter" on a checkbox also toggles the element: is that what is wanted according to the spec?

Can you elaborate on this one please?

>  Multi nested: if nothing is selected the sub-items break into two lines

This is CSS from the select not being wide enough. I've added no-wrap to the items but also the popover menu now use's min-width to ensure it's at least the size of the trigger :)

gu-stav:
>> Clear does not work (might also just be an incomplete example)

> Can you elaborate on this one please?

Apologies - I thought it was clear but apparently it does work using the trackpad, but not using the keyboard:

https://user-images.githubusercontent.com/2244375/233581357-0672698e-2162-434d-8664-88e63b5d12f6.mp4

Hitting "Enter" on the clear button does open the Popover again.

>> Hitting "Enter" on a checkbox also toggles the element: is that what is wanted according to the spec?

> Can you elaborate on this one please?

1. Select Checkbox
2. Hit "Enter"

https://user-images.githubusercontent.com/2244375/233581887-096fa7c5-c941-440c-897a-267aa6ef4cc8.mp4

I'm wondering what the expectation here is? Should it close the popover? Should it submit the form? Should it toggle the checkbox? I think on a regular checkbox it would submit the form, no?

joshuaellis:
> I'm wondering what the expectation here is? Should it close the popover? Should it submit the form? Should it toggle the checkbox? I think on a regular checkbox it would submit the form, no?

If you remove the idea it's a checkbox, because that's just an "indicator" then it's a select and hitting enter toggles the active item, in a single select this will close the popover but because it's multi it shouldn't close the popover imo. Therefore, I think this is as fine as is 👍🏼 

> clear

this is fixed 💪🏼

gu-stav (approved):

PR #958, opened 2023-04-20
Remove z-index causing navbar overlap in the CMS

markkaylor:
### What does it do?

Removes the z-index which is apparently unnecessary.

### Why is it needed?

The z-index causes the popover to overlap a navbar in the CMS
Berfore:
Screenshot 2023-04-20 at 17 44 11
After:
Screenshot 2023-04-20 at 17 37 16

### How to test it?

Run yarn build in the DS
Head to the CMS and from the root run

```
yarn link -r ../<relative-path-to-strapi-design-system>
```
In examples getstarted yarn build && yarn develop
Go to the CM or Audit Logs or the Marketplace, make sure you have enough content to be scrollable, and scroll down the page.

(I'll update the strapi docs since it seem yarn link works differently now)

### DIFF
--- a/packages/strapi-design-system/src/Popover/Popover.tsx
+++ b/packages/strapi-design-system/src/Popover/Popover.tsx
@@ -24,7 +24,6 @@ export const POPOVER_PLACEMENTS = [
 
 const PopoverWrapper = styled(Box)`
   box-shadow: ${({ theme }) => theme.shadows.filterShadow};
-  z-index: 4;
   border: 1px solid ${({ theme }) => theme.colors.neutral150};
   background: ${({ theme }) => theme.colors.neutral0};
 `;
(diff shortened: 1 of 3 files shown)


remidej (approved):
Nice improvement 👍

packages/strapi-design-system/src/Popover/Popover.tsx
```diff
@@ -24,7 +24,6 @@ export const POPOVER_PLACEMENTS = [
 
 const PopoverWrapper = styled(Box)`
   box-shadow: ${({ theme }) => theme.shadows.filterShadow};
-  z-index: 4;
```
joshuaellis, line 27: isn't this going to fail if the popover is in a modal? 🤔

joshuaellis (changes requested):

packages/strapi-design-system/src/Popover/Popover.tsx
joshuaellis: Also would it not be better to use the `zIndexes` in the theme? & update the header the same way but with a higher theme value? :)

joshuaellis:
This should have been pointed at the release branch, i've changed it now but you've got a merge conflict – probably because we don't use snapshots or the popover component anymore 👍🏼 soon as we can get that resolved I'm happy to approve this :)

Issue #955, opened 2023-04-13
[bug]: CalendarView on Date Picker does not disappear when clicking away


PritamSangani:
### What version of strapi-design-system are you using?

Design system v.1.6.6

### What's Wrong?

CalendarView on Date Picker does not disappear when clicking away. The only way to hide the date picker view, it seems, is to click back onto the field. This can lead to overlapping date picker popovers if there are 2 date pickers next to each other. This is also, in my opinion, not optimal UX, as I would expect the date picker view to be hidden when clicking away like it happens with other input fields, such as select fields.

See screenshot below of example of overlapping date pickers.

![image](https://user-images.githubusercontent.com/22857896/231846110-c7daf750-dd56-4e92-9e9b-842f96c491c0.png)

### To Reproduce

Interact with date picker on Storybook and you can see that the date picker remains on screen even when clicking away from the field.

### Expected Behaviour

The date picker view should be hidden when clicking away

PritamSangani:
One option to fix this would be to wrap the existing date picker component to use a relevant component from the Floating UI library which is already in use in the design system. An example of a similar interaction using the Floating UI can be found here: https://codesandbox.io/s/admiring-lamport-5wt3yg?file=/src/DropdownMenu.tsx

PR #967, opened 2023-04-26
close popover when clicking away

jason-liu22:
### What does it do?

I think it would be better if we can close the `Popover` when clicking away.
But it didn't work and then I made it working in that way.

https://user-images.githubusercontent.com/61251512/234651605-6835eb5d-9c70-4990-8ef2-c2e416f43349.mp4

### Related issue(s)/PR(s)
#955
This issue is not about this pull request, but it's a similar one.

I have checked #955, but `DatePicker` is different a bit, and I have found a solution.
But The solution depends on this PR.
So if this PR is approved, then I can make another PR for #955.

### DIFF
--- a/packages/strapi-design-system/src/Popover/Popover.tsx
+++ b/packages/strapi-design-system/src/Popover/Popover.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { useFloating, flip, shift, offset, autoUpdate, Placement } from '@floating-ui/react-dom';
 import styled from 'styled-components';
 
+import useClickAway from './useClickAway';
 import { Box, BoxProps } from '../Box';
 import { useIntersection } from '../hooks/useIntersection';
 import { Portal } from '../Portal';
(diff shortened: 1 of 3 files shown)


joshuaellis:
Hey, thanks for this! I actually think using the `DismissibleLayer` component would be more appropriate, it encapsulates all this logic, there's also a PR (#968) opened this AM to move it to use the Radix package as well because it handles a lot more edge cases than we originally planned – you could branch off that PR branch to rework your PR, wydt?

joshuaellis:
Currently all the `Popover` usage in Strapi doesn't render the `Popover` when it's false, adding any of the rendering logic into this component right now would be confusing and somewhat pointless. 

IMO we should wrap the `DismissibleLayer` between the `FocusTrap` and `Content` and add a new prop `onDismiss` that way we can implement the handles on the components in the CMS.

In the future when we migrate this component to `radix` we'll be exporting a `Trigger`.

joshuaellis:
This has been solved in #982

PR #968, opened 2023-04-27
Refactor DismissibleLayer & Portal

joshuaellis:
### What does it do?

* Refactors DismissibleLayer to use `radix` underneath
* Refactors portal to work in SSR envs and render immediately (no effect necessary)

### Why is it needed?

* DimissibleLayer & Portal together had some edge cases that weren't great and we had to use data attributes to account for them


packages/strapi-design-system/src/Portal/Portal.tsx
```diff
@@ -2,29 +2,16 @@ import * as React from 'react';
...
+import { Box, BoxProps } from '../Box';
 
-    setMounted(true);
-
-    return () => {
-      rootRef.current?.remove();
-    };
-  }, []);
+export interface PortalProps extends BoxProps<HTMLDivElement> {
+  container?: HTMLElement | null;
```
gu-stav, line 8: Does `null` make sense here or would it be better to allow only undefined or HTMLElement?

joshuaellis: I think it's fine to be open, the full type is `HTMLElement | undefined | null` in the sense you can omit the prop completely, so you don't have to pass `null` if you just want to use the default

gu-stav (approved):

PR #970, opened 2023-04-27
Refactor DatePickerCalendar

### LINKED ISSUE #596: Date dropdown overflow issues

- Node.js version: v16.14.0
- NPM version:
- Strapi version: v4.1.7
- Database: Sqlite
- Operating system: Ubuntu 16.04 / WSL

### Describe the bug

The width of the date selector dropdowns for changing month / year in the admin panel does not take scroll bar width into account, and content is this hidden behind the scroll bar. This is particularly problematic when the last digit of the year is cut off.

This seems to be isolated to Firefox (98.0.2), Chromium-based browsers seem to work fine.

1. Create a model with date field
2. Try and change the month/year
3. See error

### Expected behavior
Dropdown content should be visible.

### Screenshots
Screenshot 2022-04-04 194656

### Code snippets

Vanilla Strapi installation.

joshuaellis:
### What does it do?

* uses `@internationalised/date` a lot to render the calendar and localise the view e.g. the start of the week for `en-GB` is monday, not sunday! 🥳 
* Wraps it in `DismissibleLayer` so it's correctly dismissing itself
* Moves the `SingleSelectInput` to it's own component so we can _just_ use that correctly without the bloat of the `field` stuff to create composite widgets

### Why is it needed?

* DatePickerCalendar was very poorly optimised and caused jest issues because of it.

### How to test it?

* Automated tests should still pass 🪄 

### Related issue(s)/PR(s)

* resolves CONTENT-1173
* resolves #596


packages/strapi-design-system/src/DatePicker/DatePicker.tsx
```diff
@@ -102,6 +100,7 @@ export const DatePicker = ({
         aria-label={ariaLabel}
         disabled={disabled}
         type="text"
+        placeholder="--/--/--"
```
gu-stav, line 103: I think I'd prefer if we not hardcode this placeholder.

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
```diff
@@ -54,98 +74,328 @@ export const DatePickerCalendarContent = ({
...
+          value={months[localDate.month - 1]}
+          onChange={handleMonthChange}
+        >
+          {months.map((month) => (
+            <SingleSelectOption key={month} value={month}>
+              {month}
+            </SingleSelectOption>
+          ))}
+        </SingleSelectInput>
+        <SingleSelectInput
```
gu-stav, line 169: Screenshot 2023-04-28 at 11 41 47

For some reason the options of this select have more padding to the right than to the left. Before there was a scrollbar so we had to reserve space, but I think that is no longer needed?

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
```diff
@@ -54,98 +74,328 @@ export const DatePickerCalendarContent = ({
...
           <RawTr aria-rowindex={1}>
-            <DatePickerTh>{sun}</DatePickerTh>
-            <DatePickerTh>{mon}</DatePickerTh>
-            <DatePickerTh>{tue}</DatePickerTh>
-            <DatePickerTh>{wed}</DatePickerTh>
-            <DatePickerTh>{thu}</DatePickerTh>
-            <DatePickerTh>{fri}</DatePickerTh>
-            <DatePickerTh>{sat}</DatePickerTh>
+            {weekDays.map((day) => (
+              <HeaderCell key={day}>{day}</HeaderCell>
```
gu-stav, line 186: Screenshot 2023-04-28 at 11 43 22
Screenshot 2023-04-28 at 11 43 18


The clickable area/ focus outline of the Headercell is now much smaller than it used to be. Is it worth unifying with days?

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
```diff
@@ -54,98 +74,328 @@ export const DatePickerCalendarContent = ({
...
+}
+
+const HeaderCell = ({ children, ...props }: HeaderCellProps) => {
+  return (
+    <Th {...props}>
+      <Flex height={`${24 / 16}rem`} width={`${32 / 16}rem`} justifyContent="center">
+        <Typography variant="pi" fontWeight="bold" color="neutral800" aria-hidden>
+          {children.slice(0, 2)}
+        </Typography>
+        <VisuallyHidden>{children}</VisuallyHidden>
```
gu-stav, line 303: Could this just be an aria-label instead to make the HTML structure a bit smaller? This would also avoid `aria-hidden` on Typography.

packages/strapi-design-system/src/hooks/useDateFormatter.ts
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react';
+
+import { DateFormatter } from '@internationalized/date';
+
+export interface DateFormatterOptions extends Intl.DateTimeFormatOptions {
+  calendar?: string;
+}
+
+export function useDateFormatter(locale: string, options?: DateFormatterOptions): DateFormatter {
```
gu-stav, line 9: Worth a comment what it does?

gu-stav (commented):
Code looks good to me.

A few things I observed here: https://design-system-git-chore-refactor-calendar-popover-strapijs.vercel.app/?path=/story/design-system-components-datepicker--base

- [ ] The tabbing order doesn't seem to be right: for me it goes: input, button outside, popover instead of straight to the popover
- [ ] I can't select a date using "Enter" anymore
- [ ] Using "Escape" in the Popover does not set the focus back to the input; Same for "Clear"
- [ ] Do you think there is an easy way to localize the placeholder value?

| German Placeholder | German Date Selected |
|-|-|
| Screenshot 2023-04-28 at 11 51 34 | Screenshot 2023-04-28 at 11 51 39 |

- [ ] When tabbing outside the input the Popover does not close

joshuaellis:
> The tabbing order doesn't seem to be right: for me it goes: input, button outside, popover instead of straight to the popover

> Using "Escape" in the Popover does not set the focus back to the input; Same for "Clear"

> When tabbing outside the input the Popover does not close

I forgot to re-add the `FocusTrap` which is what it should have _for now_

packages/strapi-design-system/src/DatePicker/DatePicker.tsx
joshuaellis: I'm going to remove this for now, it's a prop in another PR.

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
joshuaellis: It shouldn't be focusable to be honest. It doesn't do anything and in examples like `react-aria` it's not, I think this improvement can be made in another PR this was just refactoring as opposing to improving the experience – wdyt?

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
joshuaellis: > For some reason the options of this select have more padding to the right than to the left. Before there was a scrollbar so we had to reserve space, but I think that is no longer needed?

This is because the `min-width` of the Select dropdown is supposed to be the `min-width` of the select trigger so its technically right, what's weird is we remove the border on these selects, hence why it feels odd.

packages/strapi-design-system/src/hooks/useDateFormatter.ts
joshuaellis: yes!

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
gu-stav: Ah I see. @maevalienard do you think we could render them as standard select components?

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
gu-stav: Works for me!

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
maevalienard: Instead of simple menus you mean? @gu-stav

joshuaellis: Simple menu's aren't semantically correct – menus / menu items can't have a value so instead they should be select components – they've been styled as the their simple menu counterparts to match the DS.

gu-stav: @maevalienard ping :)

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
```diff
@@ -54,98 +76,328 @@ export const DatePickerCalendarContent = ({
...
+  const years: string[] = React.useMemo(() => {
+    const minYear = minDate ? minDate.getFullYear() : initialDate.getFullYear() - DEFAULT_PAST_RANGE;
+    const maxYear = maxDate ? maxDate.getFullYear() : initialDate.getFullYear() + DEFAULT_FUTURE_RANGE;
+
+    return [...Array(maxYear - minYear + 1).keys()].map((y) => (minYear + y).toString());
+  }, [minDate, initialDate, maxDate]);
+
+  const monthFormatter = useDateFormatter(locale, { month: 'long' });
+  const months: string[] = React.useMemo(
+    () => [...Array(12).keys()].map((m) => monthFormatter.format(new Date(Date.UTC(2023, m)))),
```
gu-stav, line 96: Can we derive how many month a year has from Intl? I know there are some with 10 and 13 months.

gu-stav:
- [ ] I am not sure if this is part of this PR, but I am not able to trigger the DatePicker Popover again, after I have selected a date or tab into the input element backwards.
- [ ] Shouldn't hitting "Esc" on the month / year select only close the select and not the whole Popover? It worked like that before.

packages/strapi-design-system/src/DatePicker/DatePickerCalendar.tsx
joshuaellis: I think the amount of months is derived from the calendar not the year but yeah we can do that.

joshuaellis:
I'm actually going to close this PR because it's all in the feature branch and i did a _huge_ refactor because there was so much flakey behaviour, sorry that's confusing however I am going to address these points:

> I am not sure if this is part of this PR, but I am not able to trigger the DatePicker Popover again, after I have selected a date or tab into the input element backwards.

This isn't a problem anymore with a Combobox/DatePicker it should only appear when you click on the input or use the arrow down see [here](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-datepicker/)

> Shouldn't hitting "Esc" on the month / year select only close the select and not the whole Popover? It worked like that before.

This is also resolved on the feature branch 💪🏼 

> Hitting "Enter" on "Clear" doesn't set the focus; may be invalid

This is also resolved on the feature branch 💪🏼 

> Changing the month from something short to something long does move the year select. I know it has been like that before - should we give both columns a fixed space?

Yeah i'd like to address this – we could do `flex-basis: 50%` so they share the space equally, or just `justify-content: space-between` wdyt @maevalienard 

> I'd argue the focus on the Popover should only be set to the month by default if the keyboard was used to open the Popover? It has been like that before and makes more sense to me :)

In the feature branch you can only control the calendar grid with arrows and tabbing takes you out of the popover, we opted to follow [this pattern](https://atlassian.design/components/datetime-picker/examples) because the `activeElement` should _always_ be the input 

I'll address the months in the next PR.

PR #971, opened 2023-04-27
Infer the `multi` prop on `Option` based on the value passed to `Select`

joshuaellis:
### What does it do?

* Uses a small context to enable multi inference for the `Option` component

### Why is it needed?

* because cloneElement is evil and must be stopped

### DIFF
--- a/packages/strapi-design-system/src/Select/Option.tsx
+++ b/packages/strapi-design-system/src/Select/Option.tsx
@@ -1,4 +1,5 @@
 import { MultiSelectOption, MultiSelectOptionProps } from './MultiSelect';
+import { useSelectContext } from './Select';
 import { SingleSelectOption, SingleSelectOptionProps } from './SingleSelect';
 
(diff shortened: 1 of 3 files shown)


gu-stav (commented):
Good idea. What do you think of a simple test to make sure we do not regress on this?

joshuaellis:
> Good idea. What do you think of a simple test to make sure we do not regress on this?

Can do if you'd like, typically i'm not adding tests for deprecated components.

gu-stav:
I think in this case it might be valuable, because the deprecation did just happen and we are still using the `multi` prop everywhere. You maybe can recycle the test you already have and just set it up using the prop instead of the components?

joshuaellis:
> I think in this case it might be valuable, because the deprecation did just happen and we are still using the `multi` prop everywhere. You maybe can recycle the test you already have and just set it up using the prop instead of the components?

It was all visual, so i've opted for a snapshot test because we're really just testing that a checkbox div element renders 😅

gu-stav (approved):

PR #980, opened 2023-05-02
Convert DatePicker to Combobox

joshuaellis:
### What does it do?

* refactors the `DatePicker` to two components `DatePicker` and `DatePickerInput` the latter does not have `Field` props so you can use it in situations where you don't want it to have it's own errors / hints and the label should be "visually hidden" e.g. `DateTimePicker`
* refactors the `DatePicker` to use a combobox-esque pattern, focus is always kept on the input while the `Calendar` can be navigated with Arrow keys (as a keyboard user)
* IT DOES NOT JOIN THE CALENDAR AND INPUT TOGETHER – this is another PR.

### Why is it needed?

* Because typing a date is _so much_ easier.

### How to test it?

* [x] automated tests

### Related issue(s)/PR(s)

* resolves CONTENT-1169
* should be merged after #970 

### Note

I'm skipping the DTP tests because it requires me to refactor parts of the TimePicker to work and it's out of scope for this PR 👍🏼


joshuaellis:
@gu-stav wanted to get some visibility on this earlier while I write the test suite. We'll get Maeva to review the `feature/dtp-improvements` branch when that has it's own PR.

gu-stav (approved):

PR #982, opened 2023-05-03
fix: Content Manager, Click outside the filter popover does not close it

gitstart:
Fixes https://github.com/strapi/strapi/issues/16381 

### What does it do?
- Adds `onClose` to `Popover` component, which would be triggered whenever the user clicks outside the popover dropdown.

### Why is it needed?
- To Support toggling the visibility of the popover whenever a user clicks outside the popover.

### How to Test
1. Open the `popover`  component in the storybook.
2. Click on the button to open the popover menu.
3. Click outside the popover menu to close it. Notice that it closes as expected.

### Demo

https://user-images.githubusercontent.com/7220167/235848848-7f7feaea-59a7-4b90-aa21-d0391c29a85d.mov


joshuaellis:
IMO we should use the `DismissibleLayer` component.

gitstart:
Hi @joshuaellis,
The popover component has been updated to use `DismissibleLayer ` component.

packages/strapi-design-system/src/Popover/Popover.tsx
```diff
@@ -137,14 +138,45 @@ const PopoverScrollable = styled(Box)`
   }
 `;
 
-type PopoverProps = ScrollingProps & Pick<ContentProps, 'source' | 'spacing' | 'fullWidth' | 'placement' | 'centered'>;
+type PopoverProps = ScrollingProps &
+  Pick<ContentProps, 'source' | 'spacing' | 'fullWidth' | 'placement' | 'centered'> & {
+    /* Handles toggling visibility, when internal Popover close events are triggered. */
+    onClose?: (reason: 'onEscapeKeyDown' | 'onPointerDownOutside' | 'onFocusOutside') => void;
```
joshuaellis, line 144: I don't think we need a reason.

PR #985, opened 2023-05-03
Make DatePicker & TimePicker comboboxes

### LINKED ISSUE #648: Manually entering date is hard/not possible

### LINKED ISSUE #955: [bug]: CalendarView on Date Picker does not disappear when clicking away

joshuaellis:
### What does it do?

* Makes the `DatePicker` a `Combobox` to allow typing
* Makes the `TimePicker` a `Combobox` to allow typing

### Why is it needed?

* Improvements to the experience 🌶️ 

### How to test it?

* Automated testing

### Related issue(s)/PR(s)

* resolves  #648 
* resolves #955


joshuaellis:
This is ready for review now, it can be merged after #1107 is merged and we can launch 1.8.0 for Strapi@4.11.0

jhoward1994 (approved):

PR #991, opened 2023-05-04
Let the DateInput control the DateCalendar

joshuaellis:
### What does it do?

* When you type a value into the input, the calendar updates with what you're typing

### Why is it needed?

* It would be weird otherwise

### How to test it?

* Automated tests 🪄

### DIFF
--- a/packages/strapi-design-system/src/DatePicker/DatePicker.tsx
+++ b/packages/strapi-design-system/src/DatePicker/DatePicker.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable no-nested-ternary */
 /* eslint-disable react/no-unused-prop-types */
 import * as React from 'react';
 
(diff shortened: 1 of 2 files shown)


gu-stav (commented):
That's a great start! One thing that bothers me is that I can't use the arrow keys in the input field, to e.g. set the cursor to a position or use shift + arrow to select something.

Also I can not select the month or year select anymore.

joshuaellis:
> That's a great start! One thing that bothers me is that I can't use the arrow keys in the input field, to e.g. set the cursor to a position or use shift + arrow to select something.

That's the pattern we opted for otherwise you can't type and control the grid at the same time. It follows the same implementation pattern as https://atlassian.design/components/datetime-picker/examples

> Also I can not select the month or year select anymore.

Do you mean with a keyboard? it's because the focus remains on the `input` now so you're never actually focussed in the calendar. So you can either type a date or the grid navigation can get you to Month/Year

gu-stav (approved):

PR #997, opened 2023-05-05
fix: setup error on windows

### LINKED ISSUE #996: [bug]: Error after cloning repo

- yarn: 3.5.0
- node: v18.11.0
- design-system repo at commit strapi/design-system@5c0de191ef278ea93a0cddadbd9b5ffbda83054f

### What's Wrong?

After cloning the repo the `yarn setup` command errors with `Entry module "src/index.ts" cannot be external.` for both the icons and the primitives package.

### To Reproduce

1. clone repo
2. run `yarn install`
3. run `yarn setup`

### Expected Behaviour

The setup should succeed and not error

Minecraftschurli:
### What does it do?

Fixed the error when running `yarn setup` on windows (#996)

### Why is it needed?

Setting up the project on windows was broken

### How to test it?

Run the setup script on windows (and other os)

### Related issue(s)/PR(s)

fixes #996

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,6 +1,6 @@
 module.exports = {
   rootDir: __dirname,
-  testMatch: ['/**/__tests__/**/?(*.)+(spec|test).[jt]s?(x)'],
+  testMatch: ['**/__tests__/**/?(*.)+(spec|test).[jt]s?(x)'],
   moduleFileExtensions: ['js', 'jsx', 'ts', 'tsx'],
   globalSetup: '<rootDir>/test/global-setup.js',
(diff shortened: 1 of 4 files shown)


joshuaellis (changes requested):
I'm not sure about changing the `external` function so much – could we not just change:

```ts
external: (id) => !id.startsWith('.') && !id.startsWith('/'), 
```

to 

```ts
external: (id) => !id.startsWith('.') && !id.startsWith(path.sep),
```

would that work instead? 🤔

joshuaellis:
closing due to inactivity.

PR #998, opened 2023-05-05
Use specific intersection ids

joshuaellis:
### What does it do?

* generate a unique intersection id instead of trying to use part of the input's id

### Why is it needed?

* r18's id hook generates `:r18` which is invalid so you need to escape the CSS, this becomes an issue in instances where we pass strange ids e.g. the RelationInput where the input name is the id. It's easier just to generate an id we know will be a number or `:r18:` so we can just strip `:`.

### DIFF
--- a/packages/strapi-design-system/src/helpers/strings.ts
+++ b/packages/strapi-design-system/src/helpers/strings.ts
@@ -1,0 +1,1 @@
+export const stripReactIdOfColon = (str: string): string => str.replaceAll(':', '');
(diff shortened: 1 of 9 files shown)


packages/strapi-design-system/src/helpers/strings.ts
```diff
@@ -0,0 +1 @@
+export const stripReactIdOfColon = (str: string): string => str.replace(/:/g, '');
```
gu-stav, line 1: This would only replace the first occurrence of a colon. If that is what we want, I'd suggest to use a different name for the helper. Also: I know it is very minor, but could please add a test?

joshuaellis: I suck at regex, how do I make it strip both out? I though /g would do that?

gu-stav: `replaceAll(':', '')` would replace all of them, even if they would be somewhere in the middle of the string. If that is a concern you could do: `.replaceAll(/^:(.*):$/g, '$1')`.

joshuaellis: lets just do replace all

gu-stav (approved):

Issue #983, opened 2023-05-03
Tracking Issue: Complete TypeScript Migration


gu-stav:
### Context

Currently, the design-system is a mix of Typescript and JavaScript. The latter prevents us e.g. from publishing types. In this issue we want to track the migration process and make sure everything is typed correctly.

If you'd like to work on a component, please comment on which one and we will add you to the list.

## Components

- [x] Accordion
- [x] Alert
- [x] Avatar
- [x] Badge
- [x] BaseButton
- [x] BaseCheckbox
- [x] BaseLink
- [x] BaseRadio
- [x] Box
- [x] Breadcrumbs
- [x] Button
- [x] Card
- [x] CarouselInput @gu-stav 
- [x] Checkbox @simotae14 
- [x] Combobox
- [x] DatePicker
- [x] DateTimePicker
- [x] Dialog @Minecraftschurli 
- [x] DismissableLayer 
- [x] Divider @gu-stav 
- [x] EmptyStateLayout @gu-stav 
- [x] Field
- [x] Flex
- [x] FocusTrap
- [x] Grid @christiancp100 
- [x] Icon
- [x] IconButton
- [x] JSONInput
- [x] KeyboardNavigable
- [x] Layout @gu-stav 
- [x] Link
- [x] LinkButton
- [x] LiveRegions
- [x] Loader
- [x] Main @gu-stav 
- [x] MainNav @gu-stav 
- [x] ModalLayout @gu-stav 
- [x] NumberInput @Minecraftschurli 
- [x] Pagination @gu-stav 
- [x] Popover
- [x] Portal
- [x] ProgressBar @gu-stav 
- [x] Radio @gu-stav 
- [x] RawTable
- [x] Searchbar
- [x] Select
- [x] SimpleMenu @Minecraftschurli 
- [x] Stack
- [x] Status
- [x] SubNav @gu-stav 
- [x] Switch @Minecraftschurli 
- [x] Table
- [x] Tabs @markkaylor 
- [x] Tag
- [x] Textarea @gu-stav 
- [x] TextButton
- [x] TextInput
- [x] ThemeProvider
- [x] TimePicker
- [x] ToggleCheckbox @markkaylor 
- [x] ToggleInput @markkaylor 
- [x] Tooltip
- [x] Typography
- [x] VisuallyHidden

### v2: located in `src/v2`

- [x] Breadcrumbs @simotae14
- [x] Link
- [x] LinkButton
- [x] MainNav @Minecraftschurli 
[...]

PR #1006, opened 2023-05-08
Chore: Convert Divider to Typescript

gu-stav:
### What does it do?

Converts the `Divider` component to TS.

### Why is it needed?


packages/strapi-design-system/src/Divider/Divider.tsx
```diff
@@ -0,0 +1,23 @@
+import styled from 'styled-components';
+
+import { Box, BoxProps } from '../Box';
+
+export interface DividerProps extends BoxProps {
```
joshuaellis, line 5:
```suggestion
export interface DividerProps extends BoxProps<HTMLHRElement> {
```

joshuaellis: Also, because we override `background` & `as` maybe we should `omit` them from this type because if you pass them they won't take affect?

joshuaellis (changes requested):

packages/strapi-design-system/src/Divider/Divider.tsx
gu-stav: Both good suggestions :) Applied them and rebased ✅

joshuaellis (approved):

PR #1007, opened 2023-05-08
Chore: Convert EmptyStateLayout to Typescript

gu-stav:
### What does it do?

Converts `EmptyStateLayout` to Typescript.

### Why is it needed?


packages/strapi-design-system/src/EmptyStateLayout/EmptyStateLayout.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { Flex, FlexProps } from '../Flex';
+import { Typography } from '../Typography';
+
+export interface EmptyStateLayoutProps extends FlexProps {
+  action?: React.ReactNode;
+  content: React.ReactNode;
+  icon: React.ReactNode;
+}
+
+const EmptyStateIconWrapper = styled(Box)<BoxProps>`
```
joshuaellis, line 13:
```suggestion
const EmptyStateIconWrapper = styled(Box)`
```

You don't need to add this :)

packages/strapi-design-system/src/EmptyStateLayout/EmptyStateLayout.tsx
```diff
@@ -0,0 +1,48 @@
+import styled from 'styled-components';
+
+import { Box, BoxProps } from '../Box';
+import { Flex, FlexProps } from '../Flex';
+import { Typography } from '../Typography';
+
+export interface EmptyStateLayoutProps extends FlexProps {
```
joshuaellis, line 7: This isn't really extending `FlexProps` it's more "picking" from them, because we never pass the rest of the spread to the node imo

packages/strapi-design-system/src/EmptyStateLayout/EmptyStateLayout.tsx
gu-stav: Ok. I thought it might make sense, so that the next developer doesn't have to type it :)

joshuaellis: It's inferred automatically, the only time you need to write them is if you're passing props that `this` styled-component  uses as opposed to `Box` e.g. you can use the value `background` in this styled component because it's part of the props of `Box`, but if you passed `hasBackground` you'd need to type that type, but you don't need to extend because `styled-components` creates the union... Maybe a diagram would help if that doesn't make sense? 👀

joshuaellis (approved):

packages/strapi-design-system/src/EmptyStateLayout/EmptyStateLayout.tsx
gu-stav: It makes sense, thanks. TS seems to be way more intelligent than flow-type where you had to type every function interface.

PR #1008, opened 2023-05-08
Chore: Convert Main to Typescript

gu-stav:
### What does it do?

Converts `Main` to Typescript

### Why is it needed?


packages/strapi-design-system/src/Main/Main.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const MainWrapper = styled(Box)`
+  // To prevent global outline on focus visible to force an outline when Main is focused
+  &:focus-visible {
+    outline: none;
+  }
+`;
+
+export const Main = ({ labelledBy = 'main-content-title', ...props }: MainProps) => {
+  return <MainWrapper aria-labelledby={labelledBy} as="main" id="main-content" tabIndex={-1} {...props} />;
```
joshuaellis, line 17: Do you know why we have this `id` here? 🤔

joshuaellis (approved):

PR #1009, opened 2023-05-08
Chore: Convert Pagination to Typescript

gu-stav:
### What does it do?

Converts `Pagination` to Typescript.

### Why is it needed?


packages/strapi-design-system/src/Pagination/components.tsx
```diff
@@ -104,45 +108,28 @@ export const NextLink = ({ children, to, ...props }) => {
...
     </PageLinkWrapper>
   );
 };
 
 PageLink.displayName = 'PageLink';
 
-export const Dots = ({ children, ...props }) => (
-  <DotsWrapper {...props} as="div">
+export const Dots = ({ children, ...props }: DotsProps) => (
+  <Box {...props}>
```
gu-stav, line 129: This may seem a bit odd, but: previously we were using a `LinkWrapper` that was rendered as a `div`, which is contradictory. I don't see why the dots-abbreviation should be usable as link and therefore I've converted it to use a Box.

packages/strapi-design-system/src/Pagination/Pagination.tsx
joshuaellis, line 1: If you've changed a component to TS try to remember to convert the test file to TS as well e.g. Pagination has an e2e test we can convert :)

packages/strapi-design-system/src/Pagination/components.tsx
joshuaellis: sgtm 👍🏼

joshuaellis (approved):
Lets just convert that e2e test and this is good to merge ⭐

PR #1017, opened 2023-05-09
Chore: Convert ProgressBar to Typescript

gu-stav:
### What does it do?

Converts `ProgressBar` to Typescript.

### Why is it needed?


packages/strapi-design-system/src/ProgressBar/ProgressBar.tsx
```diff
@@ -1,11 +1,21 @@
...
+  ? Acc[number]
+  : Enumerate<N, [...Acc, Acc['length']]>;
+
+type IntRange<F extends number, T extends number> = Exclude<Enumerate<T>, Enumerate<F>>;
+
+export interface ProgressBarProps extends Omit<BoxProps, 'background' | 'hasRadius' | 'height' | 'position' | 'width'> {
+  max: IntRange<0, 101>;
+  min: IntRange<0, 101>;
+  size: 'S' | 'M';
+  value: IntRange<0, 101>;
```
joshuaellis, line 15:
```suggestion
  max?: IntRange<0, 101>;
  min?: IntRange<0, 101>;
  size?: 'S' | 'M';
  value?: IntRange<0, 101>;
```

You've set defaults below, so these are no longer required

packages/strapi-design-system/src/ProgressBar/ProgressBar.tsx
```diff
@@ -1,11 +1,21 @@
...
+}
 
-const ProgressbarBase = styled(Box)`
+const ProgressbarBase = styled(Box)<ProgressBarProps>`
```
joshuaellis, line 18: You're only using `value` here so I would do this instead: 
```suggestion
const ProgressbarBase = styled(Box)<Required<Pick<ProgressBarProps, 'value'>>>`
```
This basically gives you:

```
{
 value: IntRange<0, 101>
}
```

We can always expand it in the future if we need too.

joshuaellis (changes requested):

joshuaellis (approved):

PR #1020, opened 2023-05-09
Convert Grid to Typescript

christiancp100:
### What does it do?

Converts `Grid` and related components to Typescript

### Related issue(s)/PR(s)


packages/strapi-design-system/src/Grid/Grid.tsx
```diff
@@ -1,19 +1,23 @@
...
-import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
 import { GridContext } from './GridContext';
-import { Box } from '../Box';
+import { Box, BoxProps } from '../Box';
 import handleResponsiveValues from '../helpers/handleResponsiveValues';
 
-const GridWrapper = styled(Box)`
+export interface GridProps extends BoxProps<HTMLDivElement> {
```
joshuaellis, line 9:
```suggestion
export interface GridProps extends BoxProps {
```

Iirc `HTMLDivElement` is the default value for this generic 👍🏼

packages/strapi-design-system/src/Grid/Grid.tsx
```diff
@@ -1,19 +1,23 @@
...
+  gridCols: number;
+  gap: number | number[];
+}
+
+const GridWrapper = styled(Box)<GridProps>`
```
joshuaellis, line 14:
```suggestion
const GridWrapper = styled(Box)<Required<Pick<GridProps, 'gridCols' | 'gap'>>>`
```

Whilst it's not technically wrong (although making two params optional will probably cause an issue) I think it's clearer just to do `Required<Pick<GridProps, 'gridCols' | 'gap'>>` only because it's the props we're using at this level so it's stricter but it's typically easier to loosen a type imo. Wdyt?

packages/strapi-design-system/src/Grid/Grid.tsx
```diff
@@ -1,19 +1,23 @@
...
 
 import { GridContext } from './GridContext';
-import { Box } from '../Box';
+import { Box, BoxProps } from '../Box';
 import handleResponsiveValues from '../helpers/handleResponsiveValues';
 
-const GridWrapper = styled(Box)`
+export interface GridProps extends BoxProps {
+  gridCols?: number;
+  gap?: number | number[];
```
HichamELBSI, line 11: Do you think we could create a ResponsiveValue type that will be used in all the responsive properties? We could have a stronger typing by using the `theme.spaces` values as well 🤷🏽‍♂️ .

The `Proptypes` of this component weren't updated I think because responsive values can also be set as an object `{ desktop: ..., mobile: ..., tablet: ... }`.

HichamELBSI (changes requested):

packages/strapi-design-system/src/Grid/Grid.tsx
joshuaellis: something like `keyof DefaultTheme['spaces'] | Array<keyof DefaultTheme['spaces']>` similar to what we use in `Flex` could be an option?

HichamELBSI: Yes, it could be. We'll need to handle the object type as well `{ desktop: ..., mobile: ..., tablet: ... }`

joshuaellis: Could even do a named `tuple` too 👍🏼

HichamELBSI (approved):
Looks pretty good, nice job @christiancp100!

packages/strapi-design-system/src/Grid/Grid.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+const GridWrapper = styled(Box)<Required<Pick<GridProps, 'gridCols' | 'gap'>>>`
+  display: grid;
+  grid-template-columns: repeat(${({ gridCols }) => gridCols}, 1fr);
+  ${({ theme, gap }) => handleResponsiveValues('gap', gap, theme)}
+`;
+
+export const Grid: React.FC<GridProps> = ({ gap = '0', gridCols = 12, ...props }) => {
+  const context = React.useMemo<Required<Pick<GridProps, 'gridCols' | 'gap'>>>(
```
joshuaellis, line 21:
```suggestion
  const context = React.useMemo(
```

This should be inferred imo

packages/strapi-design-system/src/Grid/Grid.tsx
```diff
@@ -0,0 +1,31 @@
...
+  gap?: ResponsiveValue;
+}
+
+const GridWrapper = styled(Box)<Required<Pick<GridProps, 'gridCols' | 'gap'>>>`
+  display: grid;
+  grid-template-columns: repeat(${({ gridCols }) => gridCols}, 1fr);
+  ${({ theme, gap }) => handleResponsiveValues('gap', gap, theme)}
+`;
+
+export const Grid: React.FC<GridProps> = ({ gap = '0', gridCols = 12, ...props }) => {
```
joshuaellis, line 20:
```suggestion
export const Grid = ({ gap = '0', gridCols = 12, ...props }: GridProps) => {
```

While it's not wrong, we haven't been using FC at the moment (I'm not sure i see the benefits? Happy to be shown though!)

packages/strapi-design-system/src/Grid/GridItem.tsx
joshuaellis, line 1: I think we can remove the `Box` child and just make `GridItemWrapper` a `styled(Box)` and pass it _allllll_ the props – wdyt? Makes a smaller DOM footprint 👍🏼

packages/strapi-design-system/src/Grid/GridItem.tsx
```diff
@@ -19,7 +27,7 @@ const GridItemWrapper = styled.div`
   }
 `;
 
-export const GridItem = ({ col, xs, s, ...props }) => {
+export const GridItem: React.FC<GridItemProps> = ({ col, xs, s, ...props }) => {
```
joshuaellis, line 30:
```suggestion
export const GridItem = ({ col, xs, s, ...props }: GridItemProps) => {
```

packages/strapi-design-system/src/Grid/GridItem.tsx
```diff
@@ -1,12 +1,20 @@
...
+
+export interface GridItemProps extends BoxProps<HTMLDivElement> {
+  col: number;
+  s: number;
+  xs: number;
+  gap: ResponsiveValue;
+  gridCols: number;
+}
+
+const GridItemWrapper = styled.div<GridItemProps>`
```
joshuaellis, line 17: If we don't refactor this to be a styled Box, then we need to pick the specific types to the ones we're using, I also don't see where `gap` or `gridCols` is used – is this even necessary?

joshuaellis: Which makes me think we're using context for absolutely no reason?

joshuaellis (changes requested):
Minor feedback on the types

packages/strapi-design-system/src/Grid/Grid.tsx
christiancp100: I thought the same, but it's not 😢 And it made me do that ugly thing

packages/strapi-design-system/src/Grid/Grid.tsx
christiancp100: IMO it provides better type safety by explicitly indicating that the component is a function component and declaring the props type. It also gives better IntelliSense support in IDEs providing types suggestions like `displayName` and others. Even though, I see you point, if we are not using it maybe we don't want to start now, or do we? Can this be a FE sync topic?

joshuaellis: > Can this be a FE sync topic 

I think that's a great idea, if you could present the pros / cons etc. I'd be interested to see and we can do the mass change in one PR if we decide to go for it 👍🏼

Minecraftschurli:
My findings through #1066 are that the context is not necessary and both Grid and GridItem can be styled boxes.

christiancp100:
> My findings through #1066 are that the context is not necessary and both Grid and GridItem can be styled boxes.

Thank you for you help! We've considered it and you were totally right

joshuaellis (approved):

PR #1023, opened 2023-05-10
Migrate ToggleInput/ToggleCheckbox to TypeScript

markkaylor:
### What does it do?
Migrate ToggleInput to TypeScript
Migrate ToggleCheckbox to TypeScript

### Why is it needed?


packages/strapi-design-system/src/ToggleInput/ToggleInput.tsx
```diff
@@ -19,21 +18,36 @@ const ClearButton = styled(TextButton)`
   margin-left: auto;
 `;
 
+interface ToggleInputProps {
```
joshuaellis, line 21: This should probably extend `ToggleCheckboxProps` (if it exists yet) because we pass the spread to `ToggleCheckbox`. Also I think you should also extend some of the `FieldProps` which means you don't need to retype `disabled` or `error` etc as they're passed to that component :)

packages/strapi-design-system/src/ToggleInput/ToggleInput.tsx
```diff
@@ -19,21 +18,36 @@ const ClearButton = styled(TextButton)`
...
+  disabled?: boolean;
+  error?: string;
+  hint?: string;
+  id?: string;
+  label?: string;
+  labelAction?: React.ReactNode;
+  name?: string;
+  onClear?: () => void;
+  required?: boolean;
+  size?: keyof typeof sizes.input;
```
joshuaellis, line 33: We type this a lot manually, Maybe it's worth exporting a union from somewhere central for `InputSizes = "S" | "M"` 🤔  wdyt?

joshuaellis (changes requested):

packages/strapi-design-system/src/ToggleInput/ToggleInput.tsx
joshuaellis: yeah that'd be cool, I think it'd be something like `type InputSizes = keyof ThemeSizes['input']` maybe?

joshuaellis (approved):
Nice work ⭐

PR #1025, opened 2023-05-10
Update Storybook to v7

joshuaellis:
### What does it do?

* Updates storybook to v7

### Why is it needed?

* Updating deps is important kids


gu-stav (approved):
Very nice, thank you!

PR #1028, opened 2023-05-10
Chore: Simplify Textarea component

gu-stav:
### What does it do?

Simplifies the `Textarea` component a bit. It seems to have grown over time.

### Why is it needed?

I didn't want to mix the refactoring with the TS conversion.


packages/strapi-design-system/src/Textarea/TextareaInput.js
joshuaellis, line 1: Can we move this into `Textarea` please? this is inline with what we're doing with `TimePicker` or `Combobox` :)

joshuaellis:
Just that comment, but very nice 🧽

packages/strapi-design-system/src/Textarea/TextareaInput.js
gu-stav: Done in 47a23c9f7d6d0f71039cbf90b5bd9ee8d77d55de

joshuaellis (approved):

PR #1029, opened 2023-05-10
Migrate Tabs to TypeScript

markkaylor:
### What does it do?
Migrate Tabs to TypeScript

### Why is it needed?


packages/strapi-design-system/src/Tabs/TabGroup.tsx
```diff
@@ -1,14 +1,31 @@
...
+
+    const Tabs = React.Children.toArray(props.children).find((child) => {
+      if (React.isValidElement(child)) {
+        const componentType = child.type as React.ComponentType<any>;
+
+        return componentType.displayName === 'Tabs';
+      }
+
+      return false;
+    }) as React.ReactElement | undefined;
```
markkaylor, line 27: This is the only way I could find to make TypeScript happy happy...

But actually we never set a displayName on Tabs and doesn't seem to be inferred in anyway. So as far as I can tell this will always end up undefined and below on line 31 the condition will never pass. Maybe someone has some more context or understands better what's going on? Otherwise could we remove both?

packages/strapi-design-system/src/Tabs/useTabsFocus.ts
joshuaellis, line 1: This is only used in `Tabs` so lets just move it there imo

packages/strapi-design-system/src/Tabs/Tabs.tsx
```diff
@@ -185,7 +207,13 @@ export const Tab = ({ disabled, id, children, variant, hasError, index, selected
...
   }
 
-  const showRightBorder = selectedTabIndex - 1 === index;
+  const showRightBorder = () => {
+    if (selectedTabIndex) {
+      return selectedTabIndex - 1 === index;
+    }
+
+    return false;
+  };
```
joshuaellis, line 216: How come you've made this a function?

packages/strapi-design-system/src/Tabs/Tabs.tsx
```diff
@@ -15,12 +14,16 @@ const TabButton = styled.button`
   }
 `;
 
-export const Tabs = ({ children, ...props }) => {
+interface TabsProps {
```
joshuaellis, line 17: This needs to extend the props of `DefaultTabsRow` because you're spreading them into it.

Also you need to export the prop types.

packages/strapi-design-system/src/Tabs/TabGroup.tsx
joshuaellis: remove it. honestly it's awful and should never have been written in the first place.

joshuaellis (changes requested):
Sorry i can't write this everywhere lol but you keep forgetting that you're spreading props onto "something" e.g. `Box` so therefore the prop type has to `extend` the props of said element. An simple example to look at would be `Tag`. Lmk if it's not clear and we can go through it in more detail :)

packages/strapi-design-system/src/Tabs/Tabs.tsx
markkaylor: Just not sure why I would need to export the TabsProps interface here. Isn't it only used in this file?

packages/strapi-design-system/src/Tabs/Tabs.tsx
markkaylor: Probably just to make something work then forgot about it. I updated with a better solution... I think.

packages/strapi-design-system/src/Tabs/Tabs.tsx
joshuaellis: Because these types will eventually be consumed by other components e.g. in the admin panel and it's good practice to instead of writing the type again, to create a new type from the existing type (maximum inference and it's easier to see if we change a type somewhere the effect it has through out the codebase)

packages/strapi-design-system/src/Tabs/TabGroup.tsx
```diff
@@ -1,20 +1,19 @@
...
 
+export interface TabGroupProps extends Omit<TabsContextState, 'id'>, React.HTMLAttributes<HTMLDivElement> {
+  children: React.ReactNode;
+  id?: string;
+  initialSelectedTabIndex?: number;
+}
+
 export const TabGroup = React.forwardRef(
-  ({ id, initialSelectedTabIndex, label, onTabChange, variant, ...props }, ref) => {
+  ({ id, initialSelectedTabIndex, label, onTabChange = () => {}, variant, ...props }: TabGroupProps, ref) => {
```
joshuaellis, line 13: I can't comment directly with a suggestion but you type `forwardRef` as: 
```js
React.forwardRef<MyElementType, MyPropsType>((props, ref) => { /* component */})
```

joshuaellis (commented):
think it's nearly there, just a little more polishing – you definitely picked one of the more trickier components but great job so far ⭐

joshuaellis (approved):

PR #1031, opened 2023-05-10
Convert NumberInput to Typescript

Minecraftschurli:
### What does it do?
Converts NumberInput to Typescript

### Related issue(s)/PR(s)


joshuaellis (changes requested):
You're using the html attribute type for `step` which means you have to cast it, because afaik we don't pass it to the input(?) you can actually specifically type this as a number to reduce casting with `Number(value)`

joshuaellis (approved):

PR #1034, opened 2023-05-10
Convert Dialog to Typescript

Minecraftschurli:
### What does it do?
Converts Dialog to Typescript

### Related issue(s)/PR(s)


packages/strapi-design-system/src/Dialog/Dialog.tsx
```diff
@@ -29,7 +28,14 @@ const DialogHeader = styled(Flex)`
   border-bottom: 1px solid ${({ theme }) => theme.colors.neutral150};
 `;
 
-export const Dialog = ({ onClose, title, as, isOpen, id, ...props }) => {
+export interface DialogProps extends Omit<BoxProps, 'id'> {
+  onClose: () => void;
+  title: string;
+  isOpen: boolean;
+  id?: string | number | undefined;
+}
```
joshuaellis, line 36:
```suggestion
export interface DialogProps extends BoxProps {
  onClose: () => void;
  title: string;
  isOpen: boolean;
}
```

I know you were following the PropTypes, but lets just do this for simplicity 😄

joshuaellis (changes requested):
Minor changes, but otherwise looks great, thanks for the work!

joshuaellis (changes requested):

PR #1035, opened 2023-05-11
Bring primitive/Combobox popover behaviour to be in line with Select

joshuaellis:
### What does it do?

* Uses `FocusScope` to capture the focus for the input and popover of the Primitive/Combobox
* Removes the `handleDismiss` callback to "sort out" the value on blur and moves it to the input to make it more predictable
* Adds `useFocusGuards` similarly to be inline with the Select primitive

### Why is it needed?

* The popover wouldn't close when you tabbed away from it to another focussable element and if you added the expected callbacks we wanted when you leave the combobox it would cause other issues. It was simpler and perhaps more predictable to match the behaviour of the Select enforcing a user to close the popover first before tabbing away. It also means the aria-scoping makes more sense because you are in fact locked to what you can focus.

### How to test it?

* Automated testing! Turns out my tests weren't that great but doing this meant I could "unskip" them.

### DIFF
--- a/docs/stories/Select.stories.mdx
+++ b/docs/stories/Select.stories.mdx
@@ -118,7 +118,7 @@ const customizeContent = (values: string[]) => `${values.length} are currently s
           required
           placeholder="My favourite fruit is..."
           onClear={() => {
-            setValue(undefined);
+            setValues(undefined);
           }}
           value={values}
           onChange={setValues}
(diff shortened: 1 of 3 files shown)


joshuaellis:
This will be the last PR for 1.7.5 so I can release this version and prep a hotfix for the CMS product.

gu-stav (approved):

PR #1039, opened 2023-05-11
Convert Status to Typescript

Minecraftschurli:
### What does it do?
Converts Status to Typescript

### Related issue(s)/PR(s)


packages/strapi-design-system/src/Status/__tests__/Status.spec.tsx
```diff
@@ -28,8 +28,9 @@ describe('Status', () => {
 
     const textNode = queryByText('My status');
 
+    expect(textNode).not.toBeNull();
     await waitFor(() => {
-      expect(textNode.querySelector('div')).toBeInTheDocument();
+      expect(textNode!.querySelector('div')).toBeInTheDocument();
```
joshuaellis, line 33: Hmmm i'm not actually a fan of these tests at all, controversial but can you add:

```ts
// @ts-expect-error TODO: refactor these tests
```

to both the uses of `!` please? :)

joshuaellis (commented):
Minor request but the code is great, thanks!

joshuaellis (approved):

PR #1047, opened 2023-05-16
fix: LinkButton does not open in new tab

gitstart:
Fixes https://github.com/strapi/strapi/issues/16378

### What does it do?
- External links do not open a new tab with `isExternal` set to `true`. 

### Why is it needed?
- External links open in a new tab with `isExternal` set to `true` on `LinkButton`. 

### How to test it?
1. Run the app `yarn develop`
2. Find  `LinkButton` in `V2` and click on `Default` button.
3. See that it opens a new tab.

## Demo

https://github.com/strapi/design-system/assets/1501599/d21a0f27-2f7e-49d4-a775-14d4feb9e44e

### DIFF
--- a/packages/strapi-design-system/src/v2/LinkButton/LinkButton.tsx
+++ b/packages/strapi-design-system/src/v2/LinkButton/LinkButton.tsx
@@ -42,11 +42,23 @@ const LinkWrapper = styled(BaseButtonWrapper)`
 
 export const LinkButton = React.forwardRef<HTMLAnchorElement, LinkButtonProps>(
   (
-    { variant = 'default', startIcon, endIcon, disabled = false, children, size = 'S', as = BaseLink, ...props },
+    {
+      variant = 'default',
+      startIcon,
+      endIcon,
+      disabled = false,
+      children,
+      isExternal,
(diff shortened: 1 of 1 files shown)


packages/strapi-design-system/src/v2/LinkButton/LinkButton.tsx
```diff
@@ -42,11 +42,23 @@ const LinkWrapper = styled(BaseButtonWrapper)`
 
 export const LinkButton = React.forwardRef<HTMLAnchorElement, LinkButtonProps>(
```
gitstart, line 43: The current implementation of the Flex and Box components is currently limited due to the utilization of the shouldForwardProp function from styled-components. While it functions well for the majority of use cases, it can pose challenges when we utilize the as property to bind a different component for rendering in the DOM, as the necessary props are not forwarded to the component.

To illustrate this behavior, [here's a brief demonstration](https://codesandbox.io/s/styled-components-forked-ymdeqz?file=/index.js).

joshuaellis:
This is fixed in v2 of the design-system #1738

PR #1041, opened 2023-05-12
refactor: date time picker

joshuaellis:
### What does it do?

* refactors the date time picker and tests inline with changes in the subcomponents

### Why is it needed?

* The date time picker was broken otherwise!

### How to test it?

* automated testing but also, give it a go!

PR #1050, opened 2023-05-16
Allow partial date & time values to be typed and auto-assume the intended value on blur

joshuaellis:
### What does it do?

* adds the ability to type a partial date value in the DatePicker e.g. `11` and that would translate to `11th May 2023` (because it uses the calendar date as the basis which defaults to `today`)
* adds the ability to type a partial time value in the TimePicker e.g. `11` and that would translate to `11:00`

### Why is it needed?

* Part of DateTimePicker improvements

### How to test it?

* Type partial values and leave the input field 🤞🏼 

### Related issue(s)/PR(s)

* is a continuation of #1041 so should be merged afterwards


packages/primitives/src/components/Combobox/Combobox.tsx
```diff
@@ -273,9 +260,76 @@ const ComboboxTrigger = React.forwardRef<ComboboxTriggerElement, TriggerProps>((
...
+
+            if (target.hasPointerCapture(event.pointerId)) {
+              target.releasePointerCapture(event.pointerId);
+            }
+
+            /**
+             * This has been added to allow events inside the trigger to be easily fired
+             * e.g. the clear button or removing a tag
+             */
+            const buttonTarg = target.closest('button') ?? target.closest('div');
```
gu-stav, line 315: I'm wondering if closest `div` is too generic here and would always trigger, because there is always a div in the parents?

packages/strapi-design-system/src/Combobox/Combobox.tsx
```diff
@@ -30,9 +30,12 @@ export interface ComboboxInputProps
       | 'defaultTextValue'
       | 'required'
       | 'disabled'
+      | 'isPrintableCharacter'
     >,
-    Pick<FieldProps, 'error' | 'id'> {
+    Pick<FieldProps, 'error' | 'id'>,
+    Omit<ComboboxPrimitive.TextInputProps, 'required' | 'disabled' | 'value' | 'onChange' | 'size'> {
   children: React.ReactNode;
+  className?: string;
```
gu-stav, line 38: Do we really need that? Imo that is a very uncommon pattern in our codebase and I'm wondering if the Combobox should somehow translate between Radix and the components using it instead of exposing such a low level prop.

packages/strapi-design-system/src/TimePicker/TimePicker.tsx
```diff
@@ -7,15 +7,31 @@ import { ComboboxInput, ComboboxInputProps, Option } from '../Combobox/Combobox'
 import { useDesignSystem } from '../DesignSystemProvider';
 import { Field, FieldError, FieldHint, FieldLabel, FieldProps } from '../Field';
 import { Flex } from '../Flex';
+import { useControllableState } from '../hooks/useControllableState';
 import { useDateFormatter } from '../hooks/useDateFormatter';
 import { useId } from '../hooks/useId';
 
+const isPrintableCharacter = (str: string): boolean => {
+  return Boolean(str.match(/^[^a-zA-Z]*$/));
```
gu-stav, line 15: I find that confusing: if I read the regex currently it says: match any string that starts with a non-alphabetical character, which contradicts the util name.

packages/strapi-design-system/src/TimePicker/TimePicker.tsx
```diff
@@ -44,16 +62,41 @@ export const TimePickerInput = ({
    * @deprecated This is no longer used.
    */
   selectButtonTitle: _selectButtonTitle,
+  textValue: textValueProp,
+  defaultTextValue = '',
+  onTextValueChange,
+  value: valueProp,
+  defaultValue,
```
gu-stav, line 69: Just out of curiosity: why is that needed if you have `useControllableState`? I thought for the other inputs we only needed `value`.

gu-stav (commented):
Haven't done functional testing yet and added only a few (minor) comments. The seperator stuff looks very good. Well done!

gu-stav (commented):
Functional testing:

- [ ] The timepicker does not allow me to select a range and replace it (works in the datpicker):

https://github.com/strapi/design-system/assets/2244375/62aa54f9-996e-4a51-ae3a-7dd6bef8b286

- [x] similar problem as for relations in the CMS: the popover doesn't close when tabbing away (datepicker to timepicker)
- [ ] I can't delete a character in the timepicker

https://github.com/strapi/design-system/assets/2244375/3bce632e-127d-45af-bd6d-34e8d5563e67

- [] Copy and paste mistake with the error message

Screenshot 2023-05-17 at 11 45 19

- [] Don't know why that happens if the cursor is on the clock icon while using the arrow keys

https://github.com/strapi/design-system/assets/2244375/f66ef63a-eb15-4870-9638-342c8b270025

- [x] Datepicker: still the question: how do I get to the month / year select using the keyboard?

- [x] While the German date format works (DD/MM/YYYY) the separator should be `.` instead.

packages/strapi-design-system/src/TimePicker/TimePicker.tsx
joshuaellis: > which contradicts the util name

does it? I suppose the regex value infers what a printable character is... we could rename it to be `isNotAlphabeticalCharacter`?

packages/strapi-design-system/src/Combobox/Combobox.tsx
joshuaellis: > if the Combobox should somehow translate between Radix and the components using it

I'm not sure what you're suggesting here.

To answer your question, we have it because the DTP has a 70/30 flex split with the DatePicker & TimePicker (respectively), I suppose we could style the `Flex` to declare it but I find that to be more brittle than using an API which is endorsed by the library https://styled-components.com/docs/basics#extending-styles 

In response to it being uncommon in the codebase – that's why we have over-engineered dom code because we're for some reason opposed to this API – I don't think there's one answer over the other, i just think it's best to use them interchangeably based on the use-case 😅

packages/strapi-design-system/src/TimePicker/TimePicker.tsx
joshuaellis: If you're controlling the value you might need to control the text value e.g. with relations we clear the field. These were passed straight to the combobox but i've had to expose them because i'm controlling the input internally to format it correctly :)

joshuaellis:
Some of your comments are general bugs, others relate to the first PR that ideally should be merged so i'm gonna close the first and this will just have to be a bigger PR...

>  Datepicker: still the question: how do I get to the month / year select using the keyboard?

still the answer – you can't just like you cant get to the buttons on [this](https://atlassian.design/components/datetime-picker/examples), I'd argue you don't really need them if you can use a keyboard because you can type 🤔  

> While the German date format works (DD/MM/YYYY) the separator should be . instead.

Where are you seeing this? In the locale example, it's got the correct separator.

joshuaellis:
>  Don't know why that happens if the cursor is on the clock icon while using the arrow keys

I can't reproduce this.

gu-stav (commented):
More functional testing:

- [x] When selecting a data and editing something then tabbing away the change is not persisted

https://github.com/strapi/design-system/assets/2244375/1ec0e0a7-83e2-4abb-8151-188c58dc666f

similar when hitting enter:

https://github.com/strapi/design-system/assets/2244375/ae3ddc2c-2720-4d1c-b9f5-13951b329bcd

- [ ] When typing a time, scrolling is triggered

https://github.com/strapi/design-system/assets/2244375/83d38e79-50b5-48f1-a743-c26632c9e264

- [ ] After I have selected a month AND date, I can't use shift + arrow keys any longer to edit the date

https://github.com/strapi/design-system/assets/2244375/9adf2054-5b32-4a17-abb0-cb9e313ea1bc

Maybe it makes debugging easier: this happens every time on https://design-system-git-feat-partial-values-strapijs.vercel.app/?path=/story/design-system-components-datetimepicker--initial-data without changes.

joshuaellis:
> After I have selected a month AND date, I can't use shift + arrow keys any longer to edit the date

That's because the Popover is open and the arrow keys control the popover incl. left and right :)

gu-stav (commented):
- [x] Can't tab through the form any longer (set focus, tab); same on the timepicker and tabbing backwards as soon as the popover is open (IMO it should just close when tabbing)

https://github.com/strapi/design-system/assets/2244375/2db9d5ab-2908-4aac-9d8e-20e17a9949b7

- [x] Scroll trigger

>>    When typing a time, scrolling is triggered
> This one i can't reproduce but i fixed the issue with the two popovers, so hopefully that solved the issue?

It seems to happen only when I go into the table view of storybook.

- [ ] The selected time is not selected in the Popover and scrolling doesn't start there (which is annoying if you've typed 10:00 and want to use arrow-down to go to the next interval)

https://github.com/strapi/design-system/assets/2244375/fc28140f-51fa-401f-ac7b-d5d54d5b6cc8

- [x] When I type rubbish the datepicker sets the date to ... something

https://github.com/strapi/design-system/assets/2244375/ec01fcd2-d052-45b2-bdca-307322fb325b

I'm not sure what has been discussed in terms of error handling, but I'd expect the rubbish date to be in there still and an error message to be shown, because like that a typo may lead to a date entered the user didn't want.

- [ ] Using shift when the popover is open

>> After I have selected a month AND date, I can't use shift + arrow keys any longer to edit the date
> That's because the Popover is open and the arrow keys control the popover incl. left and right :)

I'm really not sure about this - I personally don't like it at all and would expect that I could still navigate within the input. What happens in the reference example you've used for the pattern?

gu-stav (approved):
Really cool stuff.

PR #1051, opened 2023-05-16
fix: the focus-trap should be on the primitive trigger

joshuaellis:
### What does it do?

* The focus trap should actually be on the input, not the children of the provider

### Why is it needed?

* Because FocusTrap expects one child not multiple in the tree

### How to test it?

* Open the combobox
* You shouldn't be able to tab out of it whilst the popover is open

### DIFF
--- a/packages/primitives/src/components/Combobox/Combobox.tsx
+++ b/packages/primitives/src/components/Combobox/Combobox.tsx
@@ -388,198 +388,174 @@ const ComboxboxTextInput = React.forwardRef<ComboboxInputElement, TextInputProps
   }, [context.textValue, context.filterValue, startsWith, context.visuallyFocussedItem, getItems, previousFilter]);
 
   return (
-    <FocusScope
-      // we make sure we're not trapping once it's been closed
-      // (closed !== unmounted when animating out)
-      trapped={context.open}
-      onMountAutoFocus={(event) => {
-        // we prevent open autofocus because we manually focus the selected item
-        event.preventDefault();
-      }}
-      onUnmountAutoFocus={(event) => {
-        inputRef.current?.focus({ preventScroll: true });
-        /**
-         * In firefox there's a some kind of selection happening after
-         * unmounting all of this, so we make sure we clear that.
-         */
-        document.getSelection()?.empty();
-        event.preventDefault();
-      }}
-    >
-      <input
-        type="text"
-        role="combobox"
-        aria-controls={context.contentId}
-        aria-expanded={context.open}
-        aria-required={context.required}
-        aria-autocomplete={context.autocomplete}
-        data-state={context.open ? 'open' : 'closed'}
-        aria-disabled={isDisabled}
-        aria-activedescendant={context.visuallyFocussedItem?.id}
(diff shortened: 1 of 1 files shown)


packages/primitives/src/components/Combobox/Combobox.tsx
```diff
@@ -273,9 +256,76 @@ const ComboboxTrigger = React.forwardRef<ComboboxTriggerElement, TriggerProps>((
...
+        <div
+          ref={forwardedRef}
+          data-disabled={context.disabled ? '' : undefined}
+          {...triggerProps} // Enable compatibility with native label or custom `Label` "click" for Safari:
+          onClick={composeEventHandlers(triggerProps.onClick, () => {
+            // Whilst browsers generally have no issue focusing the trigger when clicking
+            // on a label, Safari seems to struggle with the fact that there's no `onClick`.
+            // We force `focus` in this case. Note: this doesn't create any other side-effect
+            // because we are preventing default in `onPointerDown` so effectively
+            // this only runs for a label "click"
```
HichamELBSI, line 295: 👍🏽 Super clear thx.

HichamELBSI (approved):
Looks quite good (thinking out loud, this 400+ file could be splitted for a better readability. We could move the `ComboboxTrigger` and the `ComboxboxTextInput` in separate files, wdyt?).

joshuaellis:
> thinking out loud, this 400+ file could be splitted for a better readability. We could move the ComboboxTrigger and the ComboxboxTextInput in separate files, wdyt?

IMO there's no right answer to this one, there's pros and cons to each approach i.e. I personally prefer having deep files with a clear API surface – we export these components from this file and we know those are consumed publicly, I believe it removes a level of cognitive overhead jumping from file X to Y. In addition I'm a bit of a believer in if it's in it's own file we need to test it rigorously because you could mutate this one file without knowing what else it impacts even though it requires it's partner file to render (due to using context)

On the other side to this, it's a very complicated file I won't deny this but if the file isn't clear then we need to add more inline comments / documentation etc. wydt?

HichamELBSI:
> > thinking out loud, this 400+ file could be splitted for a better readability. We could move the ComboboxTrigger and the ComboxboxTextInput in separate files, wdyt?
> 
> IMO there's no right answer to this one, there's pros and cons to each approach i.e. I personally prefer having deep files with a clear API surface – we export these components from this file and we know those are consumed publicly, I believe it removes a level of cognitive overhead jumping from file X to Y. In addition I'm a bit of a believer in if it's in it's own file we need to test it rigorously because you could mutate this one file without knowing what else it impacts even though it requires it's partner file to render (due to using context)
> 
> On the other side to this, it's a very complicated file I won't deny this but if the file isn't clear then we need to add more inline comments / documentation etc. wydt?

For me, all this must be well tested whether it is in a separate file or not, so we could keep the current state of the tests despite the split into several files, but I understand your point of view because indeed, once they are split, we will have to export them and therefore potentially used in other contexts.
Anyway, we can leave this file like this, no worries, I really wanted to know your thoughts about this kind of files 👍🏽.

PR #1056, opened 2023-05-17
Flatten CrossCircle icon

mazzucchelli:
### What does it do?

Flatten the structure of `CrossCircle` icon

### Why is it needed?

Correctly style it in both dark and light theme using the `<Icon />` component

before:
Screenshot 2023-05-17 alle 12 52 53

after:
Screenshot 2023-05-17 alle 12 48 44


joshuaellis (approved):
really nice 🌟

PR #1058, opened 2023-05-17
Convert Breadcrumbs v2 to Typescript

simotae14:
### What does it do?

Converts `Breadcrumbs` to Typescript

### Related issue(s)/PR(s)


packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.tsx
```diff
@@ -20,15 +19,10 @@ const StyledLink = styled(BaseLink)`
   }
 `;
 
-export const CrumbLink = ({ children, ...props }) => <StyledLink {...props}>{children}</StyledLink>;
+export interface CrumbLinkProps extends BaseLinkProps {
+  to?: string | undefined;
```
joshuaellis, line 23:
```suggestion
```

I'd omit this, this prop isn't on `BaseLink` because it's a react-router-dom prop so this isn't the place for it imo

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.tsx
```diff
@@ -16,15 +15,15 @@ const StyledButton = styled(Button)`
   }
 `;
 
-export const CrumbSimpleMenu = ({ children, ...props }) => (
-  <SimpleMenu noBorder icon={<CarretDown />} as={StyledButton} size="S" {...props}>
+export interface CrumbSimpleMenuProps extends ButtonProps {
+  'aria-label': string;
+  icon?: React.ReactElement;
```
joshuaellis, line 20:
```suggestion
```

Not sure we need this?

joshuaellis (changes requested):
Minor tweaks, but otherwise great work ⭐

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.tsx
```diff
@@ -16,15 +15,15 @@ const StyledButton = styled(Button)`
...
+}
+
+export const CrumbSimpleMenu = ({ children, ...props }: CrumbSimpleMenuProps) => (
+  <SimpleMenu icon={<CarretDown />} as={StyledButton} size="S" {...props}>
```
gu-stav, line 24: Where did the `noBorder` prop go? I'm not sure, but I think to remember it is needed, because it behaves, but doesn't look like a simple menu.

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.tsx
simotae14: otherwise the `SimpleMenu` gave me an error
Schermata 2023-05-24 alle 11 19 49

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.tsx
simotae14: But the SimpleMenu has not the noBorder prop, here you can see the error I had with the noBorder attribute

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbLink.tsx
simotae14: ok

packages/strapi-design-system/src/v2/Breadcrumbs/CrumbSimpleMenu.tsx
gu-stav: Maybe the type definitions are wrong? I'm not sure. In any case: the menu looks broken: https://design-system-git-chore-breadcrumbs-ts-strapijs.vercel.app/?path=/story/design-system-components-v2-breadcrumbs--base

Screenshot 2023-05-24 at 12 03 00

joshuaellis: @simotae14 did you solve this puzzle?

gu-stav: I've also created an issue, because this is already broken on main: https://github.com/strapi/design-system/issues/1090

simotae14: ok then I can take a look at the issue. Maybe in the meantime you can review this PR which is unrelated to the bug

joshuaellis (approved):
Compared this version to the current version on main ✅

PR #1066, opened 2023-05-19
Convert Grid to Typescript and remove the context

Minecraftschurli:
### What does it do?
Converts Grid to Typescript and removes the context

### Related issue(s)/PR(s)
alternative to https://github.com/strapi/design-system/pull/1020 by @christiancp100


joshuaellis:
Thanks for this, but i think we'll stick with https://github.com/strapi/design-system/pull/1020 👍🏼

PR #1068, opened 2023-05-19
Convert v2/MainNav to Typescript

Minecraftschurli:
### What does it do?
Converts v2/MainNav to Typescript

### Related issue(s)/PR(s)

### DIFF
--- a/packages/strapi-design-system/src/v2/MainNav/MainNav.tsx
+++ b/packages/strapi-design-system/src/v2/MainNav/MainNav.tsx
@@ -10,7 +10,7 @@ const MainNavWrapper = styled(Flex)<{ condensed: boolean }>`
   border-right: 1px solid ${({ theme }) => theme.colors.neutral150};
 `;
 
-export interface MainNavProps extends FlexProps {
+export interface MainNavProps extends FlexProps<'nav'> {
   condensed?: boolean;
 }
 
--- a/packages/strapi-design-system/src/v2/MainNav/NavBrand.tsx
+++ b/packages/strapi-design-system/src/v2/MainNav/NavBrand.tsx
@@ -34,6 +34,10 @@ export interface NavBrandProps extends BaseLinkProps {
 export const NavBrand = React.forwardRef<HTMLAnchorElement, NavBrandProps>(
   ({ workplace, title, icon, ...props }, ref) => {
     const condensed = useMainNav();
+    /**
+     * TODO: this shouldn't be here, because we're assuming you're
+     * passing a ReactRouter Link which isn't necessarily the case.
+     */
     props.to = props?.to ?? '/';
 
     if (condensed) {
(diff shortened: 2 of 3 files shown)


packages/strapi-design-system/src/v2/MainNav/MainNav.tsx
```diff
@@ -1,17 +1,20 @@
...
+import { Flex, FlexProps } from '../../Flex';
 
-const MainNavWrapper = styled(Flex)`
+const MainNavWrapper = styled(Flex)<{ condensed: boolean }>`
   width: ${({ condensed }) => (condensed ? 'max-content' : `${224 / 16}rem`)};
   border-right: 1px solid ${({ theme }) => theme.colors.neutral150};
 `;
 
-export const MainNav = ({ condensed, ...props }) => {
+export interface MainNavProps extends FlexProps {
```
joshuaellis, line 13:
```suggestion
export interface MainNavProps extends FlexProps<'nav'> {
```

packages/strapi-design-system/src/v2/MainNav/NavBrand.tsx
```diff
@@ -0,0 +1,77 @@
...
+  icon: React.ReactNode;
+  title: string;
+  to?: string;
+  workplace: string;
+}
+
+export const NavBrand = React.forwardRef<HTMLAnchorElement, NavBrandProps>(
+  ({ workplace, title, icon, ...props }, ref) => {
+    const condensed = useMainNav();
+    props.to = props?.to ?? '/';
```
joshuaellis, line 37:
```suggestion
   /**
   * TODO: this shouldn't be here, because we're assuming you're
   * passing a ReactRouter Link which isn't necessarily the case.
   */
    props.to = props?.to ?? '/';
```

I know you didn't write this per say, but i want the comment so we remember to not do this 😅

packages/strapi-design-system/src/v2/MainNav/NavCondense.tsx
```diff
@@ -1,14 +1,13 @@
...
 import { ChevronRight, ChevronLeft } from '@strapi/icons';
-import PropTypes from 'prop-types';
 import styled from 'styled-components';
 
 import { useMainNav } from './MainNavContext';
 import { Icon } from '../../Icon';
 import { VisuallyHidden } from '../../VisuallyHidden';
 
-const NavCondenseWrapper = styled.button`
+const NavCondenseWrapper = styled.button<{ condensed: boolean }>`
```
joshuaellis, line 10: Can we please refactor this to use `Flex`? A lot of this css could be moved to props instead imo.

joshuaellis (changes requested):
Few minor tweaks, looks good though! :)

joshuaellis (approved):
Nice job!

PR #1080, opened 2023-05-24
Convert Table to Typescript

Minecraftschurli:
### What does it do?
Converts Table to Typescript

### Related issue(s)/PR(s)


packages/strapi-design-system/src/Table/index.ts
```diff
@@ -3,4 +3,4 @@ export * from './Tbody';
 export * from './Thead';
 export * from './Tr';
 export * from './Cell';
-export * from './TFooter';
+export * from './TFooter.js';
```
joshuaellis, line 6:
```suggestion
export * from './TFooter';
```

joshuaellis (approved):

PR #1081, opened 2023-05-24
Fix: Use theme values for lineHeight

gu-stav:
### What does it do?

Fixes how `lineHeight` is applied to `Box` components. Instead of using theme values (based on index) it applied the value directly, which lead to `line-height: 4` instead of `line-height: ${theme.lineHeights[4]}`.

### Why is it needed?

Fixes `Textarea` styles.

### Related issue(s)/PR(s)

Broken since https://github.com/strapi/design-system/pull/1030

### DIFF
--- a/packages/strapi-design-system/src/Box/Box.tsx
+++ b/packages/strapi-design-system/src/Box/Box.tsx
@@ -25,7 +25,6 @@ export type BoxProps<TElement extends keyof JSX.IntrinsicElements = 'div'> = Rea
     | 'animation'
     | 'textAlign'
     | 'textTransform'
-    | 'lineHeight'
   > & {
     /**
      * JavaScript hover handler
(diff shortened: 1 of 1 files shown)


joshuaellis (approved):

gu-stav:
@joshuaellis I wonder what you think about this being a breaking change? Technically it is, because if you'd have passed in `1` it will result in different values afterwards. However: the cloud only has 2 places that would be affected: https://github.com/search?q=repo%3Astrapi%2Fcloud%20lineHeight&type=code The CMS has none: https://github.com/search?q=repo%3Astrapi%2Fstrapi%20lineHeight&type=code

Still: theoretically there is a chance this would break something, but I think would argue the previous implementation was wrong and therefore it is a bugfix?

joshuaellis:
Didn't the changes you made previously break the CMS? so therefore this is a fix and not a breaking change? 😄

gu-stav:
Yeah, but just looking at how `lineHeight` behaves in isolation: it will be different after this PR and the same values will result in a different line-height. I am a bit worried, but also willing to say it is a bugfix, because the previous behavior just didn't make sense.

PR #1043, opened 2023-05-12
Chore: Convert Layout to Typescript

gu-stav:
### What does it do?

Chore: Convert `Layout` to Typescript

### Why is it needed?

https://github.com/strapi/design-system/issues/983

PR #1087, opened 2023-05-26
Chore: ActionLayout - fix actions condition

gu-stav:
### What does it do?

Fixes rendering of the `ActionLayout`.

### Why is it needed?

In https://github.com/strapi/design-system/pull/1043 I've made a mistake and converted:

```
startActions || endActions ? (<Component />) : null
```

to

```
if (!startActions || !endActions) {
  return null;
}
```

which is not the same. The operator should be AND.

This will help to fix the broken tests for Audit Logs in https://github.com/strapi/strapi/pull/16843


HichamELBSI (approved):
👍🏽

HichamELBSI:
We should have some test for this case at some point btw :)

gu-stav:
@HichamELBSI Yes I agree and will make sure to write some as a follow-up :)

Issue #1079, opened 2023-05-24
[bug]: Button startIcon and endIcon are too big


HichamELBSI:
### What version of strapi-design-system are you using?

Strapi Design System 1.7.7

### What's Wrong?

We have noticed that the size of the icons used for the `startIcon` and `endIcon` Button props has been changed. 

### To Reproduce

- Open the [Button with icons story](https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-button--icons)
- The icons seems to be rendered at a size of `1rem`

<img width="632" alt="Screenshot 2023-05-24 at 11 25 26" src="https://github.com/strapi/design-system/assets/7756284/6a759ce5-9953-4dfb-b36c-6c21ab8eeacc">

### Expected Behaviour

The expected size of the icons is  `0.75rem`

gu-stav:
Fixed by https://github.com/strapi/design-system/pull/1089

PR #1089, opened 2023-05-26
Fix: Button start and end icon size

gu-stav:
### What does it do?

Fixes the size of icons in `Button` components.

### Why is it needed?

It lead to alignment issues and doesn't follow the UI Kit.

### How to test it?

|Before|After|
|-|-|
| Screenshot 2023-05-26 at 15 50 30 | Screenshot 2023-05-26 at 15 50 13 |

### Related issue(s)/PR(s)

- Broke in https://github.com/strapi/design-system/pull/878
- Fixes https://github.com/strapi/design-system/issues/1079

### DIFF
--- a/packages/strapi-design-system/src/Button/Button.tsx
+++ b/packages/strapi-design-system/src/Button/Button.tsx
@@ -19,13 +19,18 @@ const rotation = keyframes`
   }
 `;
 
-const LoadingWrapper = styled.div`
+const LoaderAnimated = styled(Loader)`
   animation: ${rotation} 2s infinite linear;
   will-change: transform;
 `;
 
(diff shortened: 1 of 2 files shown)


packages/strapi-design-system/src/Button/Button.tsx
```diff
@@ -19,14 +19,19 @@ const rotation = keyframes`
   }
 `;
 
-const LoadingWrapper = styled.div`
+const LoaderAnimated = styled(Loader)`
```
markkaylor, line 22: 👍 nice

markkaylor (approved):
looks good to me

Issue #1096, opened 2023-06-01
CarouselSlide component is not working


smoothdvd:
### Required System information

- Node.js version: 18.16.0
- NPM version: pnpm 8.6.0
- Strapi version: 4.10.7
- Database: PostgreSQL 15
- Operating system: Linux
- Is your project Javascript or Typescript: Typescript

### Describe the bug

CarouselSlide component is not working

### Steps to reproduce the behavior

1. Create a collection which has a field with media ( multiple images) type
2. add 50 images
3. See error

### Expected behavior

CarouselSlide component is working correctly ( size and layout)

### Screenshots

<img width="726" alt="Screenshot 2023-06-01 at 4 53 35 PM" src="https://github.com/strapi/strapi/assets/22420/562e652b-8a28-4ca8-bf9a-8b5ee89aff07">

gu-stav:
Fix has been released in https://github.com/strapi/design-system/releases/tag/v1.7.10. It will be available with 4.11 next Wednesday in the CMS. 

Bump PR in the CMS: https://github.com/strapi/strapi/pull/16899

Thanks for reporting the issue.

PR #1052, opened 2023-05-16
Chore: Convert CarouselInput to Typescript

gu-stav:
### What does it do?

Convert `CarouselInput` to Typescript.

### Why is it needed?

https://github.com/strapi/design-system/issues/983

PR #1097, opened 2023-06-01
Fix: CarouselInput - fix slide rendering

gu-stav:
### What does it do?

- Adds new example images for the carousel - when using the same for every slide the effect is not visible
- Extends `Flex` to receive a `display` prop, which defaults to `flex`; this should have been possible according to the types, but wasn't
- Fixes https://github.com/strapi/design-system/issues/1096: it broke because the `CarouselSlide` component never received `display: none`
- `CarouselImage`s were also not centered anymore

Screenshot 2023-06-01 at 17 47 30

### Why is it needed?

Fixes https://github.com/strapi/design-system/issues/1096

### How to test it?

https://design-system-git-fix-carousel-input-strapijs.vercel.app/?path=/story/design-system-components-carouselinput--base

### Related issue(s)/PR(s)

- Fixes https://github.com/strapi/design-system/issues/1096
- Broke in https://github.com/strapi/design-system/pull/1052


markkaylor (approved):
LGTM

jhoward1994 (approved):

PR #1114, opened 2023-06-06
Chore: Update MainNav to Typescript

gu-stav:
### What does it do?

No refactoring, because the component is not used anymore. Converts the component to Typescript.

### Why is it needed?


joshuaellis (approved):
nice job, shame it's not used anymore

PR #1118, opened 2023-06-07
Chore: Convert SubNav to Typescript

gu-stav:
### What does it do?

Converts `SubNav` to Typescript. No refactoring because the component is not in use anymore.

### Why is it needed?


packages/strapi-design-system/src/SubNav/SubNavHeader.tsx
```diff
@@ -14,17 +13,39 @@ import { IconButton } from '../IconButton';
 import { Searchbar, SearchForm } from '../Searchbar';
 import { Typography } from '../Typography';
 
+export interface SubNavHeaderProps {
+  as: keyof HTMLElementTagNameMap;
```
joshuaellis, line 17: This is interesting, I didn't know about this type – where did you learn it? Should it be applied elsewhere in the codebase? :)

packages/strapi-design-system/src/SubNav/SubNavHeader.tsx
```diff
@@ -14,17 +13,39 @@ import { IconButton } from '../IconButton';
...
+  id: string;
+  label: string;
+  onChange: ChangeEventHandler<HTMLInputElement>;
```
joshuaellis, line 20: you've set a lot of these props as required, yet you've set a default? I think if they're required, I wouldn't set a default – personally i opt to make them optional, but that's really splitting hairs so it doesn't matter which you prefer.

packages/strapi-design-system/src/SubNav/SubNavSection.tsx
```diff
@@ -9,6 +8,14 @@ import { Box } from '../Box';
 import { Flex } from '../Flex';
 import { useId } from '../hooks/useId';
 
+export interface SubNavSectionProps {
+  badgeLabel: string;
+  children: React.ReactNode;
+  collapsable: boolean;
```
joshuaellis, line 14: same here with required yet setting defaults

packages/strapi-design-system/src/SubNav/SubNavHeader.tsx
joshuaellis: I think it's incorrect looking at the typings for `styled-components`:

```ts
type StyledComponentPropsWithAs<
    C extends string | React.ComponentType<any>,
    T extends object,
    O extends object,
    A extends keyof any,
    AsC extends string | React.ComponentType<any> = C,
    FAsC extends string | React.ComponentType<any> = C
> = StyledComponentProps<C, T, O, A, FAsC> & { as?: AsC | undefined; forwardedAs?: FAsC | undefined };
```

Maybe all `as` calls should be `StyledComponentPropsWithAs['as']`? Also, it definitely is not required imo

gu-stav: I've updated it to `string | React.ComponentType<any>`. I still think `keyof HTMLElementTagNameMap | React.ComponentType<any>` would be stricter, but keeping it inline with styled-components sounds good to me.

joshuaellis (approved):

PR #1120, opened 2023-06-07
Chore: Fix type generation

gu-stav:
### What does it do?

- Fixes the `generate:types` tsc command; `--declarationDir` requires `--declaration`
- Fixes the return type of the test util; before it failed with "The inferred type of 'render' cannot be named without a reference to '@testing-library/dom/node_modules/pretty-format'. This is likely not portable. A type annotation is necessary."

### Why is it needed?

Paves the path towards published types and allows me to learn.

### How to test it?

Run `yarn generate:types` in the design-system package.

### DIFF
--- a/packages/strapi-design-system/package.json
+++ b/packages/strapi-design-system/package.json
@@ -54,7 +54,7 @@
     "build": "yarn build:prod",
     "build:prod": "vite build",
     "clean": "rimraf dist node_modules",
-    "generate:types": "tsc --noEmit false --emitDeclarationOnly --declarationDir dist",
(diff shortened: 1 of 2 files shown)


joshuaellis:
> Fixes the return type of the test util; before it failed with "The inferred type of 'render' cannot be named without a reference to '@testing-library/dom/node_modules/pretty-format'. This is likely not portable. A type annotation is necessary."

wooo well done.

> Fixes the generate:types tsc command; --declarationDir requires --declaration

I think we'll probably opt to use the same plugin we use in `primitives` when we're ready fyi

PR #1139, opened 2023-06-14
fix: Pass the 'open' prop to ComboboxPrimitive.Root in the DS Combobo…

boazpoolman:
…x component

### What does it do?

Passes the `internalIsOpen` value as the `open` prop for the `<ComboboxPrimitive.Root>` component in the design-system's main `<ComboboxInput>` component.

### Why is it needed?

So I can controll the open state when I implement the Combobox component

### How to test it?

Implement the Combobox component as described here:
https://design-system-git-main-strapijs.vercel.app/?path=/docs/design-system-components-combobox--basic

Pass the `open` prop to try and open the combobox programatically.

With this PR change it works properly, without it will not open the combobox.


boazpoolman:
Do you guys have a timeline on when 1.8.1 will be released? :)

Issue #1090, opened 2023-05-26
[bug]: Breadcrumbs (v2) abbrevation menu is broken


gu-stav:
### What version of strapi-design-system are you using?

- Strapi design system version: 1.7.9

### What's Wrong?

The menu when breadcrumbs are abbreviated looks broken:

<img width="251" alt="Screenshot 2023-05-26 at 15 53 57" src="https://github.com/strapi/design-system/assets/2244375/f654a6bc-45c5-4a71-b851-ab0622494232">

### To Reproduce

See http://localhost:6006/?path=/story/design-system-components-v2-breadcrumbs--with-menu

### Expected Behaviour

I expect to look the menu to look good.

PR #1150, opened 2023-06-21
fix: Breadcrumbs (v2) abbrevation menu is broken

simotae14:
### What does it do?

It fixes the problem we had inside the Breadcrumb v2 
Schermata 2023-06-21 alle 14 24 40
The CarretDown icon is an error based on the Mockups
Schermata 2023-06-21 alle 14 26 00
You can find them here
https://www.figma.com/file/doUJzYdMJTjpgtw6Lyaa7D/%5B%E2%9C%85Done%5D-Content-squad---Media-folders?type=design&node-id=1741-30931&t=6Xu5lry10tCSaxRB-0

### Why is it needed?

Because in prod the actual implementation is not the desired one

### How to test it?

You can just check the storybook generated (v2 Breadcrumb with menu) 

### Related issue(s)/PR(s)
https://github.com/strapi/design-system/issues/1090

### DIFF
--- a/packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
+++ b/packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
@@ -27,23 +27,25 @@ const MenuRoot = DropdownMenu.Root;
 
 interface TriggerProps extends ButtonProps {}
 
-const MenuTrigger = forwardRef<HTMLButtonElement, TriggerProps>(({ size, ...props }, ref) => {
-  return (
-    <DropdownMenu.Trigger asChild>
-      <Button
-        ref={ref}
-        type="button"
-        variant="ghost"
-        endIcon={<CarretDown width={`${6 / 16}rem`} height={`${4 / 16}rem`} aria-hidden />}
-        paddingTop={size === 'S' ? 1 : 2}
-        paddingBottom={size === 'S' ? 1 : 2}
-        paddingLeft={size === 'S' ? 3 : 4}
-        paddingRight={size === 'S' ? 3 : 4}
-        {...props}
-      />
-    </DropdownMenu.Trigger>
-  );
-});
+const MenuTrigger = forwardRef<HTMLButtonElement, TriggerProps>(
+  ({ size, endIcon = <CarretDown width={`${6 / 16}rem`} height={`${4 / 16}rem`} aria-hidden />, ...props }, ref) => {
+    return (
+      <DropdownMenu.Trigger asChild>
+        <Button
+          ref={ref}
+          type="button"
+          variant="ghost"
+          endIcon={endIcon}
+          paddingTop={size === 'S' ? 1 : 2}
+          paddingBottom={size === 'S' ? 1 : 2}
+          paddingLeft={size === 'S' ? 3 : 4}
+          paddingRight={size === 'S' ? 3 : 4}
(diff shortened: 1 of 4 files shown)


joshuaellis (changes requested):
The padding feels off with the button on hover. It makes the button too big, are the values correct?

simotae14:
I have changed the previous values of the padding, let me know, @joshuaellis, if you think it is better the situation now based on the Figma document, thanks

packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
```diff
@@ -25,16 +25,18 @@ const MenuRoot = DropdownMenu.Root;
  * MenuTrigger
  * -----------------------------------------------------------------------------------------------*/
 
-interface TriggerProps extends ButtonProps {}
+interface TriggerProps extends ButtonProps {
+  noIcon?: boolean;
```
gu-stav, line 29: I don't know what the others think, but I am not a big fan of this prop, because it doesn't scale well.

What do you think about passing in `endIcon` instead? It could be set to `<CarretDown width={`${6 / 16}rem`} height={`${4 / 16}rem`} aria-hidden />` by default, and we allow overwriting it from outside?

simotae14: I understand your point, and what do you suggest to pass from the CrumbSimpleMenu component as value for the endIcon? null?

gu-stav: Yes :)

remidej (commented):
Hey Simone. Thanks for looking at this issue, I appreciate the attention to details :)

The code changes LGTM! I just unfortunately can't do QA as you asked me to, because I'm running into an error when trying to build the project:
Screenshot 2023-06-30 at 14 12 34

It's likely an error in my local setup, but unfortunately I don't have time to investigate it before leaving for the weekend, sorry about that

simotae14:
Don't worry @remidej  and thanks for your check...I will ask someone else to do a quick QA on this bugfix...enjoy your weekend

packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
```diff
@@ -27,14 +27,16 @@ const MenuRoot = DropdownMenu.Root;
...
+const MenuTrigger = forwardRef<HTMLButtonElement, TriggerProps>(({ size, endIcon, ...props }, ref) => {
   return (
     <DropdownMenu.Trigger asChild>
       <Button
         ref={ref}
         type="button"
         variant="ghost"
-        endIcon={<CarretDown width={`${6 / 16}rem`} height={`${4 / 16}rem`} aria-hidden />}
+        endIcon={
+          endIcon !== undefined ? endIcon : <CarretDown width={`${6 / 16}rem`} height={`${4 / 16}rem`} aria-hidden />
```
joshuaellis, line 38:
```suggestion
          endIcon
```

joshuaellis:
The padding was still wrong and I think you misunderstood Gustav's comment RE. default values. I've solved this to move this ticket along.

joshuaellis (approved):

simotae14:
ok, thank you @joshuaellis

PR #1157, opened 2023-06-23
fix: subcontent overflow

### LINKED ISSUE #1142: DynamicZone: Add component above/below list overflows container

- Node.js version: v18.14.1
- NPM version: 9.3.1
- Strapi version: @strapi/strapi@4.11.1
- Database: Postgres
- Operating system: debian bullseye
- Is your project Javascript or Typescript: javascript

### Describe the bug

In dynamic zones, we can now (since https://github.com/strapi/strapi/pull/16826 afaik) add components above and below existing components, which is very helpful, thanks for that! 
The popover element which shows available components has a height limited to `15rem`, but no scrollbars. The list of components therefore breaks the layout and overflows the container. 

1. Add a component to a DynamicZone
2. Click Button labeled "More actions" (Three horizontal dots icon)
3. Hover over either "Add component above" or "Add component below"
4. If you have more than ~6 components (depending on groups), the list will overflow the containing popper/Popover element

### Expected behavior

Two options:
- Do not limit the height of the Popover container
- Provide scrollbars for the list of components

### Screenshots

strapi--add-component-above

joshuaellis:
### What does it do?

* add overflow="auto" to Menu.SubContent

### Why is it needed?

* Items broke out of container

### Related issue(s)/PR(s)

* resolves #1142

### DIFF
--- a/packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
+++ b/packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
@@ -285,6 +285,7 @@ const MenuSubContent = forwardRef<HTMLDivElement, SubContentProps>((props, ref) 
           maxHeight="15rem"
           padding={1}
           alignItems="flex-start"
+          overflow="auto"
           {...props}
         />
       </DropdownMenu.SubContent>
--- a/yarn.lock
+++ b/yarn.lock
@@ -3354,17 +3354,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@radix-ui/react-compose-refs@npm:1.0.0":
-  version: 1.0.0
-  resolution: "@radix-ui/react-compose-refs@npm:1.0.0"
-  dependencies:
-    "@babel/runtime": ^7.13.10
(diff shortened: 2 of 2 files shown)


joshuaellis:
The bundle will fail because its a broken on the target branch.

joshuaellis:
> Does that mean the menu would display a vertical scrollbar? 🤔 I am wondering if we should include e.g. Lucas here, because I imagine the UX to no be great and maybe we should thrive for a better long-term solution, because the list of components might be long in many cases?

IMHO we shouldn't be capping the height of any popover but this is the design spec:

CleanShot 2023-06-26 at 10 15 04@2x

I think this is a subject for the DS sync?

gu-stav:
> 

I think I can live with that for non-nested Popovers, but I imagine it to be a bit hairy when using a nested menu. Could either be an issue for a DS Sync or a topic for @lucasboilly. I think since he isn't part of the sync meeting (yet) it makes sense to prepare it with him?

gu-stav (approved):

PR #1165, opened 2023-06-27
fix: date picker typing localisation 

### LINKED ISSUE #1162: The datetime field is always displayed in DD/MM/YYYY format

- Node.js version: 18.16.0
- NPM version: 9.6.5
- Strapi version: 4.11.0
- Database: mysql8
- Operating system: debian
- Is your project Javascript or Typescript: Typescipt

### Describe the bug

The datetime field is always displayed in DD/MM/YYYY format

### Expected behavior

Display in language-appropriate format.

#### Definition

スクリーンショット 2023-06-08 17 56 32

#### Display

English
スクリーンショット 2023-06-08 17 51 53

Japanese
スクリーンショット 2023-06-08 17 51 23

### Additional context

Updated from strapi 4.10.2.
In 4.10.2, the format was YYYY/MM/DD.
We are primarily using Japanese and are not familiar with the DD/MM/YYYY format.

Perhaps this has been a problem since the beginning, but it has become difficult for us to understand.

joshuaellis:
### What does it do?

* Allows the typing of the date in any given locale
* Sets the correct placeholder text based on the locale
* Always updates the text value by the `formatter.format` to ensure the locale is respected

### Why is it needed?

* Locales like japanese where not correctly respected.

### How to test it?

* Automated tests added 🪄 

### Related issue(s)/PR(s)

* resolves #1162

### DIFF
--- a/packages/strapi-design-system/src/DatePicker/DatePicker.tsx
+++ b/packages/strapi-design-system/src/DatePicker/DatePicker.tsx
@@ -71,11 +71,6 @@ interface DatePickerContextValue {
   onTriggerChange: (trigger: DatePickerTriggerElement | null) => void;
   onValueChange: (value: CalendarDate | undefined) => void;
   required: boolean;
-  /**
-   * Localised separator for date values e.g. `/` if en-GB,
-   * think `os.separator` for dates
-   */
-  separator: string;
   textInput: DatePickerTextInputElement | null;
   textValue?: string;
   timeZone: string;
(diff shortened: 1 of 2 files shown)


packages/strapi-design-system/src/DatePicker/DatePicker.tsx
```diff
@@ -549,12 +571,49 @@ const DatePickerTextInput = React.forwardRef<DatePickerTextInputElement, TextInp
...
+                  return 1;
+                }
+
+                if (b.type === 'month') {
+                  return -1;
+                }
+
+                return 0;
+              })
+              .map((part) => part.value);
```
Marc-Roig, line 616: ⚠️ Backender commenting on a frontend PR ⚠️ 

This is more of a question:
I am normally used to split logic and UI, so you have the domain defined elsewhere with how a datepicker should work, and you use those functionalities in the UI definitions. When I am reading the UI , I normally try to understand the functionality not the implementation, and having both in the same place makes it a bit more difficult to follow.

Do you think it could be useful to split domain logic here? Or it would be harder to track where each thing goes to?

Also, using a domain forces you to think things in an abstract way.

Marc-Roig (approved):

Marc-Roig:
Looks good ! ✨

packages/strapi-design-system/src/DatePicker/DatePicker.tsx
joshuaellis: EXCELLENT QUESTION. 

Typically we don't do that, because we don't test internals and if we export the internals e.g. the logic to convert any date format e.g. YY/DD/MM to DD/MM/YY – we then have to test that independently but realistically we're more satisfied with testing how a user interacts with things, the implementation is not as important because 9/10 at this level we're rarely sharing such niche code & it would make our suite more brittle than we deem valued.

Also from my perspective I believe in deep modules (big files) with a small as possible API surface so people understand whats for public consumption etc.

How do you feel about that?

PR #1172, opened 2023-06-30
fix: date picker with some timezones would change date automatically

### LINKED ISSUE #1171: Datepicker issues (causes data corruption)

- Node.js version: 18
- Strapi version: 4.11.1
- Database: PostgreSQL
- Operating system: Redhat
- Is your project Javascript or Typescript: Javascript

### Describe the bug

Recent changes were made to the datepickers in 4.11.0 (https://github.com/strapi/strapi/pull/16805).  Our QA team has been doing testing to ensure that these changes are ready for production, and a critical bug was discovered.  

- If you click into a date field that already has a value, and you don’t change anything, the date **decrements** by 1 day every time the field loses focus.  
- If you select a date using the datepicker then the date **increments** by 1 day when the field loses focus.  
- These are specifically Date fields (not DateTime fields)  

It was also noted that the datepicker now goes from Saturday to Friday instead of Sunday to Saturday.  I know there has been a long debate about whether the first day of the week should be Sunday or Monday, but choosing Saturday as the first day of the week seems unique and arbitrary.

joshuaellis:
### What does it do?

* removes `intl` polyfill (not needed anymore, JSDOM has it)
* uses `getLocalTimeZone` for the timezone context
* uses timzone to format the date correctly to a UTC format

### Why is it needed?

* because the time was always `00:00` any timezone -GMT would go to the day before on update

### Related issue(s)/PR(s)

* resolves #1171

### DIFF
--- a/packages/strapi-design-system/src/DatePicker/__tests__/DatePicker.spec.tsx
+++ b/packages/strapi-design-system/src/DatePicker/__tests__/DatePicker.spec.tsx
@@ -344,7 +344,7 @@ describe('DatePicker', () => {
     it('should format by en locale', async () => {
       const { getByRole, user } = render({ selectedDate: new Date('Tue Sep 06 2022'), locale: 'en' });
 
-      expect(getByRole('combobox')).toHaveValue('9/6/2022');
+      expect(getByRole('combobox')).toHaveValue('09/06/2022');
 
       await user.click(getByRole('combobox', { name: 'date picker' }));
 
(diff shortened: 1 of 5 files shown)


packages/strapi-design-system/src/DatePicker/__tests__/DatePicker.spec.tsx
```diff
@@ -394,6 +394,48 @@ describe('DatePicker', () => {
 
       expect(getByRole('gridcell', { name: '2022年12月18日日曜日' })).toHaveAttribute('aria-selected', 'true');
     });
+
+    /**
+     * TODO: the spying doesn't work :(
+     */
+    it.skip('should not change the date automatically if the timezone is not UTC but is America/Los_Angeles', async () => {
```
jhoward1994, line 401: Is this test meant to be skipped? it seems to be working for me

joshuaellis: It still passes because the timezone is set to UTC by default inside the component.

jhoward1994 (approved):

PR #1177, opened 2023-07-04
chore: add changesets

joshuaellis:
### What does it do?

* adds changeset to the design-system
* removes lerna from the design-system
* converts scripts to use NX

### Why is it needed?

Maintaining release branches are hard, realistically we can move quicker if everything is merged to `main` and we use changesets to indicate the type of change we're adding, therefore no one needs to think about the right milestone to use, the right base branch to use nor the right target branch – we remove the hard work and focus on shipping.

We can additionally include the ["changeset bot"](https://github.com/apps/changeset-bot) to help enforce this practice of writing our own changesets.


gu-stav:
@joshuaellis Could you add a paragraph to the CONTRIBUTING file, so that I (and others) can know what the steps would be to create a new changeset?

joshuaellis:
> @joshuaellis Could you add a paragraph to the CONTRIBUTING file, so that I (and others) can know what the steps would be to create a new changeset?

Great idea!

gu-stav (approved):
Let's give it a try 🚀

PR #1179, opened 2023-07-05
fix: make label an optional prop

joshuaellis:
### What does it do?

* Makes it so you must pass either `label` or `aria-label` as a prop
* no longer renders the label element if `label` is not included

### Why is it needed?

* There are a lot of places where we don't have a visible label in the UI and because of how we want our components it still tries to render the `label` element which causes the flex gap to still be there, no we don't render that element – re-aligning a lot of the UI.

### How to test it?

* Automated testing to make sure it's still accessible with just an `aria-label` prop.

### Related issue(s)/PR(s)

* resolves CONTENT-1557

### DIFF
--- a/packages/strapi-design-system/src/Select/MultiSelect.tsx
+++ b/packages/strapi-design-system/src/Select/MultiSelect.tsx
@@ -15,7 +15,7 @@ import { useIntersection } from '../hooks/useIntersection';
 import { Tag } from '../Tag';
 import { Typography } from '../Typography';
 
-export type MultiSelectProps = Omit<SelectParts.MultiSelectProps, 'value' | 'multi'> &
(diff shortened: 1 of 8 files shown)


packages/strapi-design-system/src/Select/MultiSelect.tsx
```diff
@@ -164,7 +170,7 @@ export const MultiSelect = ({
         >
           <SelectParts.Trigger
             ref={triggerRef}
-            aria-label={label}
+            aria-label={label ?? ariaLabel}
```
gu-stav, line 173: Should this be the other way around, so that `ariaLabel` takes precedence over label in case both are passed in?

gu-stav (approved):

PR #1182, opened 2023-07-07
[Fake PR] Set checkbox color to blue

remidej:
### What does it do?

Some silly changes

### Why is it needed?

Showing @lucasboilly how to open a PR, will delete soon


packages/strapi-design-system/src/Checkbox/Checkbox.tsx
```diff
@@ -42,7 +42,7 @@ export const Checkbox = ({ children, disabled = false, id, hint, error, ...props
   return (
     <Field id={generatedId} hint={hint} error={error}>
       <Flex direction="column" alignItems="stretch" gap={1}>
-        <CheckboxLabel as="label" textColor="neutral800" disabled={disabled}>
+        <CheckboxLabel as="label" textColor="primary700" disabled={disabled}>
```
remidej, line 45: I would prefer purple

```suggestion
        <CheckboxLabel as="label" textColor="alternative700" disabled={disabled}>
```

PR #1183, opened 2023-07-07
refactor: ToggleInput should not render two labels

joshuaellis:
### What does it do?

* refactors the ToggleInput to not use ToggleCheckbox underneath so it doesn't render two labels
* deprecates ToggleCheckbox cause we don't need two

### Why is it needed?

* aria

### How to test it?

* I added a test suite for the component because it was ✨ missing ✨


packages/strapi-design-system/src/ToggleInput/ToggleInput.tsx
```diff
@@ -50,19 +182,27 @@ export const ToggleInput = ({
...
         <FieldError />
       </Flex>
     </FieldWrapper>
   );
 };
 
-ToggleInput.displayName = 'ToggleInput';
+const FieldWrapper = styled(Field)`
+  max-width: 320px;
+`;
```
gu-stav, line 200: Since `Field` is of type `Box` now I think you can pass maxWidth as a prop instead.

gu-stav (approved):

PR #1202, opened 2023-07-13
feat: SubNavHeader searchPlaceholder prop

### LINKED ISSUE #1201: [feat]: SubNavHeader search's placeholder can't be customized

Be able to set the `SubNavHeader` search field placeholder
https://github.com/strapi/design-system/blob/3ad7300bcb1c963efc9253e08debffd4adda8be0/packages/strapi-design-system/src/v2/SubNav/SubNavHeader.tsx#L88

The placeholder is currently statically set to `e.g: strapi-plugin-abcd` and is not really helpful

Add a `searchPlacholder` prop to the component and use it

ChristopheCVB:
### What does it do?

Add a `searchPlaceholder` prop to SubNavHeader component

### Why is it needed?

resolves #1201 

### Related issue(s)/PR(s)

#1201

### DIFF
--- a/packages/strapi-design-system/src/v2/SubNav/SubNavHeader.tsx
+++ b/packages/strapi-design-system/src/v2/SubNav/SubNavHeader.tsx
@@ -26,6 +26,7 @@ export interface SubNavHeaderProps {
   onClear?: React.MouseEventHandler<HTMLInputElement>;
   onSubmit?: React.FormEventHandler<HTMLInputElement>;
   searchLabel?: string;
+  searchPlaceholder?: string;
   searchable?: boolean;
   value?: string;
 }
@@ -34,6 +35,7 @@ export const SubNavHeader = ({
   as = 'h2',
   label,
   searchLabel = '',
+  searchPlaceholder = '',
   searchable = false,
   onChange = () => {},
   value = '',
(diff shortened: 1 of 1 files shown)


gu-stav (commented):
Thank you for this contribution! Would you be able to add a test for this case? 🙏🏼

ChristopheCVB:
Unfortunately, when I run `yarn setup` on the root project, I get some errors, here's one:
```
lerna notice cli v7.1.1


    ×  @strapi/ui-primitives:build
       vite v4.4.3 building for production...
src/components/Collection/Collection.tsx(24,10): error TS4060: Return type of exported function has or is using private name 'CollectionProviderProps'.

✓ 0 modules transformed.
✓ built in 1.72s
Entry module "src/index.ts" cannot be external.
error during build:
RollupError: Entry module "src/index.ts" cannot be external.
    at error (file:///C:/Users/chris/Documents/Personal/design-system/node_modules/rollup/dist/es/shared/node-entry.js:2245:30)
    at ModuleLoader.loadEntryModule (file:///C:/Users/chris/Documents/Personal/design-system/node_modules/rollup/dist/es/shared/node-entry.js:24677:20)
    at async Promise.all (index 0)


    ×  @strapi/icons:build
```

Also, I'm not a React developer, never dealt with tests in that environment :/
Do you have some tips?

gu-stav:
@ChristopheCVB Sorry to hear. I see you are running windows? Are you using WSL or something else? I think we have some known problems here and if you prefer I can add tests for you, because fixing the setup might take a bit.

ChristopheCVB:
Yeah, windows, no wsl, using power shell 7.
If you can add them for me, I'll be extremely grateful

PR #1241, opened 2023-07-31
Fix #1223: date time picker date changes automatically with timezone

### LINKED ISSUE #1223: [bug]: DateTimePicker changes date automatically (only for some timezones?)

- Strapi Design System version: 1.8.1

### What's Wrong?

When selecting a date in the DateTimePicker, that date minus one day is shown in the UI. If you click on the corresponding label, another day is subtracted from the date.

Possibly related to timezones and [this PR for the DatePicker](https://github.com/strapi/design-system/pull/1172/files#). I'm in the GMT+2 timezone. 

### To Reproduce

- Go to [the DateTimePicker in Storybook](https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-datetimepicker--base)
- Select today's date
- See that yesterday's date is show in the UI
- Click on the 'Publish at' label
- See that another day is subtracted in the UI

Possibly related to the users timezone. I'm in GMT+2

### Expected Behaviour

I expect the date I pick to be shown in the UI, and for it not to change when I click the label.

marob:
### What does it do?

Fix #1223

### Why is it needed?

This keeps the selected date in DateTimePicker instead of changing it (-1 day) on every change detection (this behavior was detected on UTC+2 timezone with time < 02:00, I suppose it's the same for UTC+X with time < X:00)

### How to test it?

Use UTC+X (with X>0) timezone, select a date (without time), on blur, the date changes to date - 1 day. Same behavior when selecting date and time < X:00.

### Related issue(s)/PR(s)

#1223


gu-stav:
Just a quick heads up: we will spend some time next week to fix most of the outstanding datepicker issues and review this PR.

Thank you for your contribution!

marob:
> Just a quick heads up: we will spend some time next week to fix most of the outstanding datepicker issues and review this PR.
> 
> Thank you for your contribution!

OK, nice.

@gu-stav Do you plan to try with different timezone in unit tests?
I made a quick search/test on that topic: it doesn't seem to be so easy to put in place (mocks, ...) but I think it would really improve components quality.

gu-stav:
> Do you plan to try with different timezone in unit tests?

I'm not sure what we are going to do both, because I won't be the one doing it. What we do in other parts of the CMS is to strip the TZ information for dates generated by clients.

Do you have a preference or need?

marob:
With this issue, my fix is: https://github.com/strapi/design-system/pull/1241/files#diff-8377105c3d9c5190078b42ed97a864902995af3a6ec5021b221c08fbe6f6174cR185
So, I replace `getLocalTimeZone()` by `'UTC'`.

There **is** [a test](https://github.com/strapi/design-system/blob/fd35533c43217d3ee202b39574ad81128e8baaf3/packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.spec.tsx#L107) that should have detected the issue, but as [all tests are run in UTC](https://github.com/strapi/design-system/blob/fd35533c43217d3ee202b39574ad81128e8baaf3/test/global-setup.js#L2), `getLocalTimeZone()` and `'UTC'` are the same in test execution context. If the test was run in another timezone (UTC+x), it should have detected the issue.
So what I'm suggesting is finding some mock library dedicated to unit tests that enables to set the timezone during the test. In that way, we could make some tests execute the same code and assertions in different timezones in a loop with a different timezone at every execution or duplicate tests with the same code (except we set different timezones) and assertions.

Of course, we should do those multi-timezones tests on every date-related tests.

What do you think?

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.spec.tsx
joshuaellis, line 1: These changes feel somewhat superfluous, can you explain why you think they're necessary?

joshuaellis (changes requested):

joshuaellis:
I've confirmed the fix works and doesn't have any regressions – thanks for that contribution, we do have a test that was looking at avoiding these sorts of regressions here – https://github.com/strapi/design-system/blob/525c935f32a6cd86fbea090a59c94f83790c5bbf/packages/strapi-design-system/src/DatePicker/__tests__/DatePicker.spec.tsx#L401 

serving a similar purpose to what you're describing, specifically around the fact we need to have the ability to mock `Intl.DateTimeFormat.resolvedOptions` but I couldn't get it to work:

```ts
      let resolvedOptions = Intl.DateTimeFormat.prototype.resolvedOptions;

      const resolvedOptionsSpy = jest.spyOn(Intl.DateTimeFormat.prototype, 'resolvedOptions');

      resolvedOptionsSpy.mockImplementation(function (this: Intl.DateTimeFormat) {
        let s = resolvedOptions.call(this);
        s.timeZone = 'America/Los_Angeles';

        return s;
      });
```

If you have any ideas on how to do it, i'd love to see them, ideally i don't want to use an external library as this _should_ be enough.

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.spec.tsx
marob: It's not required, but american format is understood only by americans, while yyyy/mm/dd is understood by everyone and is not susceptible to be misinterpreted (which is clearly not the case of american one).
I was suggesting this change for readability (and it should be applied in every test).

joshuaellis: I think we can omit it from this PR imo, let's keep it focussed – I feel like opening this discussion has a much wider impact than just this test file (as you've pointed out) :)

marob:
@joshuaellis Also, whatever the mock system we put in place, I think `@internationalized/date` will pose problem.

Indeed, according to the code of [`getLocalTimeZone()`](https://github.com/adobe/react-spectrum/blob/6d3277337aa875dcb176a5f6a3f1e380875a94fc/packages/%40internationalized/date/src/queries.ts#L121), the timezone is cached in a `localTimeZone` variable and we don't have any way to reset/change it.
So as soon as the `getLocalTimeZone()` has been called once, we won't be able to change the timezone returned by `getLocalTimeZone()` without having to reload the `@internationalized/date` somehow.
And I don't see how to do it without having to run the whole test suite one time per timezone...

packages/strapi-design-system/src/DateTimePicker/__tests__/DateTimePicker.spec.tsx
marob: I've removed the corresponding commit

joshuaellis (approved):

PR #1266, opened 2023-08-08
fix: date picker & time picker input patterns

### LINKED ISSUE #1233: BROKEN FILTER: Date Regexp never match

- Node.js version: v18
- NPM version: latest
- Strapi version: 4.11.7
- Database: postgres
- Operating system: linux
- Is your project Javascript or Typescript: typescript
- Browser: latest Chrome, latest Firefox, latest Edge

### Describe the bug

Admin UI filter doesn't accept any date. It says the input should match the pattern. 

1. Go to Content Manager
2. Click on Filter
3. Try to filter a date
4. See pattern match error

### Expected behavior

If I enter a date or select it from the popup date selector it should be a valid input for the date filter

### Screenshots

image

### Additional context

I think the pattern is invalid.
What we have now in the strapi code: \d2/\d2//d4

The / should be escaped with \ and the digits should be enclosed in {}

I suggest this pattern:
\d{2}\/\d{2}\/\d{4}

joshuaellis:
### What does it do?

* correctly escapes the separator in the input pattern
* Adds an automated test that checks the `validity.patternMismatch` is falsey when you use the input as standard e.g. clicking an option

### Why is it needed?

* Users would get form validation failures

### How to test it?

* automated testing 🪄 

### Related issue(s)/PR(s)

* resolves #1233
* resolves CONTENT-1639


gu-stav (approved):
💅🏼

PR #1268, opened 2023-08-08
feat: add forwardRefs to components

joshuaellis:
### What does it do?

* adds `forwardRef` to many components – typically the ones we use as `inputs` in the Content Manager

### Why is it needed?

* so we can auto-focus them for visual editing


gu-stav (approved):
💅🏼

PR #1272, opened 2023-08-10
chore: use strapi/eslint config for ui-primitives

joshuaellis:
### What does it do?

* uses `@strapi/eslint-config@0.2.0` for the FE typescript

### Why is it needed?

* create unity across projects

### How to test it?

* lint github action should pass


gu-stav (approved):
Very nice 💅🏼

PR #1273, opened 2023-08-11
Add a flag to the build config of the design-system to publish types

simotae14:
### What does it do?

Passing a env var we can publish types in the DS

### Why is it needed?

We want to convert the helper-plugin and to do it we need the types from the design-system.
We can’t publish a version of this because it will be a breaking change, instead we can use an ENV flag to produce the types on demand and use yarn link to connect the design-system to the monorepo.

### How to test it?

Run the yarn build with a env var 
`DTS=true yarn build`
Then, as said, with yarn link you can connect your local DS with the CMS

### DIFF
--- a/packages/strapi-design-system/vite.config.ts
+++ b/packages/strapi-design-system/vite.config.ts
@@ -4,6 +4,7 @@ import typescript from '@rollup/plugin-typescript';
 import react from '@vitejs/plugin-react';
 import glob from 'tiny-glob';
 import { defineConfig } from 'vite';
+import dts from 'vite-plugin-dts';
 
 export default defineConfig(async () => {
   const paths = await glob('./src/**/!(*.spec|*.e2e|*.test).{js,svg,ts,tsx}');
@@ -37,6 +38,13 @@ export default defineConfig(async () => {
         plugins: [typescript()],
       },
     },
-    plugins: [react()],
+    // We need to pass entryRoot: 'src' as an argument to dts because otherwise when we run the build,
+    // a src folder is generated inside the dist folder containing all the d.ts files.
+    // This is due to how the tsconfig.json file was defined, specifically this part:
+    //  "baseUrl": ".",
+    //  "paths": {
+    //    "@test/*": ["./test/*"]
+    //  }
+    plugins: process.env.DTS !== 'true' ? [react()] : [dts({ entryRoot: 'src' }), react()],
   };
 });
--- a/packages/strapi-icons/package.json
+++ b/packages/strapi-icons/package.json
@@ -25,11 +25,10 @@
     "react-dom": "^17.0.0 || ^18.0.0"
   },
   "scripts": {
-    "build": "yarn generate:icons && yarn build:prod && yarn generate:types",
+    "build": "yarn generate:icons && yarn build:prod",
     "build:prod": "vite build",
     "clean": "rimraf src dist node_modules",
-    "generate:icons": "svgr -- ./assets/icons",
-    "generate:types": "tsc --noEmit false --emitDeclarationOnly --declarationDir dist"
+    "generate:icons": "svgr -- ./assets/icons"
   },
   "gitHead": "c74900b0ee3525510d266dc83c9743cb24dafced"
 }
--- a/packages/strapi-icons/vite.config.ts
+++ b/packages/strapi-icons/vite.config.ts
@@ -1,6 +1,7 @@
 import { resolve } from 'path';
 import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
+import dts from 'vite-plugin-dts';
 
 export default defineConfig({
   build: {
@@ -22,5 +23,5 @@ export default defineConfig({
       },
     },
   },
-  plugins: [react()],
+  plugins: [dts(), react()],
 });
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,6 +114,12 @@ In your application, link your local copy of the design system with [`yarn link`
 yarn link -r ../<relative-path-to-strapi-design-system>
 ```
 
+You can also link a local copy of a specific package. For example, if you want to link the package strapi-design-system, you can run:
+
+```
+yarn link -r ../<relative-path-to-strapi-design-system>/packages/strapi-design-system
+```
+
 You should also remove the webpack alias for `@strapi/design-system` in the Strapi monorepo at `packages/core/admin/webpack.alias.js`
 
 Your application should now be using your local copy of the design system.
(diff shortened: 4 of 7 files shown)


packages/strapi-design-system/vite.config.ts
```diff
@@ -15,7 +17,7 @@ export default defineConfig(async () => {
       exclude: [],
     },
     build: {
-      emptyOutDir: false,
+      emptyOutDir: env.DTS === 'true',
```
joshuaellis, line 20: How come we needed to do this?

joshuaellis (changes requested):
You don't need to use `loadEnv` to load ENV variables – where did you read this? 😄 

You've left `vite-plugin-dts` just in the package.json of `@strapi/ui-primitives`, we currently hold `vite` globally – do you think those two packages should be scoped locally to our packages or remain at the global level?

`@strapi/icons` is still running it's own `yarn generate:types` function so the DTS plugin is effectively redundant at this point so this should be resolved.

Please ask any questions if you're unsure! 🌞

simotae14:
Thanks @joshuaellis I followed instructions to add env var inside the vite config https://stackoverflow.com/questions/66389043/how-can-i-use-vite-env-variables-in-vite-config-js but if there is a better way let me know about it, thanks.
You are right, I missed to add vite-plugin-dts in the other packages, based on your suggestion we can add it just in the packages where we need it, like you did for primitives?

packages/strapi-design-system/vite.config.ts
simotae14: I added this line of code because in the previous version, the value was set to false. Every time I ran "yarn build" after running "DTS=true yarn build", the dist folder would contain the old build result with the environmental variable. Therefore, my goal is to always have a completely new version in the dist folder when I run "DTS=true yarn build"

packages/strapi-icons/vite.config.ts
```diff
@@ -18,7 +18,7 @@ export default defineConfig({
       external: (id) => !id.startsWith('.') && !id.startsWith('/'),
       output: {
         dir: 'dist',
-        preserveModules: true,
+        preserveModules: process.env.DTS !== "true",
```
joshuaellis, line 21: You can't make this change because it'll break the current API where people can still do `import { X } from '@strapi/icons/X`

IMO I would just use the dts plugin and remove how it currently works to generate the types, im not sure why it's this way at the moment

packages/strapi-design-system/vite.config.ts
```diff
@@ -15,7 +16,7 @@ export default defineConfig(async () => {
       exclude: [],
     },
     build: {
-      emptyOutDir: false,
+      emptyOutDir: process.env.DTS === 'true',
```
joshuaellis, line 19: I think this will break storybook when you are developing i.e. you'll make a change in the DS and it'll empty the folder and storybook will crash.

packages/strapi-design-system/vite.config.ts
```diff
@@ -37,6 +38,6 @@ export default defineConfig(async () => {
         plugins: [typescript()],
```
joshuaellis, line 38: I'm not sure you need this plugin when `DTS === true` as the `dts` plugin will handle this instead.

packages/strapi-design-system/vite.config.ts
```diff
@@ -37,6 +38,6 @@ export default defineConfig(async () => {
...
       },
     },
-    plugins: [react()],
+    plugins: process.env.DTS !== 'true' ? [react()] : [dts({ entryRoot: 'src' }), react()],
```
joshuaellis, line 41: why do you have to declare `entryRoot`? we don't in the `primitives` package? :)

joshuaellis (changes requested):

joshuaellis:
I'll review this all again when all the checks pass 😄

packages/strapi-design-system/vite.config.ts
simotae14: Because otherwise it creates an src folder inside the dist folder with all the d.ts files

packages/strapi-icons/vite.config.ts
simotae14: ok

packages/strapi-design-system/vite.config.ts
joshuaellis: Really? that's weird, I would write a comment to highlight this then, because it's not clear.

simotae14: and it is because of this line of code in the tsconfig   https://github.com/strapi/design-system/blob/1d6bc42b4f9a0798db39b25308fea68f8f86e9e1/packages/strapi-design-system/tsconfig.json#L19
```
"baseUrl": ".",
    "paths": {
      "@test/*": ["./test/*"]
    }
```

packages/strapi-design-system/vite.config.ts
simotae14: ok

packages/strapi-design-system/vite.config.ts
simotae14: If I remove the typescript plugin with DTS true then I have this error when I try to build
typescript plugin

packages/strapi-design-system/vite.config.ts
joshuaellis: > I would write a comment to highlight this then

see previous comment 👍🏼

joshuaellis (approved):

packages/strapi-icons/package.json
```diff
@@ -25,7 +25,7 @@
     "react-dom": "^17.0.0 || ^18.0.0"
   },
   "scripts": {
-    "build": "yarn generate:icons && yarn build:prod && yarn generate:types",
```
madhurisandbhor, line 28: Can we add DTS=true to "build"/"generate:types" script to generate types directly from command?
e.g: "build": "yarn generate:icons && cross-env DTS=true yarn build:prod",

Also we are not using generate:types script anymore, should we remove/replace it?

packages/strapi-icons/vite.config.ts
```diff
@@ -22,5 +23,5 @@ export default defineConfig({
       },
     },
   },
-  plugins: [react()],
+  plugins: process.env.DTS !== 'true' ? [react()] : [dts(), react()],
```
joshuaellis, line 26: see other comment, we can just use the plugin, there's no need for a flag because we were already publishing types.

joshuaellis (changes requested):

packages/strapi-icons/vite.config.ts
simotae14: ok

madhurisandbhor (approved):
Nice 👍

joshuaellis (approved):

joshuaellis:
@simotae14 as per the contributing guidelines, please squash your commits before merging 👍🏼

PR #1282, opened 2023-08-14
chore: upgrade to Storybook7

### LINKED ISSUE #1071: [bug]: Storybook webpack not reloading after vite bundles primitives package

Strapi Design System v1.7.7

### What's Wrong?

Storybook's Webpack is not reloading after Vite bundles our @strapi/primitives package. This causes errors during local development, which can be frustrating.

Screenshot 2023-05-22 at 12 17 32

### To Reproduce

1. Run the local dev environment (follow the development workflow of the contributing guide)
2. Make a change on one of the primitive components.
3. The Storybook live reload may trigger some errors related to the primitive package bundle.

### Expected Behaviour

We expect live reload to work without any issues after making changes to any of the packages 😄

joshuaellis:
### What does it do?

* Updates storybook to v7
* Updates storybook to use typescript
* Updates storybook to use vite
* Updates storybook setup so we're aliasing src, no more waiting around for build times to compile ✅ 
* Updates storybook to use CSF3.0 & mdx entries
* Updates playwright to 1.37.0
* Updates e2e tests to await the story title to appear
* Applies `@strapi/eslint-config` to docs

### Why is it needed?

* updates are good
* The vite version of storybook dynamically loads the stories and as such on a longer wait from the server the test will fail, we now actually do some assertion which has a standard timeout procedure should there be an error
* CSF3.0 is written in TSX which means we can actually see the issues in our stories, this was a leading factor in how long this upgrade has taken, the eslint config for it all was poor.

### Related issue(s)/PR(s)

* resolves #1071


joshuaellis:
@gu-stav this is now definitely ready for review 👍🏼

gu-stav (approved):
Very readable, very clean & blazing fast updates. I like it.

PR #1283, opened 2023-08-14
fix(Combobox): check children to display no options message

ayhandoslu:
Checking only `!creatable && !loading` should not be enough to display the "No options found" message.

If any children exist, unfortunately, it's displaying "No options found" after the children on the list.


joshuaellis:
I don't think your fix is the correct approach, you're assuming people omit the children from the react tree in response to the filtering abilities. Realistically we most likely need a more thorough check in the `NoValueFound` component in the respected primitive.

PR #1285, opened 2023-08-16
chore: add commitlint

joshuaellis:
### What does it do?

* adds commitlint, config & husky hook
* updates contributing doc
* separates code of conduct doc

### Why is it needed?

* commitlint standardises our approach to commit messages making the history more useful


gu-stav (approved):
Let's try it out and see where it takes us.

PR #1297, opened 2023-08-17
feat(icons): First SVG coloring API idea

gu-stav:
### What does it do?

This is the first idea for what an SVG coloring API could look like using an svgr template.

#### Single color icons

Most icons use one color. We would have to remove the hard-coded colors and replace them with `currentColor`:

```html
<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
    <path d="...." fill="currentColor"/>
</svg>
```

Usage:

```jsx
import { Icon } from '@strapi/icons'

function Component() {
  return (
    <Icon color="primary100">
  )
}
```

#### Multi color icons

For icons using multiple colors we could use a similar approach with named colors using CSS custom properties in the background:

```html
<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
    <style>
     #logoRocket {
       fill: var(--color-logo-rocket, red);
     }
    </style>

    <path id="logoRocket" d="...." />
</svg>
```

Usage:

```jsx
import { Icon } from '@strapi/icons'

function Component() {
  return (
    <Icon colors={{ 'logo-rocket': 'primary100' }}>
  )
}
```

### Why is it needed?

Explore paths to remove default colors from SVGs.

### DIFF
--- a/packages/strapi-icons/package.json
+++ b/packages/strapi-icons/package.json
@@ -29,7 +29,7 @@
     "build": "yarn generate:icons && yarn build:prod && yarn generate:types",
     "build:prod": "vite build",
     "clean": "rimraf src dist node_modules",
-    "generate:icons": "svgr -- ./assets/icons",
+    "generate:icons": "svgr --template template.cjs -- ./assets/icons",
     "generate:types": "tsc --noEmit false --emitDeclarationOnly --declarationDir dist"
   },
   "gitHead": "c74900b0ee3525510d266dc83c9743cb24dafced"
--- a/packages/strapi-icons/template.cjs
+++ b/packages/strapi-icons/template.cjs
@@ -1,0 +1,34 @@
+function template(
+  { imports, interfaces, componentName, props, jsx, exports },
+  { tpl }
+) {
+  const iconName = componentName + 'Component';
+  const styleName = componentName + 'Wrapper';
+  const svgIconComponent = '<' + iconName + ' {...props} />';
+  const wrapperComponentStart = '<' + styleName + ' colors={colors}>';
+  const wrapperComponentEnd = '</' + styleName + '>';
+  const toRender = wrapperComponentStart + svgIconComponent + wrapperComponentEnd;
+
+  return tpl`
+    import * as React from 'react';
+    import styled from 'styled-components';
+
+    const ${styleName} = styled.div\`
+      \${({ theme, color }) => \`
+        color: \${theme.colors[color]};
+      \`};
+
+      \${({ theme, colors }) => \`
+        \${Object.entries(colors).map(([name, value], index) => \`
+          --var-color-\${color.name}: \${theme.colors[value]};
+        \`)}
+      \`};
+    \`;
+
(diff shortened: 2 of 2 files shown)


joshuaellis:
Cool start! I think single colour icons are pretty straight forward, we set `currentColor` and strongly advise people use that because you're right `:hover` becomes complicated! I see why you want to add the theme colors to the SVG components and while the single colour icons don't require it because they'd use `currentColor` it would be harder to distinguish between the two in formatting, although a wrapper seems a bit over-engineered, imo i would just wrap them all in a `styled` function and do it that way, less dom nesting.

My main question comes around more than one colour icon – why do we need to control the colours in these scenarios? Do we have an audit? It feels like we're creating a solution to a problem we shouldn't have, if they're only available in one style then maybe it's about highlighting this?

gu-stav:
> imo i would just wrap them all in a styled function and do it that way, less dom nesting.

Yeah I can explore if that is possible using a template.

> My main question comes around more than one colour icon – why do we need to control the colours in these scenarios? 

I think we do. If you search for `svg` in all `*.js` files of the CMS you will find about 400 styled overwrites which are all more or less like:


```js
const FieldContainer = styled(Flex)`
  border: 1px solid
    ${({ theme, isSibling }) => (isSibling ? theme.colors.neutral150 : theme.colors.primary200)};

  svg {
    width: ${10 / 16}rem;
    height: ${10 / 16}rem;

    path {
      fill: ${({ theme, isSibling }) => (isSibling ? undefined : theme.colors.primary600)};
    }
  }
`;
```

By using an explicit color prop we are consistent with the rest of the design-system. `currentColor` is a sensible default for most instances, but if needed (e.g. for the `isSibling` example above) developers have full control.

The other benefit I see in wrapping svg in a styled component would be that we can get rid of all these dimension overwrites and just use `width` / `height` attributes like for any other component.

joshuaellis:
I think there was some confusion with this line:

> My main question comes around more than one colour icon – why do we need to control the colours in these scenarios?

i meant with an icon that has more than one color like your example:

```jsx
<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
    <style>
     #logoRocket {
       // using a custom property with a fallback value
       fill: var(--color-logo-rocket, red);
     }
    </style>

    <path id="logoRocket" d="...." />
</svg>
```

Why do we need to control all the colours? I'd assume these are "complex" icons and are specific to their design as opposed to needing to blend into the ui? 🤔

gu-stav:
> I think there was some confusion with this line:

Ah I see. I agree - I think we can ignore that for now and see this PR as exploration on whether this would be possible. 👍🏼 

Do you have any other ideas that we could explore? Otherwise I'd share this idea in the DS Slack channel, see if anyone else has feedback and if there aren't any objections make the implementation a task for the DS@v2 epic.

joshuaellis:
No I think it's a very straight forward solution – ensure the inputs are good quality, ensure the output is `currentColor` and we can pass theme color names to change said color 😎

PR #1319, opened 2023-08-31
feat(IconButton): add `size` and `variant` props

gu-stav:
#### Add a `size` prop

So far the `IconButton` had only one size. For Review Workflows we need them to be large too, according to the design, to match the height of the `Select` component.

This PR adds support for the 3 sizes defined for regular buttons: `S | M | L`

Screenshot 2023-08-31 at 14 51 00

[Preview](https://design-system-git-feat-iconbutton-variants-strapijs.vercel.app/?path=/story/design-system-components-iconbutton--sizes)

#### Add a `variant` prop

So far `IconButton` had only one variant, which is very similar to the tertiary variant of Button: https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-button--variants

This PR adds the `secondary` variant, to be able to style the `IconButton` in the same way as `Button` (see screenshot).

In v2 we probably want to unify the variants across `Button` and `IconButton`, because the current defaults are different.

[Preview](https://design-system-git-feat-iconbutton-variants-strapijs.vercel.app/?path=/story/design-system-components-iconbutton--variants)

> **Note**
> This also replaces the inline snapshot for the modal, because with prettier@3 they can no longer be updated.

```
Summary of all failing tests
@strapi/design-system:  FAIL  packages/strapi-design-system/src/ModalLayout/__tests__/ModalLayout.spec.tsx
@strapi/design-system:   ● Test suite failed to run
@strapi/design-system:     Jest: Inline Snapshots are not supported when using Prettier 3.0.0 or above.
@strapi/design-system:     See https://jestjs.io/docs/configuration/#prettierpath-string for alternatives.
@strapi/design-system:       at saveInlineSnapshots (node_modules/jest-snapshot/build/InlineSnapshots.js:101:15)
```

### Why is it needed?

See above.

### How to test it?

- https://design-system-git-feat-iconbutton-variants-strapijs.vercel.app/?path=/story/design-system-components-iconbutton--sizes

- https://design-system-git-feat-iconbutton-variants-strapijs.vercel.app/?path=/story/design-system-components-iconbutton--variants

Refs: https://github.com/strapi/strapi/pull/17611


packages/strapi-design-system/src/IconButton/constants.ts
```diff
@@ -0,0 +1,10 @@
+// TODO: we should align the default state in v2 with the Button
```
joshuaellis, line 1: have you added this to the board?

packages/strapi-design-system/src/IconButton/constants.ts
joshuaellis, line 1: does this need to be in it's own file? is it used by anyone else? 🤔

joshuaellis (commented):
Overall looks nice ✨ a great use case for visual regression testing ngl

packages/strapi-design-system/src/IconButton/constants.ts
gu-stav: yes 👍🏼

joshuaellis (approved):

joshuaellis:
Can you rebase and add a changeset please? 🙏🏼

gu-stav:
Yes I was waiting for that 👍🏼

joshuaellis (approved):

PR #1322, opened 2023-09-01
fix: dts build / pkg exports & husky

joshuaellis:
### What does it do?

* Fixes the build pipeline for types
* Fixes the package.json export mapping
* Fixes huksy not being installed by default

### Why is it needed?

* in #1273 it turned out the way the ternary was written that we'd always build the types if we omitted the flag which would have introduced a breaking change if done without awareness.
* while working on the helper-plugin I realised that v2 types were not accessible, so I added a specific path for those exports, afterwards they still weren't being picked up I realised in the helper-plugin we need to use node16 module resolution – which showed the issue in the exports map, from what i've learnt doing strapi/strapi#17747 and through other libraries, its _alot_ easier not to have `type: "module"`.
* Husky wasn't working well with yarn 3, I re-followed the README docs of said library

### How to test it?

* Link the DS to helper-plugin
* change Notifications to `tsx`
* edit `tsconfig/client` to have `moduleResolution: "Node16"`
* You should have types for both the root import and the v2 import path


packages/primitives/package.json
```diff
@@ -2,10 +2,9 @@
   "name": "@strapi/ui-primitives",
   "version": "1.9.0",
   "license": "MIT",
-  "type": "module",
```
gu-stav, line 5: Why did you have to remove `type: module`? 🤔

joshuaellis: This is a bit interesting, so following work on `react-spring` around the package.json etc. There was surprisingly a twitter thread at the same time i was _lightly_ involved in with some interesting people – https://twitter.com/devongovett/status/1629161014812897282?s=20 anyway, an output of that was that at this point, using `type: module` would cause more headaches than it's worth when you're trying to support CJS & ESM at the same time, something imo we should be doing because the ecosystem is still very fractured.

You can read the redux maintainer's blog post on the subject [here](https://blog.isquaredsoftware.com/2023/08/esm-modernization-lessons/) which was equally as interesting.

the TLDR; is that adding `type: module` was more of a headache for a variety of bundlers which whilst this isn't a package 1million people use in 100000 environments I'd rather support all the things properly so we're not having issues if we make a jump to X or the project evolves etc. 👍🏼

gu-stav (approved):

PR #1332, opened 2023-09-06
chore: move to turbo repo

joshuaellis:
### What does it do?

* goes full turbo(repo)

### Why is it needed?

* IMHO Turbo's more intuitive to use and easier to set up 🤷🏼‍♀️ 

### How to test it?

* Give it a go and see how you feel


.github/workflows/bundle-size.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - run: yarn install
 
       - name: Build packages
-        run: yarn build
+        run: yarn build --filter=\!@strapi/design-system-docs
```
udimberto, line 30: Just a small suggestion to all these `.github/workflows/*.yml` files:

```suggestion
        run: yarn build:ci
```

No strong motivation.
Not a blocking issue at all.

I just think that:
By having this new script `build:ci` in `package.json`, we will have all the arguments (--filter, in this case) into this single script, which can be easier to maintain for further potential changes.

WDYT?

udimberto (approved):
### LGTM

.github/workflows/bundle-size.yml
joshuaellis: I think it's an interesting idea for sure, if we were doing more than one filter i'd probably say yeah we should do it, but on the other hand abstracting it means someone has to look one more place to change it and with no testing ability you can't ensure there's no knock on effect... tricky balancing act imo.

PR #1333, opened 2023-09-06
fix(IconButton): respect `borderWidth`

gu-stav:
### What does it do?

Apply only border-color to `IconButton`s, bceause they also receive `borderWidth` and `noBorder` to set the border width. This bug has been introduced in https://github.com/strapi/design-system/pull/1319 and leads to too many borders being rendered.

### Why is it needed?

Fixes a bug.

### How to test it?

Add `noBorder` or `borderWidth={0}` attributes to `variants`.

### Related issue(s)/PR(s)

- https://github.com/strapi/design-system/pull/1319

### DIFF
--- a/.changeset/forty-seals-tan.md
+++ b/.changeset/forty-seals-tan.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': patch
+---
+
+fix(IconButton): respect `borderWidth`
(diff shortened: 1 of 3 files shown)


joshuaellis (approved):

gu-stav:
> Chromatic's visual regression could have caught this in the offending PR. What do you think about implementing this sooner rather than later and work on improving the setup as part of v2?

Yes and no. Right now there is no dedicated example for `noBorder` which is why I didn't see it. We would probably have to create examples for all props and their important combinations first, which is something we could start immediately. I'd also like to explore if we should rather use knobs and configuration per component to achieve that over having one example per prop?

joshuaellis:
>  I'd also like to explore if we should rather use knobs and configuration per component to achieve that over having one example per prop

yeah i definitely agree.

PR #1334, opened 2023-09-06
fix(Box): improve border CSS rendering

gu-stav:
### What does it do?

Improves rendering of borders. The previous behavior seems very inconsistent to me and is hart to maintain, as several CSS properties interfere with each other.

This is more of a discussion starter, because I am not sure if we want to implement this in v1 since there is a chance this breaks a few components.

### Why is it needed?

I think we can assume we can render a `1px solid` border, if:

- `border-width` AND `border-color` is set (border-style: solid)
- `border-color` is set (border-width: 1px, border-style: solid)
- `border-color` AND `border-style` is set (border-width: 1px)

Simplified: we can assume a border of `1px solid` as soon as at least `border-color` is set. `borderStyle` and `borderWidth` would overwrite the style / width.


packages/strapi-design-system/src/BaseButton/BaseButton.tsx
```diff
@@ -36,8 +36,6 @@ export const BaseButton = React.forwardRef<HTMLButtonElement, BaseButtonProps>(
         disabled={disabled}
         padding={2}
         hasRadius
-        background={background}
-        borderColor="neutral200"
```
gu-stav, line 40: Every component that inherits from `BaseButton` does set these properties explicitly.

PR #1335, opened 2023-09-07
feat: export theme types

joshuaellis:
### What does it do?

* exports out theme as `StrapiTheme` for use with `styled-components`

### Why is it needed?

* So the `helper-plugin` can know what theme values exist

### Notes

* even though i'm adding something, this is not a changeset because types are not published.


markkaylor (approved):
Nice looks good! Thanks

PR #1344, opened 2023-09-12
fix: The content-manager header has a z-index of 10 while the modal has a z-index of 4 

gitstart:
### What does it do?
-  Fixes [issue](https://github.com/strapi/strapi/issues/17871) where the Modal and Dialog appear behind the content-manager because they have a lower z-index value. 

### Why is it needed?
- Improve UI/UX when the Modal and Dialog is correctly positioned when opened.

### How to test it?
1. Run the app `yarn develop`
2. Find  `combined-w-scroll` in `Components` and click on `Add an entry` button.
3. See that the modal is in front of header.

### **Demo**

https://www.loom.com/share/06f94943db7046fb955bbf8530fa6f65

### **Before**
image

### **After**
image

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.

### DIFF
--- a/packages/strapi-design-system/src/Dialog/Dialog.tsx
+++ b/packages/strapi-design-system/src/Dialog/Dialog.tsx
@@ -47,7 +47,7 @@ export const Dialog = ({ onClose, title, as = 'h2', isOpen, id, ...props }: Dial
 
   return (
     <Portal>
-      <DialogWrapper padding={8} position="fixed" zIndex={4}>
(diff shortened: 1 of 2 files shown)


marco-villanueva:
If only the z-index of the modal is increased, no options, comboboxes, popovers and I don't know if there are any other components. 

In my case what I did was to put the header of the content-manger in 5, the modal in 5 and the others that are hardcoded as the dialog in 5.

joshuaellis:
not needed in design-system@2

PR #1351, opened 2023-09-14
fix: export LinkButton type

Feranchz:
### What does it do?

Export the v2 LinkButton to start using it in other packages.


gu-stav (approved):

gu-stav (changes requested):
Sorry :) Code looks good, but please add a changeset 🙏🏼

gu-stav (approved):

joshuaellis (approved):

PR #1360, opened 2023-09-24
fix: fixs dateValue?.toDate not found error

Sheikh-JamirAlam:
### What does it do?

Srapi Admin dashboard page crashes as when we go to the Advanced Settings option of a Date Field when type is datetime. Here's a screen recording of the issue:-

https://github.com/strapi/design-system/assets/48979376/b806307e-163a-4bb9-85cc-d7cd29107a13

The error says `dateValue?.toDate is not a function` . So this occurs due to [`selectedDate={dateValue?.toDate('UTC')}`](https://github.com/strapi/design-system/blob/6233e279f773ccedd38dd179d2d3d21951bc1538/packages/strapi-design-system/src/DateTimePicker/DateTimePicker.tsx#L195). At initial state dateValue is an empty string "" and therefore when we try to perform toDate() function on it, it fails. Therefore, I made the changes to make sure if dateValue is empty string by default it selectedDate will be undefined.

### Why is it needed?

Fixes [#18123](https://github.com/strapi/strapi/issues/18123). Initially `selectedDate` should be undefined, else it will break.

### DIFF
--- a/packages/strapi-design-system/src/DateTimePicker/DateTimePicker.tsx
+++ b/packages/strapi-design-system/src/DateTimePicker/DateTimePicker.tsx
@@ -192,7 +192,7 @@ export const DateTimePicker = React.forwardRef<DatePickerElement, DateTimePicker
             </VisuallyHidden>
             <DatePicker
               {...props}
-              selectedDate={dateValue?.toDate('UTC')}
+              selectedDate={dateValue ? dateValue.toDate('UTC') : undefined}
               onChange={handleDateChange}
               error={typeof error === 'string'}
               required={required}


joshuaellis:
Can you actually fill in the template explaining what you’ve done and why with detailed explanation. Low effort PRs will be closed, the context is important for us to understand historically our changes.

Sheikh-JamirAlam:
> Can you actually fill in the template explaining what you’ve done and why with detailed explanation. Low effort PRs will be closed, the context is important for us to understand historically our changes.

@joshuaellis Editted my previous PR, sorry about the inconvenience.

joshuaellis:
> At initial state dateValue is an empty string "" and therefore when we try to perform toDate() function on it, it fails. Therefore, I made the changes to make sure if dateValue is empty string by default it selectedDate will be undefined.

I find this weird, we shouldn't be passing strings as dates imo 🤔 I think i'll have to investigate this in the CMS before proceeding with this PR...

joshuaellis:
This should be fixed in the CMS not here, you're just masking the bug.

PR #1371, opened 2023-09-28
feat(design-system): add onCloseAutoFocus

markkaylor:
### What does it do?

Adds the onCloseAutoFocus to single select

### Why is it needed?

We need to be able to prevent the default behavior for the Rich Text Editor in the CMS

### How to test it?

You can test it in the CMS with this PR https://github.com/strapi/strapi/pull/18193

### DIFF
--- a/packages/strapi-design-system/src/Select/SingleSelect.tsx
+++ b/packages/strapi-design-system/src/Select/SingleSelect.tsx
@@ -11,6 +11,7 @@ import { useIntersection } from '../hooks/useIntersection';
 import { Typography } from '../Typography';
 
 type SingleSelectPropsWithoutLabel = Omit<SelectParts.SingleSelectProps, 'value'> &
+  Pick<SelectParts.ContentProps, 'onCloseAutoFocus'> &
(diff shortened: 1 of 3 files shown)


packages/strapi-design-system/src/Select/SingleSelect.tsx
```diff
@@ -81,6 +80,7 @@ export interface SingleSelectInputProps extends Omit<SingleSelectPropsWithoutLab
   'aria-label'?: string;
   id?: string;
   label?: string;
+  onCloseAutoFocus?(event: Event): void;
```
markkaylor, line 83: ℹ️ I'm not sure I've added this in the right spot and I don't know if we also want to add this for multi-select or not 🤔

joshuaellis: I would extend the type again doing `Pick<SelectParts.ContentProps, 'onCloseAutofocus'>` instead imo.

And for consistency so they're in alignment I would make the same change on the MultiSelect, the closer their APIs the easier it is to understand how to use them imo

joshuaellis (commented):
You just need to add a changeset for your feature addition, see the CONTRIBUTING guidelines for more info :)

joshuaellis (approved):
TFO2mwVPIFoOJcuTSC

PR #1381, opened 2023-10-03
fix: Combobox rendered text when value prop changes

mazzucchelli:
### What does it do?

Updates the Combobox rendered value if the corresponding prop changes "externally" (without clicking an option)

### Why is it needed?

Because if the prop changes, the "internal" value is actually updated but the rendered text not

### How to test it?

I tested this manually, let me know if I should update UT

```
// docs/stories/Combobox.stories.tsx

export const Depending = {
  render: () => {
    const [value, setValue] = useState('');

    const options = ['apple', 'avocado', 'banana', 'kiwi', 'mango', 'orange', 'strawberry'];

    const handleChange = (index) => {
      setValue(options[index]);
    };

    return (
      <Flex direction="column" alignItems="stretch" gap={11}>
        {options.map((option, index) => (
          <Button key={index} onClick={() => handleChange(index)}>
            {index}
          </Button>
        ))}
        <span style={{ color: 'white' }}>{value}</span>
        <Combobox
          placeholder="My favourite fruit is..."
          label="Fruits"
          value={value}
          onChange={setValue}
          onClear={() => setValue('')}
        >
          {options.map((option) => (
            <ComboboxOption key={option} value={option}>
              {option}
            </ComboboxOption>
          ))}
        </Combobox>
      </Flex>
    );
  },

  name: 'depending',
} satisfies Story;

```

### Related issue(s)/PR(s)

### DIFF
--- a/.changeset/long-buses-yawn.md
+++ b/.changeset/long-buses-yawn.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': patch
+---
+
+Fixed a visual bug for Combobox component when value is updated externally
--- a/packages/strapi-design-system/src/Combobox/Combobox.tsx
+++ b/packages/strapi-design-system/src/Combobox/Combobox.tsx
@@ -191,6 +191,15 @@ export const ComboboxInput = React.forwardRef<ComboboxInputElement, ComboboxInpu
       skipWhen: !internalIsOpen,
     });
 
+    React.useEffect(() => {
+      /**
+       * When the value prop changes "externally" update the text input value too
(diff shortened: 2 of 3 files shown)


joshuaellis (changes requested):
Do you think we can add a test case? I worry that this could easily regress 😬

joshuaellis:
Do you know how to add a changeset? 🤩

joshuaellis (approved):
Awesome Seth Meyers GIF by Saturday Night Live

PR #1392, opened 2023-10-10
make labelledBy prop optional for Main component and remove redundant type

hekystyle:
### What does it do?

1. makes `labelledBy` optional because there is default value assignment inside component. I used `| undefined` by design so it's compatible with https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes.
2. removed `string` type because it's already part of `React.ReactNode` type

### Why is it needed?

Better developer experience.


joshuaellis (approved):
thanks for the contribution! ⭐

PR #1394, opened 2023-10-11
Omit selectedTabIndex and selectTabIndex in TabGroup

christiancp100:
### What does it do?
It omits selectedTabIndex and selectTabIndex in TabGroup because they are not required as props since they are being passed to the context inside the TabGroup component. 

### Why is it needed?

To avoid TS errors.

### How to test it?

Check that the TabGroup component doesn't fail when you don't pass the selectTabIndex and selectedTabIndex properties to it. 

### Related issue(s)/PR(s)


joshuaellis:
@christiancp100 this needs a changeset before we can proceed. But the code change looks fine 👍🏼

joshuaellis (approved):

PR #1413, opened 2023-10-16
fix: make title optional

christiancp100:
### What does it do?

Make the title property of the Alert component optional

### Why is it needed?

Because on the Strapi codebase is already being used without specifying the title. And during the migration from JS to TS the errors are starting to appear

### How to test it?

Use the Alert component and check that no title is required


joshuaellis (approved):

joshuaellis:
Can you add a changeset please

PR #1417, opened 2023-10-18
Add spark icon for blocks stable release badge

madhurisandbhor:
### Why is it needed?
 
As a part of Blocks stable release, we are introducing "New" release badge with sparks icon.


remidej (approved):
LGTM but I would rename the PR to something like "add spark icon"

joshuaellis (changes requested):
you've not added a changeset.

Issue #1420, opened 2023-10-19
[bug]: When there is no scroll bar on the page, the floating position of Tooltip will shift


BlueBaidu:
### What version of strapi-design-system are you using?

DS:v1.12.1
Operating system: Windows
Browser: Chromium Version 118.0.5993.71 (official version) (64 bit), Microsoft Edge Version 114.0.1823.37 (official version) (64 bit)

### What's Wrong?

When there is no scroll bar on the page, the first time the hover is on the icon, the position of the Tooltip shifts and does not appear directly above the icon. After hovering again, the Tooltip position returns to normal.
![image](https://github.com/strapi/design-system/assets/101248999/d4394574-a6d4-4cd6-9ff3-763398d738c3)
![image](https://github.com/strapi/design-system/assets/101248999/f13ad0bf-cab0-4cf5-a661-3643a005a9f9)

### To Reproduce

This behaviour is also visible on the demo documentation page:
https://api-0khig.strapidemo.com/admin/plugins/content-type-builder/content-types/api::place.place

### Expected Behaviour

The suspension position of Tooltip exhibits consistency between the first hover and subsequent hovers

joshuaellis:
solved by #1722

PR #1421, opened 2023-10-19
fix: 修复在页面没有滚动条时，Tooltip位置出现偏移的问题

BlueBaidu:
### What does it do?

Fixed an issue where the position of the tooltip shifted when the hover was first placed on the icon without a scroll bar on the page.

Adjusted the CSS style to ensure that the tooltip performs consistently on the first hover and subsequent hovers.

### Why is it needed?

The previous issue resulted in poor user experience, as the offset of the tooltip position when the hover is first placed on the icon can cause confusion.

This modification fixes this issue and improves the usability of the user interface.

### How to test it?

Open the application on a page without a scroll bar, Please refer to the demo document address:

https://api-0khig.strapidemo.com/admin/plugins/content-type-builder/content-types/api::place.place

Try hovering on the icon for the first time and observe if the position of the tooltip is correct.

Hover again on the icon and check if there is no positional offset.

### Related issue(s)/PR(s)

Fixed: https://github.com/strapi/design-system/issues/1420


joshuaellis:
not needed in design-system@2

Issue #1418, opened 2023-10-18
[bug]: `v2/SimpleMenu` with Icon does not Show Icon


derschiw:
### What version of strapi-design-system are you using?

NPM Version: 8.19.2
Node.js: v16.18.1
Strapi Design System: 1.12.0
Browser: Firefox 118.0.1, Chromium Version 118.0.5993.54 (Official Build) Arch Linux (64-bit)

### What's Wrong?

When using the `SimpleMenu` and adding an `IconButton` then there is no icon visible. Instead the button is just empty in a specified color. 

```js
<SimpleMenu aria-label="Notifications" as={IconButton} icon={<Bell />}>
```

This behaviour is also visible on the documentation page:
https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-v2-simplemenu--with-icons

### To Reproduce

You can see it in the documentation:
https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-v2-simplemenu--with-icons

### Expected Behaviour

The icon should be visible in the button.

PR #1435, opened 2023-10-26
fix: SimpleMenu - trigger style props as "IconButton"

udimberto:
### What does it do?

Update the style props applied to the SimpleMenu/`MenuTrigger` component, in case of this trigger is used as a `IconButton`.

### Why is it needed?

To fix the problem that the SimpleMenu/`MenuTrigger`, when used as `IconButton`, is not showing the icon inside the button (due to the button-padding applied, not lefting any space to render the icon-svg into it).

### How to test it?

Running the Storybook locally, or checking the [online preview](https://design-system-git-fix-simple-menu-icon-button-strapijs.vercel.app/?path=/story/design-system-components-v2-simplemenu--with-icons).

Check that the `v2/SimpleMenu` is displaying the Icon inside the trigger-button, when using the combined props:
```tsx
<SimpleMenu as={IconButton} icon={<More />}>
  ...
</SimpleMenu>
```

### Related issue(s)/PR(s)

#1418

### DIFF
--- a/packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
+++ b/packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
@@ -29,6 +29,11 @@ interface TriggerProps extends ButtonProps {}
 
 const MenuTrigger = forwardRef<HTMLButtonElement, TriggerProps>(
   ({ size, endIcon = <CarretDown width={`${6 / 16}rem`} height={`${4 / 16}rem`} aria-hidden />, ...props }, ref) => {
+    const isSizeS = size === 'S';
+    const isIconButton = (props.as as any)?.displayName === 'IconButton';
+    const paddingY = isSizeS ? 1 : 2;
+    const paddingX = isSizeS ? 3 : 4;
+
     return (
       <DropdownMenu.Trigger asChild>
         <Button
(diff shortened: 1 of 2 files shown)


packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
```diff
@@ -29,17 +29,23 @@ interface TriggerProps extends ButtonProps {}
 
 const MenuTrigger = forwardRef<HTMLButtonElement, TriggerProps>(
   ({ size, endIcon = <CarretDown width={`${6 / 16}rem`} height={`${4 / 16}rem`} aria-hidden />, ...props }, ref) => {
+    const isSizeS = size === 'S';
+    const isIconButton = (props.as as any)?.displayName === 'IconButton';
```
joshuaellis, line 33: I'm not sure about this, I don't think you're really supposed to access those props.

Does it make sense for this component to adjust its styling based on the as prop which realistically could be any component.

Is this to fix an issue in cloud? How come you can't change the padding there? 🤔

joshuaellis (changes requested):
Discussion

packages/strapi-design-system/src/v2/SimpleMenu/Menu.tsx
udimberto: Yeah, I agree about the conditional with `displayName` is not very good.

But, look, I am trying to fix this problem here:
https://design-system-git-main-strapijs.vercel.app/?path=/story/design-system-components-v2-simplemenu--with-icons
A variation of `SimpleMenu`, which uses the `IconButton` as the menu trigger (instead of a regular text).
IDK when, but a recent update on the DS Button result in this "missing-icon" behavior here, where the padding don't let the SVG appear.

So, yes, I am trying to fix a visual issue for Strapi Cloud Dashboard too
😅

HichamELBSI:
Is this PR still relevant ? @udimberto

udimberto:
Not anymore, @HichamELBSI, thank you

PR #1516, opened 2023-12-15
fix: TabGroup react to changes of the initialSelectedTabIndex value

simotae14:
### What does it do?

It solves an issue we had in the TabGroup inside the CMS used in the Markeplace and Content Releases, it doesn't react to the changes of the initialSelectedTabIndex prop

### Why is it needed?

Because for example when you try to use the browser back button or you click on the side bar menu on the left the tabs are not updated with the current selection

### How to test it?

Link this DS version to your CMS and try to, for example, to the Marketplace page admin/marketplace
- then click on the tab Providers
- then use the Back button of the browser you need to see the Plugins tab selected instead of the Providers one

### DIFF
--- a/.changeset/tiny-birds-tell.md
+++ b/.changeset/tiny-birds-tell.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': minor
+---
+
+Fix: Added useEffect on TabGroup component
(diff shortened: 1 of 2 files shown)


joshuaellis (changes requested):
Is the correct solution? does the initialSelectedTabIndex change as we use the application? Because if we're using the URL as the state for tabs maybe we'd be better off using the `useControllableState` hook?

Also I your changeset message could be clearer, you've said what you've done but not why you've done it e.g. "Added useEffect in TabGroup so state can be controlled outside of component" :)

simotae14:
we have found a solution in the CMS https://github.com/strapi/strapi/pull/19229

PR #1548, opened 2024-01-08
fix(design-system): checkbox svg need quotes to work correctly

joshuaellis:
### What does it do?

* adds quote marks to where we inline the string from the checkbox asset import

### Why is it needed?

* the CSS comes out incorrectly resulting in no checks in the CMS

### Related issue(s)/PR(s)

* mentioned in https://github.com/strapi/strapi/pull/19136#issuecomment-1881242950

### DIFF
--- a/.changeset/brave-chefs-behave.md
+++ b/.changeset/brave-chefs-behave.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': patch
+---
+
+fix: checkbox svg needs quote marks around to work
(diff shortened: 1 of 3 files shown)


remidej:
@joshuaellis do you know why I don't see it working in the [Vercel preview link](https://design-system-git-fix-checkbox-icon-strapijs.vercel.app/?path=/docs/design-system-components-checkbox--docs)?

joshuaellis:
because the assets URLs are handled differently between development & production, which is not ideal. thanks for letting me know

joshuaellis:
@remidej do you mind checking now? I think it's working, but would like your eyes!

remidej (approved):
✅✅✅

PR #1605, opened 2024-02-22
Feat: new filter "contains" is added on combobox for autocomplete type:list

madhurisandbhor:
### What does it do?

Till now we were filtering by default matching starting character and autocomplete accepted values were: "none", "list", "both". Now we have added a new option `contains` to filter by substring. 

However `filter:'contains'` option is only available on `autocomplete='list'` as having for `autocomplete="both"` would not work as expected. Also to have backward compatibility and for consistency across all supported filter types, we are introducing object counterparts of existing types.

### Why is it needed?
Filtering by substring is required in timezone dropdown in scheduling releases as TZ starts with UTC and not actual TZ value.

### How to test it?

You can test newly added filter option in storybook under `primitives/combobox` as well as `DS/combobox` component

### DIFF
--- a/docs/stories/Combobox.stories.tsx
+++ b/docs/stories/Combobox.stories.tsx
@@ -184,12 +184,12 @@ export const Creatable = {
   name: 'creatable',
 } satisfies Story;
 
-type TAutocomplete = 'none' | 'list' | 'both' | { type: 'list'; filter: 'startsWith' | 'contains' };
+type Autocomplete = 'none' | 'list' | 'both' | { type: 'list'; filter: 'startsWith' | 'contains' };
 
 export const Autocomplete = {
   render: () => {
     const [value, setValue] = useState('');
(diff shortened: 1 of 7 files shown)


docs/stories/Combobox.stories.tsx
```diff
@@ -184,10 +184,18 @@ export const Creatable = {
   name: 'creatable',
 } satisfies Story;
 
+type TAutocomplete = 'none' | 'list' | 'both' | { type: 'list'; filter: 'startsWith' | 'contains' };
```
madhurisandbhor, line 187: could not import ComboboxProps here to fetch type of autocomplete 😕

joshuaellis (commented):
I was meant to send this yesterday but I didn't post it!

joshuaellis (changes requested):
Really great work, super clear and concise, good documentation updates too. Lets add some tests for the filtering updates and it's probably good to go! ⭐

joshuaellis (approved):

PR #1674, opened 2024-04-09
Fix: select padding

madhurisandbhor:
### What does it do?

Fixes top bottom paddings of the select 

### Why is it needed?

The top and bottom paddings of the Select disappear when there's multiple lines of tags


joshuaellis (approved):
⭐

PR #1675, opened 2024-04-09
Fix:  accordian complete header clickable issue

madhurisandbhor:
### What does it do?

Accordian complete header is made clickable.

### Why is it needed?

The padding of the "header" of the Accordion is not included in the click area


joshuaellis (approved):
Season 5 Nbc GIF by The Office

PR #1684, opened 2024-04-12
feat: add zIndex prop to Dialog

markkaylor:
### What does it do?

- add a zIndex prop with default 4 

### Why is it needed?

- to give more controll

### How to test it?

- pass the prop and make sure the value comes through


remidej (approved):

joshuaellis (approved):
🥳

PR #1695, opened 2024-04-18
Update Styled Components to V6

christiancp100:
### What does it do?

* Updates styled-components peer dep & local dep from v5 to v6
* Changes Box/Grid/Flex & Typography to react components over styled-components
* Disallow use of `as` and instead use the `tag` prop to change underlying DOM element
* Default color of Typography is `currentColor` similar to SVGs

### Why is it needed?

* Keep the design-system dependencies updated
* Integrate inbuilt type system from `styled-components`

### Related issue(s)/PR(s)

* resolves DX-1188


docs/custom.d.ts
joshuaellis, line 1: for your PR, i would put this back for the time being

joshuaellis:
I've had a look at this, on the surface it looks fine because _most_ components use `Box` so the `isValidProp` will handle it for us, but in the instance of `CardAsset` which is just a `styled.div` if we pass something like `isUp` then it shows a warning:

CleanShot 2024-04-19 at 08 34 42@2x

I think lets go with the original approach to make Box a React component & alias the props to their `$` equivalent. This way, we have "one way" of doing everything and we're less prone to introduce these warnings.

joshuaellis:
Okay I think this is good to go, to solve a lot of the issues we had around how `as` and `forwardAs` behaves, i've just made the `tag` prop, also benefit is it's not tied to the styling library should we choose to change in the future.

joshuaellis (approved):

PR #1703, opened 2024-04-25
chore!: remove cms specific components

madhurisandbhor:
### What does it do?

* Removed CMS specific components. with  new design system V2, we are moving any specific CMS specific component to its own package as DS expands to cloud as well.

### Related Issues

* resolves DX-1381


joshuaellis (changes requested):
Can we write a more comprehensive changeset about which components specifically were removed? Also lets not forget to add these to the BREAKING_CHANGES doc :)

Also before this is merged we need to make the opposite PR in the CMS that copies over the components 💪🏼

joshuaellis:
See https://design-system.strapi.io/iframe.html?viewMode=docs&id=getting-started-migration-guides-v1-to-v2--docs&globals=#removed-cms-specific-components you should be able to import them from `@strapi/strapi`, if they're not available I would open an issue in that repo instead.

brianjd:
Noted, thanks @joshuaellis

arilanto:
What is CMS ???

PR #1709, opened 2024-04-29
fix(design-system): codemirror packages that aren't in package.json are declared as external

### LINKED ISSUE #1706: [bug]: 2.0.0-beta.2 codemirror error Unrecognized extension value in extension set

2.0.0-beta.2

### What's Wrong?

Use of the JSON input throws the following error:

Screenshot 2024-04-26 at 12 03 36

### To Reproduce

On [Strapi v5/main](https://github.com/strapi/strapi/tree/v5/main) go to any content-type in the content manager that is using the JSON input. You will see the error. 

### Expected Behaviour

No error when using the JSON input

joshuaellis:
### What does it do?

* finds the codemirror deps we access but aren't declared in the package.json and make them external to avoid being bundled

### Why is it needed?

* JSON fields crash the strapi app because we load more than one version of codemirror state

### Related issue(s)/PR(s)

* resolves #1706

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -161,6 +161,19 @@ if we're already in pre-release mode, this is fine and you should continue.
 
 We recommend you use [`yalc`](https://github.com/wclr/yalc). If you're working in a monorepo setup (e.g. the Strapi monorepo), then ensure you use the `--pure` option and follow the [advanced guide](https://github.com/wclr/yalc#use-with-yarnpnpm-workspaces). This works a lot better than `yarn link`.
 
+### Example steps
+
+These steps assume you're in the root of the design-system project and on MacOS and adding to the Strapi Monorepo (a normal use case).
+
+```shell
+cd packages/strapi-design-system # this could be icons too if you preferred.
+yalc publish
+cd ../../your-strapi-repo
+yalc add @strapi/design-system --pure
+```
+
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -159,35 +159,7 @@ if we're already in pre-release mode, this is fine and you should continue.
...
-You should also remove the webpack alias for `@strapi/design-system` in the Strapi monorepo at `packages/core/admin/webpack.alias.js`
-
-Your application should now be using your local copy of the design system.
-
-To revert back to the released version of the design system use [`yarn unlink`](https://yarnpkg.com/cli/unlink#usage):
-
-```
-yarn unlink ../<relative-path-to-strapi-design-system>
-```
+We recommend you use [`yalc`](https://github.com/wclr/yalc). If you're working in a monorepo setup (e.g. the Strapi monorepo), then ensure you use the `--pure` option and follow the [advanced guide](https://github.com/wclr/yalc#use-with-yarnpnpm-workspaces). This works a lot better than `yarn link`.
```
innerdvations, line 162: Do you have an example line to run so we don't have to figure it out from the yalc docs?

joshuaellis: done ✅

innerdvations (approved):
LGTM but I didn't test it

PR #1710, opened 2024-04-30
chore!: make CardAction a react component

joshuaellis:
### What does it do?

* makes `CardAction` a react component

### Why is it needed?

* follows the same principles we've had elsewhere of not exporting styled components to abstract the library specific APIs which in theory, would make upgrading easier in the future

### Related issue(s)/PR(s)

* resolves DX-1375


packages/design-system/src/components/Card/CardAction.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Flex, FlexComponent, FlexProps } from '../Flex';
+
+type CardActionPosition = 'end' | 'start';
+
+type CardActionProps = Omit<FlexProps<'div'>, 'direction' | 'gap' | 'position'> & {
+  position: CardActionPosition;
+};
+
+const CardActionImpl = ({ position, ...restProps }: CardActionProps) => {
+  return <CardAction $position={position} {...restProps} direction="row" gap={2} />;
```
innerdvations, line 13: Do we intend to prevent direction and gap from being passed in as props?

joshuaellis: It's omitted from the types, and it'll be overwriten, im not sure what else you'd do?

jhoward1994 (approved):

PR #1713, opened 2024-05-03
feat!: refactor field component API

joshuaellis:
### What does it do?

* refactors the Field API to be a composite component
* uses a default value so you can use our inputs in and out of the Field component

### Related issue(s)/PR(s)

* resolves DX-1386


packages/design-system/.eslintrc
```diff
@@ -8,5 +8,6 @@
     // TODO: re-enable this rules
     "react/display-name": "off",
     "@typescript-eslint/no-explicit-any": "warn",
+    "import/export": "off",
```
simotae14, line 11: Why did you add this?

simotae14 (approved):
Great work as always

I saw there are part of code repeated in some input fields
```
const hasError = Boolean(error) || hasErrorProp;
    const id = field.id ?? idProp;
    const required = field.required || requiredProp;

    let ariaDescription: string | undefined;
    if (error) {
      ariaDescription = `${id}-error`;
    } else if (field.hint) {
      ariaDescription = `${id}-hint`;
    }
```
but to be honest I can find a better way to do it.

packages/design-system/.eslintrc
joshuaellis: typescript manages this for us, and there were weird errors where eslint thought there was an issue with dual exports but it wasn't an issue since we use TS namespacing 🤔

joshuaellis:
> Great work as always
> 
> I saw there are part of code repeated in some input fields
> 
> ```
> const hasError = Boolean(error) || hasErrorProp;
>     const id = field.id ?? idProp;
>     const required = field.required || requiredProp;
> 
>     let ariaDescription: string | undefined;
>     if (error) {
>       ariaDescription = `${id}-error`;
>     } else if (field.hint) {
>       ariaDescription = `${id}-hint`;
>     }
> ```
> 
> but to be honest I can find a better way to do it.

Yeah I agree, I'll keep it mind when working on other bits, it might be that the API could be made better again 🤔

PR #1718, opened 2024-05-07
chore!: remove size from inputs

joshuaellis:
chore!: remove size from inputs

### What does it do?

* removes `size` from inputs and uses padding instead

### Why is it needed?

* we don't use size, it's a maintenance burden and really, we should be using padding's anyway so it works with RTL

### Related issue(s)/PR(s)

* resolves DX-1296


packages/design-system/src/components/Badge/Badge.tsx
```diff
@@ -1,26 +1,28 @@
...
 import { Flex, FlexComponent, FlexProps } from '../Flex';
 import { Typography } from '../Typography';
 
-const Base = styled<FlexComponent>(Flex)<{ $size: BadgeSize }>`
-  border-radius: ${({ theme, $size }) => ($size === 'S' ? '2px' : theme.borderRadius)};
-  height: ${({ $size, theme }) => theme.sizes.badge[$size]};
-`;
-
-export type BadgeSize = 'S' | 'M';
+type BadgeSizes = 'S' | 'M';
```
innerdvations, line 6: For types, singular makes more sense to me. If `size: BadgeSize` I know it's a badge size. If I see `size: BadgeSizes` I expect I can pass in an array of sizes or something.

packages/design-system/src/components/Accordion/Accordion.tsx
```diff
@@ -10,8 +10,8 @@ import { Typography } from '../Typography';
 import { AccordionContext } from './AccordionContext';
 import { AccordionDescription, AccordionTitle } from './AccordionToggle';
 
-type AccordionSize = 'S' | 'M';
-type AccordionVariant = 'primary' | 'secondary';
+type AccordionSizes = 'S' | 'M';
```
innerdvations, line 13: same comment as BadgeSizes

packages/design-system/src/components/Combobox/Combobox.tsx
```diff
@@ -296,38 +297,29 @@ export const Combobox = React.forwardRef<ComboboxInputElement, ComboboxProps>(
...
   background: ${({ theme }) => theme.colors.neutral0};
   overflow: hidden;
   display: flex;
   align-items: center;
   justify-content: space-between;
   gap: ${({ theme }) => theme.spaces[4]};
-  min-height: ${({ theme, $size }) => getThemeSize('input')({ theme, size: $size })};
 
-  &[data-disabled] {
+  s &[data-disabled] {
```
innerdvations, line 322: What is `s`?

packages/design-system/src/components/Badge/Badge.tsx
joshuaellis: You think? I was super on the fence about this, and i don't really mind – I wanted to find a library that did something i could draw upon, if you feel strongly about it im happy to do it 😄

innerdvations: Yeah, I think so, although I feel bad that we've had that "deprecated: use BadgeSizes" and now we're undeprecating it (once you remove the comment) instead

joshuaellis: I added the deprecation here, so dw lol

innerdvations (approved):

PR #1719, opened 2024-05-07
chore!: remove deprecated options

joshuaellis:
### What does it do?

* removes all our deprecated options

### Why is it needed?

* they were deprecated

### Related Issues

* resolves DX-1403


packages/design-system/src/components/Combobox/Combobox.tsx
```diff
@@ -20,7 +20,7 @@ import { Typography } from '../Typography';
  * ComboboxInput
  * -----------------------------------------------------------------------------------------------*/
 
-export interface ComboboxProps
+interface ComboboxProps
```
innerdvations, line 23: `defaultTextValue` in the `Omit` down below shouldn't be optional

packages/design-system/src/components/Combobox/Combobox.tsx
innerdvations, line 1: Is there a particular reason you're moving all the exports to the bottom instead of inline? At least in the CMS we've been moving towards the opposite.

packages/design-system/src/components/DateTimePicker/DateTimePicker.tsx
innerdvations, line 1: `once` no longer needs to be imported

innerdvations (commented):
I didn't QA btw, just a code review

packages/design-system/src/components/Combobox/Combobox.tsx
joshuaellis: You have in the BE, in the FE we've been going down lol, I don't think it really matters either way if *someone* adds an eslint rule i'll have abide by it, I just prefer seeing whats exported in one go rather than running up and down a file trying to understand. This is particularly vital in the DS where we then do an `export * `, don't wanna export a fn for no reason!

packages/design-system/src/components/Combobox/Combobox.tsx
joshuaellis: I've just realised how fucked up this type actually is.... 🤦🏼‍♀️

innerdvations (approved):
LGTM, but I didn't QA it

PR #1722, opened 2024-05-09
feat!: refactor Tooltip to use radix-ui

joshuaellis:
### What does it do?

* refactors `Tooltip` to use `radix-ui`
* refactors IconButton API to only use children & label for a cohesive API across the rest of the components & adds `withTooltip` prop to control in-built tooltip availability

### Why is it needed?

* less code to maintain ✅ 

### Related issue(s)/PR(s)

* resolves DX-1212


packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,69 +1,93 @@
...
-          position="absolute"
-          {...props}
-        >
-          {visible && <VisuallyHidden id={descriptionId}>{description}</VisuallyHidden>}
-          <Typography tag="p" variant="pi" fontWeight="bold" textColor="neutral0">
-            {label || description}
-          </Typography>
-        </TooltipWrapper>
-      </Portal>
+const scaleIn = keyframes`
```
innerdvations, line 69: Wouldn't this be fadeIn?

innerdvations (approved):
Didn't QA it but code LGTM

packages/design-system/src/components/Tooltip/Tooltip.tsx
joshuaellis: yes, it would

PR #1732, opened 2024-05-31
chore: refactor ProgressBar to use radix primitive

joshuaellis:
### What does it do?

* refactors ProgressBar to use radix primitive & animates it

### Related issue(s)/PR(s)

* resolves DX-1208


jhoward1994:
The source is show as this at http://localhost:6006/?path=/docs/components-progressbar--docs

image

It seems we are using the <Canvas> component in the same way as other components so not sure whats going on. any ideas?

madhurisandbhor (approved):
We can also just show base version in docs, its animated so its bit confusing to user if animation is a part of ProgressBar component 🤔

joshuaellis:
> We can also just show base version in docs, its animated so its bit confusing to user if animation is a part of ProgressBar component 🤔

nice idea 👍🏼

PR #1734, opened 2024-06-05
chore: refactor Switch to use radix primitive

joshuaellis:
### What does it do?

* convert switch to radix primitive

### How to test it?

* visual regression testing will be added via this pr 🪄 

### Related issue(s)/PR(s)

* resolves DX-1210


packages/design-system/src/components/Switch/Switch.tsx
```diff
@@ -1,94 +1,111 @@
...
-  ({ label, onChange, onLabel = 'On', offLabel = 'Off', selected, visibleLabels = false, ...props }, ref) => {
+const SwitchImpl = React.forwardRef<HTMLButtonElement, SwitchProps>(
+  (
+    {
+      visibleLabels,
+      onLabel = 'On',
+      offLabel = 'Off',
+      onCheckedChange: onCheckedChangeProp,
+      checked: checkedProp,
+      defaultChecked,
```
madhurisandbhor, line 25: defaultChecked is a required prop?

packages/design-system/src/components/Switch/Switch.tsx
```diff
@@ -1,94 +1,111 @@
...
-const Switch = React.forwardRef<HTMLButtonElement, SwitchProps>(
-  ({ label, onChange, onLabel = 'On', offLabel = 'Off', selected, visibleLabels = false, ...props }, ref) => {
+const SwitchImpl = React.forwardRef<HTMLButtonElement, SwitchProps>(
+  (
+    {
+      visibleLabels,
+      onLabel = 'On',
+      offLabel = 'Off',
+      onCheckedChange: onCheckedChangeProp,
+      checked: checkedProp,
```
madhurisandbhor, line 24: I think instead of renaming props, we could just alias as you already did, to avoid making changes everywhere its used.

```suggestion
      onChange: onCheckedChangeProp,
      selected: checkedProp,
```

packages/design-system/src/components/Switch/Switch.tsx
```diff
@@ -1,94 +1,111 @@
...
-  selected?: boolean;
   visibleLabels?: boolean;
 }
 
-const Switch = React.forwardRef<HTMLButtonElement, SwitchProps>(
-  ({ label, onChange, onLabel = 'On', offLabel = 'Off', selected, visibleLabels = false, ...props }, ref) => {
+const SwitchImpl = React.forwardRef<HTMLButtonElement, SwitchProps>(
+  (
+    {
+      visibleLabels,
```
madhurisandbhor, line 20: optional?

packages/design-system/src/components/Switch/Switch.tsx
joshuaellis: no, but `undefined` is fine as a default value :)

packages/design-system/src/components/Switch/Switch.tsx
joshuaellis: it is optional :) again, it just doesn't require a default value

jhoward1994 (approved):

packages/design-system/src/components/Switch/Switch.tsx
madhurisandbhor: ohk, but in this context `onChange` and `onCheckedChange` wont be same? what could be the other event

madhurisandbhor (approved):

packages/design-system/src/components/Switch/Switch.tsx
joshuaellis: onCheckedChange gets `checked` state so it's recieves a boolean value, onChange is the event, but if it's controlled it won't have changed iirc. Most of our usecases, we should prefer `onCheckedChange` it's a lot easier to work with. The checkbox component also has the state `indeterminate` which therefore avoids the need for another prop 🤓

madhurisandbhor: thanks for the explanation!! 👍

PR #1735, opened 2024-06-05
feat: reimplement size prop

joshuaellis:
### What does it do?

* re-adds the size prop to inputs

### Why is it needed?

* bit of confusion on the direction

### Related issue(s)/PR(s)

* resolves DX-1296


jhoward1994 (approved):
👍🏻

PR #1737, opened 2024-06-10
Feat: responsive styles

madhurisandbhor:
### What does it do?

This PR introduces responsive props to components like `Box`, `Flex` to start with. To pass responsive props on components, user has to use object notation as below,

`<Component padding={{initial: 2, small: 4, medium: 6, large: 8}} />`

User can also pass arrays to shorthand properties like `padding`/`margin`.

`<Component padding={{initial: [2, 4], small: 4, medium: [2, 4, 4], large: "24px" }} />`

### Why is it needed?

To ensure design system is responsive, [read more](https://strapi-inc.atlassian.net/browse/DX-1382).


docs/stories/Box.mdx
```diff
@@ -24,11 +24,14 @@ The `Box` component allows creation of containers according to the guidelines of
...
 <Canvas of={BoxStories.Base} />
 
-## Box with responsive spaces
+## Box with responsive styles
 
-The `Box` component with responsive padding added.
+Responsive styles allow you to specify different values for properties based on different screen sizes or breakpoints.
+The values are defined as an object or array and are applied using a `min-width` condition. If a value is not specified
+for a particular breakpoint, it will inherit the value from the previous breakpoint. The breakpoints are defined in
+theme as : `initial: 0px, small: 520px, medium: 768px, large: 1280px`.
```
madhurisandbhor, line 32: I think we can add a page under foundation for responsive styles and explain there.

packages/design-system/src/themes/common-theme.ts
```diff
@@ -46,6 +51,11 @@ export const commonTheme: CommonTheme = {
...
   spaces: ['0px', '4px', '8px', '12px', '16px', '20px', '24px', '32px', '40px', '48px', '56px', '64px'],
   borderRadius: '4px',
+  breakpoints: {
+    small: '@media(min-width: 520px)',
+    medium: '@media(min-width: 768px)',
+    large: '@media(min-width: 1280px)',
+  },
   mediaQueries: {
     tablet: `@media (max-width: 1100px)`,
     mobile: `@media (max-width: 550px)`,
```
madhurisandbhor, line 61: If this PR looks good for padding/margin, I will remove these existing breakpoints and update at required places

packages/design-system/src/helpers/__tests__/handleResponsiveValues.test.ts
```diff
@@ -0,0 +1,183 @@
...
+      marginTop: {
+        initial: 1,
+        small: '2rem',
+      },
+    };
+    expect(handleResponsiveValues(values, lightTheme)).toMatchInlineSnapshot(`
+      "padding-inline-end: 4px;
+      padding-block-start: 4px;
+      padding-block-end: 4px;
+      padding-inline-start: 4px;
```
joshuaellis, line 108: i would expect `padding-inline-end` to read `8px` here because we've used `paddingRight`? 🤔

packages/design-system/src/helpers/__tests__/handleResponsiveValues.test.ts
```diff
@@ -0,0 +1,183 @@
...
+import { handleResponsiveValues } from '../handleResponsiveValues';
+
+describe('handleResponsiveValues', () => {
+  it('should handle simple string values', () => {
+    const values = { padding: '1rem', margin: '4px' };
+    expect(handleResponsiveValues(values, lightTheme)).toMatchInlineSnapshot(`
+      "padding-block-start: 1rem;
+      padding-inline-end: 1rem;
+      padding-block-end: 1rem;
+      padding-inline-start: 1rem;
```
joshuaellis, line 11: how come we have to use `padding-X-[start|end]` can't we use `padding-X` instead? 🤔 whats the rationale behind this?

packages/design-system/src/helpers/handleResponsiveValues.ts
```diff
@@ -40,52 +26,146 @@ type ResponsiveCSSProperties = Pick<
...
+  marginRight: 'margin-inline-end',
+  marginTop: 'margin-block-start',
+  marginBottom: 'margin-block-end',
+};
 
-export type ResponsiveValue<TCSSProp extends keyof ResponsiveCSSProperties = any> =
-  | ResponsiveValueObject
-  | DefaultThemeOrCSSProp<'spaces', TCSSProp>
-  | ResponsiveValueTuple;
+const fillCssValues = (value: Array<string | keyof DefaultTheme['spaces']>) => {
```
joshuaellis, line 68: This needs some JSDoc explaining what it does and why it does it etc. It's not clear just reading the code.

packages/design-system/src/helpers/__tests__/handleResponsiveValues.test.ts
joshuaellis, line 1: i'd add a test that focusses on specificity overriding it's "typical" usage e.g. paddingTop > padding

joshuaellis (changes requested):
Looking good so far. This only works with spaces, how do you plan to extend it to work with other props?

packages/design-system/src/helpers/__tests__/handleResponsiveValues.test.ts
madhurisandbhor: Correct, CSS specificity is not taken care here so it wont work, we have to change the order for tests to work

packages/design-system/src/helpers/__tests__/handleResponsiveValues.test.ts
madhurisandbhor: To simplify, when we introduce array there are 4 possibilities, so I thought of just mapping to all 4 logical props, instead of just 2 by checking if there are only 2 values, otherwise map to 4.

packages/design-system/src/helpers/__tests__/handleResponsiveValues.test.ts
madhurisandbhor: but here its just returning the style string, for specificity it seems to work the way we order values

packages/design-system/src/helpers/__tests__/handleResponsiveValues.test.ts
joshuaellis: okay sounds good, just make sure to write all this in the mapping object so everyone else in the future knows! :)

packages/design-system/src/helpers/__tests__/handleResponsiveValues.test.ts
joshuaellis: i just meant more tests like the one i've pointed out is broken 😄

packages/design-system/src/components/Box/Box.tsx
```diff
@@ -234,48 +238,53 @@ const Box = forwardRef(<C extends React.ElementType = 'div'>(props: BoxProps<C>,
...
+      ? `${theme.breakpoints.medium} { display: none; } ${theme.breakpoints.large} { display: inherit}`
+      : undefined}
+  ${({ theme, $hiddenS }) =>
+    $hiddenS
+      ? `${theme.breakpoints.small} { display: none; } ${theme.breakpoints.medium} { display: inherit}`
+      : undefined}
+  ${({ theme, $hiddenXS }) =>
+    $hiddenXS
+      ? `${theme.breakpoints.initial} { display: none; } ${theme.breakpoints.small} { display: inherit}`
+      : undefined}
```
joshuaellis, line 287: I'd get rid of this and call it a breaking change. People can now pass `display` with a responsive object.

packages/design-system/src/components/Box/Box.tsx
```diff
@@ -321,6 +330,38 @@ const StyledBox = styled.div<PropsToTransientProps<TransientBoxProps>>`
...
+      'border-radius': props.$hasRadius ? theme.borderRadius : props.$borderRadius,
+      'border-style': props.$borderStyle,
+      'border-width': props.$borderWidth,
+      'border-color': generateResponsiveValues(theme.colors, props.$borderColor),
+      border: getBorderProp({
+        theme,
+        $borderColor: props.$borderColor,
+        $borderStyle: props.$borderStyle,
+        $borderWidth: props.$borderWidth,
+      }),
```
joshuaellis, line 360: this is confusing, in `handleResponsiveValues` you extract values from the spaces object of the theme but here you handle other props like colors. I think you should be doing them in the same place everywhere either in this function of the other to be consistent. You could alternatively, create a pipe function where you:
1. get the value from the theme or use the value provided (to allow CSS values)
2. pass to the repsonsive function to extract and create the media queries
food for thought.

packages/design-system/src/themes/common-theme.ts
joshuaellis: these are still here, lets remove them 😄

packages/design-system/src/components/Box/Box.tsx
madhurisandbhor: Yes, I realised it yesterday and I will generate responsive values at single place. 
With pipe wont it be separate media query per prop which we dont want?

joshuaellis: yeah you're probably right.

packages/design-system/src/themes/common-theme.ts
```diff
@@ -14,11 +14,13 @@ export interface CommonTheme {
     tooltip: 1000;
   };
   spaces: ['0px', '4px', '8px', '12px', '16px', '20px', '24px', '32px', '40px', '48px', '56px', '64px'];
-  borderRadius: '4px';
-  mediaQueries: {
-    tablet: string;
-    mobile: string;
+  breakpoints: {
+    initial: string;
```
jhoward1994, line 18: is it worth using something like

type MediaQueryString = `@media(${'min-width' | 'max-width'}: ${number}px)`;

to be more specific?

packages/design-system/src/helpers/theme.ts
```diff
@@ -3,11 +3,11 @@ import { DefaultTheme } from 'styled-components';
 import { isKeyOf } from './objects';
 
 export function extractStyleFromTheme<TKey extends keyof DefaultTheme, TSection extends DefaultTheme[TKey]>(
-  themeSection: TSection,
+  themeSection: TSection | null | undefined,
   key: string | number | symbol | undefined,
```
jhoward1994, line 7: Can we do this?

```suggestion
  themeSection?: TSection | null,
  key?: string | number | symbol,
```

jhoward1994:
Should we remove this? https://github.com/strapi/design-system/blob/c7ebedd626ddb8247ef98fd7dadf543ad8f6c031/docs/stories/Flex.mdx#L22-L23

packages/design-system/src/helpers/handleResponsiveValues.ts
```diff
@@ -1,91 +1,272 @@
...
-  theme: DefaultTheme,
-) => {
-  if (!value) {
-    return undefined;
+  | 'margin'
+  ? ResponsiveProperty<DefaultThemeOrCSSProp<T, K> | Array<DefaultThemeOrCSSProp<T, K>>>
+  : ResponsiveProperty<DefaultThemeOrCSSProp<T, K>>;
+
+/**
+ * this should ONLY ever be CSS property names, never shorthands or alaises.
```
jhoward1994, line 32:
```suggestion
 * this should ONLY ever be CSS property names, never shorthands or aliases.
```

docs/stories/Box.stories.tsx
```diff
@@ -12,44 +12,83 @@ type Story = StoryObj<typeof Box>;
...
     <Box
-      padding={[11, 6, 1]}
-      background="primary700"
-      shadow="filterShadow"
-      hiddenXS
-      borderColor="danger600"
+      paddingTop={{ initial: '4rem', medium: '14px', large: 6 }}
+      padding={['2rem', '4rem', '8rem']}
+      marginTop={{ initial: 1, medium: '1rem', large: '22px' }}
+      margin={{ initial: 2, medium: [2, '4px'], large: [6, '4px', 8] }}
```
jhoward1994, line 29: I'm a bit confused by the array breakpoint types and how they're applied.

I'm understanding it as e.g. the large breakpoint here would become:

'margin-block-start': 6, 'margin-inline-end': '4px', 'margin-block-end': 8, 'margin-inline-start': 6 (again?)

is that right?

packages/design-system/src/helpers/theme.ts
madhurisandbhor: there is a subtle difference I think, but in this case we can do optional way as well.

simotae14 (approved):

PR #1738, opened 2024-06-10
fix: icon button sizes & update button docs

joshuaellis:
### What does it do?

* refactor LinkButton to just be a wrapper of Button (less bugs)
* refactor IconButton to be more powerful & correct it's sizes
* update documentation for all "Buttons"

### Related issue(s)/PR(s)

* resolves DX-1379
* resolves strapi/strapi#16378


docs/stories/04-components/IconButton.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+- [Overview](#overview)
+- [Usage](#usage)
+- [Props](#props)
+- [Variants](#variants)
+- [Button Groups](#button-groups)
+
+## Overview
+
+A type of button that does not have a visible label but instead an icon to denote it's action. The component does
```
remidej, line 18:
```suggestion
A type of button that does not have a visible label but instead an icon to denote its action. The component does
```

docs/stories/04-components/LinkButton.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+<Canvas of={LinkButtonStories.Base} />
+
+## Usage with other routing libraries
+
+### React Router
+
+To use the LinkButton component with a routing library (e.g. react-router-dom), you'll need to pass the `NavLink`
+component to the `tag` prop in order to replace the default HTML anchor `<a />`. You'll now be able to pass all
+`NavLink` props.
```
remidej, line 37: Are the types smart enough now to understand the props of the tags are expected? I believe that wasn't the case in v1

docs/stories/04-components/LinkButton.mdx
```diff
@@ -0,0 +1,63 @@
...
+import { NavLink } from 'react-router-dom';
+
+<LinkButton tag={NavLink} to="/home">
+  Home
+</LinkButton>;
+```
+
+### NextJS
+
+For NextJS, you'll need to wrap the `LinkButton` with the `NextLink` component
```
remidej, line 50: I wonder why that is? IIRC next no longer requires you to put an anchor inside NextLink

docs/stories/04-components/TextButton.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+<Canvas of={TextButtonStories.EndIcon} />
+
+## Usage with other routing libraries
+
+### React Router
+
+To use the LinkButton component with a routing library (e.g. react-router-dom), you'll need to pass the `NavLink`
+component to the `tag` prop in order to replace the default HTML anchor `<a />`. You'll now be able to pass all
+`NavLink` props.
```
remidej, line 59: Does this mean NavLink will by default render an anchor if you don't pass a tag? I'm confused as nothing about TextButton implies link behavior to me. I would expect the HTML to be just a button

packages/design-system/src/components/BaseLink/BaseLink.tsx
```diff
@@ -1,31 +1,43 @@
...
       />
     );
   },
 );
 
+const BaseLink = styled<BoxComponent<'a'>>(Box)`
+  text-decoration: none;
+
+  &:visited {
+    color: inherit;
```
remidej, line 36: Which color would be inherited exactly? Is it the browser's default or one from our DS?

packages/design-system/src/components/IconButton/IconButton.tsx
```diff
@@ -1,45 +1,29 @@
...
-   * @default true
-   */
-  withTooltip?: boolean;
-};
+type IconButtonProps<C extends React.ElementType = 'button'> = FlexProps<C> &
+  Pick<ButtonProps, 'size' | 'variant'> & {
+    children: React.ReactNode;
+    disabled?: boolean;
+    /**
+     * This isn't visually rendererd, but required for accessibility.
```
remidej, line 18:
```suggestion
     * This isn't visually rendered, but required for accessibility.
```

packages/design-system/src/components/IconButton/IconButton.tsx
```diff
@@ -1,45 +1,29 @@
...
+     */
+    label: string;
+    onClick?: React.MouseEventHandler<HTMLButtonElement>;
+    /**
+     * @default true
+     */
+    withTooltip?: boolean;
```
remidej, line 25: When would we not want a tooltip? It seems to me like it allows unaccessible implementations

remidej (commented):
This design system of ours is getting pretty cool 🔥 

Side note, I wish I had access to the Vercel org to see the preview

packages/design-system/src/components/IconButton/IconButton.tsx
joshuaellis: It's still accessible, the implementation of an IconButton wraps `AccessibleIcon` which renders an invisible label. We have many buttons in strapi that are IconButtons without a tooltip at all, you used to do it by passing `aria-label` instead of `label`.

packages/design-system/src/components/BaseLink/BaseLink.tsx
joshuaellis: the DS, so if the color is `primary600` it'll stay as that 👍🏼

docs/stories/04-components/TextButton.mdx
joshuaellis: Yeah I don't like the name either, but i'm just matching Product naming. The idea for them supposedly is not to be used as buttons but only for navigation 🤪 we can raise this with product and deprecate during v2 lifecycle if they agree?

docs/stories/04-components/LinkButton.mdx
joshuaellis: It does indeed 🪄

docs/stories/04-components/LinkButton.mdx
remidej: The change happened [between Next 12 and 13](https://nextjs.org/docs/app/building-your-application/upgrading/app-router-migration#link-component). Since the latest is 15 I think we can put the modern version. So based on their [current docs](https://nextjs.org/docs/pages/api-reference/components/link) I think this should work:

```jsx
import { LinkButton } from '@strapi/design-system';
import NextLink from 'next/link';


<LinkButton tag={NextLink} href="/home">Home</LinkButton>
```

joshuaellis: well that's even easier ⭐

docs/stories/04-components/TextButton.mdx
remidej: that sounds good!

joshuaellis:
> Side note, I wish I had access to the Vercel org to see the preview

did you click visit preview? i've tweaked some settings and using incognito i can still see the preview, so might work now?

remidej:
>did you click visit preview? i've tweaked some settings and using incognito i can still see the preview, so might work now?

My bad, I was following the link from the CI checks, but the direct link in the comment works perfectly

remidej (approved):
nice!

packages/design-system/src/components/LinkButton/LinkButton.tsx
```diff
@@ -1,90 +1,30 @@
...
+import { Button, ButtonProps } from '../Button';
 
-type LinkButtonProps<C extends React.ElementType = 'a'> = BaseLinkProps<C> & {
-  disabled?: boolean;
-  endIcon?: React.ReactNode;
-  size?: ButtonSize;
-  startIcon?: React.ReactNode;
-  variant?: ButtonVariant;
-};
+type LinkButtonProps<C extends React.ElementType = typeof BaseLink> = ButtonProps<C>;
```
madhurisandbhor, line 8: this is not clear, 

why this is not `type LinkButtonProps<C extends React.ElementType = "button"> = ButtonProps<C>;`

how this LinkButton wrapped exactly? even this syntax is confusing for me, `<Button<typeof BaseLink>`. If you could please elaborate or discuss on call.

joshuaellis: Because the `tag` component used in the `Button` component below is `BaseLink`. In our polymorphic components, the C generic denotes the element which corresponds directly to the `tag` prop. 

The `<Button<typeof BaseLink>` is because all components are just functions, and the `Button` component is a generic function that takes it's element type. So in this scenario because we're working with two layers of polymoprhism we add the component as the generic that our props correspond too to avoid type errors.

madhurisandbhor: ohk, I will check how this polymorphism works here, thanks

docs/stories/04-components/IconButton.stories.tsx
```diff
@@ -0,0 +1,166 @@
...
+      <IconButton label="Edit">
+        <Icons.Pencil />
+      </IconButton>
+      <IconButton label="Clone">
+        <Icons.Duplicate />
+      </IconButton>
+      <IconButton label="Delete" variant="danger">
+        <Icons.Trash />
+      </IconButton>
+    </IconButtonGroup>
```
madhurisandbhor, line 163: You will need to add source code as its incorrect for `all variants` and `group`. Also it would have been great if we also show withTooltip example in one of the example, so it replaces the default label?

Issue #1724, opened 2024-05-10
[bug]: Badge add transparent as possible backgroundColor value and find a way to overwrite the textColor with the prop passed


simotae14:
### What version of strapi-design-system are you using?

DS - v2.0.0-beta.3
Node - v20.10.0
NPM - v10.2.3
YARN - v4.0.1
Browser - Chrome

### What's Wrong?

it is impossible to pass transparent as backgroundColor (we have a Type issue)

And if you try to pass a value to the textColor prop it doesn't apply

### To Reproduce

Create a Badge content in the cms and try to pass the two props, transparent is not allowed as backgroundColor and the textColor doesn't work

### Expected Behaviour

I think transparent needs to be an option for the the background color and the textColor needs to be applied

cybertron288:
Hi @simotae14 @joshuaellis,
I have raised the PR for above issue, https://github.com/strapi/design-system/pull/1740 can you please review the changes.

PR #1740, opened 2024-06-10
Fix/badge style props

cybertron288:
## What does it do?
This PR introduces two enhancements to the Badge component #1724 

### Text Color Prop for Badge Content:

Added a new prop textColor to the Badge component, allowing the text color to be customized.
Users can now set the textColor prop, which will dynamically change the text color of the badge.

### Background Color Enhancement:

Ensured that transparent is allowed as a value for backgroundColor.
This issue seems to have already been addressed previously. I'm not able to reproduce, working file with backgroundColor: transparent prop

## Why is it needed?
Customization: Allowing users to change the text color of the badge and background color as backgroundColor.

### How to test it?

Text Color Prop:
Add a Badge component in the CMS and set the textColor prop to various colors.
Verify that the text color of the badge changes accordingly.

Background Color Validation:
Try setting transparent as the backgroundColor value for the Badge component.
Ensure that it is not accepted and does not cause any UI issues.

Related issue(s)/PR(s)
#1724


madhurisandbhor:
Fixed here, https://github.com/strapi/design-system/pull/1764

PR #1749, opened 2024-06-21
fix: filters shouldn't be removed after clicking on filter tag

madhurisandbhor:
### What does it do?

Filter click action passed on icon instead of tag itself, adjusted styles to show cursor pointer on hover

### Why is it needed?

filters shouldn't be removed after clicking on filter tag

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/design-system/issues/679

### DIFF
--- a/packages/design-system/src/components/Tag/Tag.tsx
+++ b/packages/design-system/src/components/Tag/Tag.tsx
@@ -3,21 +3,32 @@ import * as React from 'react';
 import { styled } from 'styled-components';
 
 import { Box, type BoxComponent } from '../Box';
-import { Flex, FlexComponent, FlexProps } from '../Flex';
+import { Flex, FlexProps } from '../Flex';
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/Tag/Tag.tsx
```diff
@@ -21,28 +32,28 @@ export const Tag = ({ children, icon, disabled = false, onClick, ...props }: Tag
       background={disabled ? 'neutral200' : 'primary100'}
       color={disabled ? 'neutral700' : 'primary600'}
       paddingLeft={3}
-      paddingRight={3}
-      onClick={handleClick}
+      paddingRight={1}
```
remidej, line 35: I think TagWrapper no longer needs to be a button

remidej (approved):

remidej (approved):

PR #1753, opened 2024-06-26
fix: move the Modal Content inside the Overlay component and change the background rule of the Overlay

simotae14:
### What does it do?

It fixes the wrong behaviour in the Modal Content component, the children components are not scrollable, I moved the ContentImpl as a child of the Overlay component and I changed the background rule in the Overlay style to don't affect the opacity of the Overlay children

### Why is it needed?

Because otherwise scrolling is not working for example in dropdown used in the Modal body.
Here an example of the issue

https://github.com/strapi/design-system/assets/2589748/f6177b1d-1de9-459b-be63-5e46fcba7c26

and here the solution

https://github.com/strapi/design-system/assets/2589748/b47e2a95-1624-4195-b844-dc7f22a9583d

### How to test it?

You can test the solution directly in the Modal DS storybook https://64f9707bb212b1f1332bfd97-ehiswajggi.chromatic.com/?path=/story/components-modal--base

### Related issue(s)/PR(s)

fixes https://github.com/strapi/strapi/issues/20599

### DIFF
--- a/.changeset/short-pets-attack.md
+++ b/.changeset/short-pets-attack.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': minor
+---
+
+fixed Modal Content scroll issues
--- a/.changeset/short-pets-attack.md
+++ b/.changeset/short-pets-attack.md
@@ -1,5 +1,5 @@
 ---
-'@strapi/design-system': minor
+'@strapi/design-system': patch
 ---
 
 fixed Modal Content scroll issues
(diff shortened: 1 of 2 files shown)


.changeset/short-pets-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@strapi/design-system': minor
```
remidej, line 2: shouldn't it be a patch?

remidej (dismissed):

remidej (approved):

numr:
@alexandrebodin 
Has this issue been completely fixed? I still have problems using the latest design-system2.0.0-rc.14. I can't scroll through <MultiSelect> in the <Dialog.Content>, as shown in the screenshot I sent

https://github.com/user-attachments/assets/2eb5a2e7-b48f-47ef-9d50-e060b9b233bb

PR #1756, opened 2024-07-09
chore: set changesets to beta

jhoward1994:
### What does it do?

If we want to release a new v2 Beta (Beta7) for this change

https://github.com/strapi/design-system/commit/daae1f0b4da19febf0804783195c9e1c6245b82c

we should change changesets from RC -> Beta mode so it can handle things correctly in https://github.com/strapi/design-system/pull/1755


jhoward1994:
Closing as we are staying with RC releases

PR #1758, opened 2024-07-16
fix: simple menu with icon

madhurisandbhor:
### What does it do?

Simple menu with icon should work now as expected.

### Related issue(s)/PR(s)

Fixes https://github.com/strapi/design-system/issues/1418


packages/design-system/src/components/SimpleMenu/Menu.tsx
```diff
@@ -24,23 +25,50 @@ const MenuRoot = DropdownMenu.Root;
...
+    const props = {
+      ...rest,
+      ref,
+      type: 'button' as const,
+      variant: 'ghost' as const,
+      paddingTop: size === 'S' ? 1 : 2,
+      paddingBottom: size === 'S' ? 1 : 2,
+      paddingLeft: size === 'S' ? 3 : 4,
+      paddingRight: size === 'S' ? 3 : 4,
+    };
```
remidej, line 59: optional but maybe we can cast a type here where it's defined? It would be convenient if someone needs to modify this object later

packages/design-system/src/components/SimpleMenu/SimpleMenu.tsx
```diff
@@ -65,9 +63,7 @@ const SimpleMenu = React.forwardRef<HTMLButtonElement, SimpleMenuProps>(
 
     return (
       <Menu.Root onOpenChange={handleOpenChange}>
-        <Menu.Trigger ref={composedRef} {...props}>
-          {props.label}
-        </Menu.Trigger>
+        <Menu.Trigger ref={composedRef} {...props} />
```
remidej, line 66: In the storybook docs I still see this syntax:

```jsx
<Menu.Trigger>Actions</Menu.Trigger>
```

should it be updated?

packages/design-system/src/components/SimpleMenu/Menu.tsx
```diff
@@ -24,23 +25,50 @@ const MenuRoot = DropdownMenu.Root;
...
+  tag: typeof IconButton;
+  icon: React.ReactNode;
+};
+
+type TriggerProps = TriggerPropsWithButton | TriggerPropsWithIconButton;
 
 const MenuTrigger = React.forwardRef<HTMLButtonElement, TriggerProps>(
-  ({ size, endIcon = <CaretDown width="0.6rem" height="0.4rem" aria-hidden />, ...props }, ref) => {
+  (
+    { label, size, endIcon = <CaretDown width="0.6rem" height="0.4rem" aria-hidden />, tag = Button, icon, ...rest },
```
remidej, line 47: Could you double check that the caret icon is the right size? It looks really tiny in the storybook example

madhurisandbhor: yeah..updated!

remidej (approved):

Issue #1626, opened 2024-03-07
[bug]: ComboBox doesn't show `noOptionsMessage` when there are no options to display


jhoward1994:
### What version of strapi-design-system are you using?

DS - v1.16.0
Node - v20.8.0
NPM - v10.1.0
Browser - Brave (Chromium)

### What's Wrong?

<img width="322" alt="image" src="https://github.com/strapi/design-system/assets/48524071/84dbe73b-ca86-4cde-a6ff-b24848462bfb">

In a ComboBox with 0 items to display I still see a small bar

### To Reproduce

Using the combobox within the strapi content manager to add relations.

If there are no entries for the target content type you should see this behaviour

### Expected Behaviour

I think I should see whatever is defined in `noOptionsMessage`

cybertron288:
Hi @simotae14 @joshuaellis,
I have raised the PR for above issue, #1743  please review the changes.

PR #1762, opened 2024-07-23
fix: comboBox does not show noOptionsMessage when there are no options

cybertron288:
## What does it do?
This PR fixes when there are no options available in ComboBox then noOptionsMessage should show #1626 

## Why is it needed?
when there are no options in basic combobox it shows small bar. instead it should display `noOptionsMessage`

## How to test it?
in combobox storybook code, comment or remove all ComboboxOption of Basic combobox, it should display **No Fruit(s) found** 
Screenshot 2024-06-13 at 9 09 20 PM

## Related issue(s)/PR(s)
The related issue is mentioned here: #1626


cybertron288:
Hi @madhurisandbhor, will make it open once I update the storybook.

cybertron288:
Hi @madhurisandbhor, added docs and updated storybook, let me know anything else is required.

madhurisandbhor:
It would be great if you add unit test instead of story to show `no options`, its not really adding any meaning to show it as a story. Also please add a changeset as suggested [here](https://github.com/strapi/design-system/pull/1762#issuecomment-2245999416)

madhurisandbhor:
fixed by https://github.com/strapi/design-system/pull/1770

PR #1763, opened 2024-07-24
fix: modals broken in safari

remidej:
### What does it do?

Fixes an issue where modals were not displaying their footer (when non scrollable) or their entire body (when scrollable) in Safari.

You can see the issue if you open the current Modal storybook in Safari:

CleanShot 2024-07-24 at 12 15 53@2x
CleanShot 2024-07-24 at 12 17 19@2x

You can also see the issue in the CMS on v5/main using Safari. Try adding an asset to the media library, or replacing an asset, or adding a new field in the CTB, or editing an input in configure the view.

### How to test it?

You can test the fix from Storybook by checking the stories in my screenshots above in Safari. Also make sure that the scrolling works (you may need to shrink your browser height to see it)

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/20556

### DIFF
--- a/.changeset/large-dots-cheer.md
+++ b/.changeset/large-dots-cheer.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': patch
+---
+
+fix modals not displaying properly in safari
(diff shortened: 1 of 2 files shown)


remidej:
@simotae14 I think your #1759 actually fixes that 🙃 the storybook version I tested didn't have your changes yet

Issue #1768, opened 2024-07-30
[bug]: missing border radius on popover


madhurisandbhor:
### What version of design-system are you using?

"@strapi/design-system": "2.0.0-rc.7"

### What's Wrong?

Missing border radius on popover

### To Reproduce

Check popover border radius

### Expected Behaviour

Should have default border radius of 4px

Issue #1767, opened 2024-07-30
[bug]:  menu trigger with icon button having variant ghost has black border on active state


madhurisandbhor:
### What version of design-system are you using?

"@strapi/design-system": "2.0.0-rc.7"

### What's Wrong?

When menu trigger is used to show button with icon having variant ghost has black border styled on active state.

### To Reproduce

Try to use menu trigger with icon button and variant ghost, check active style

### Expected Behaviour

Black border should not be visible for variant ghost

PR #1769, opened 2024-07-30
fix: multiple UI styles fixes

madhurisandbhor:
### What does it do?
1. Ghost button active border style removed
2. On popover default border radius added
3. Field input padding updated to show correct highlight in autocomplete

### Related issue(s)/PR(s)

Fixes: https://github.com/strapi/design-system/issues/1768, https://github.com/strapi/design-system/issues/1767

### DIFF
--- a/packages/design-system/src/components/Field/Field.tsx
+++ b/packages/design-system/src/components/Field/Field.tsx
@@ -206,19 +206,6 @@ const InputElement = styled.input<{ $size: InputProps['size'] }>`
   width: 100%;
   background: inherit;
 
-  ${(props) => {
-    switch (props.$size) {
-      case 'S':
-        return css`
-          padding-block: ${props.theme.spaces[1]};
-        `;
-      default:
-        return css`
-          padding-block: ${props.theme.spaces[2]};
-        `;
-    }
-  }};
-
   ::placeholder {
     color: ${({ theme }) => theme.colors.neutral500};
     opacity: 1;
@@ -253,11 +240,13 @@ const InputWrapper = styled<FlexComponent>(Flex)<{
         return css`
           padding-inline-start: ${props.$hasLeftAction ? props.theme.spaces[3] : props.theme.spaces[4]};
           padding-inline-end: ${props.$hasRightAction ? props.theme.spaces[3] : props.theme.spaces[4]};
+          padding-block: ${props.theme.spaces[1]};
         `;
       default:
         return css`
           padding-inline-start: ${props.$hasLeftAction ? props.theme.spaces[3] : props.theme.spaces[4]};
           padding-inline-end: ${props.$hasRightAction ? props.theme.spaces[3] : props.theme.spaces[4]};
+          padding-block: ${props.theme.spaces[2]};
         `;
     }
   }}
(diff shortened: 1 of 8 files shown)


packages/design-system/src/components/Field/Field.tsx
```diff
@@ -253,11 +240,13 @@ const InputWrapper = styled<FlexComponent>(Flex)<{
         return css`
           padding-inline-start: ${props.$hasLeftAction ? props.theme.spaces[3] : props.theme.spaces[4]};
           padding-inline-end: ${props.$hasRightAction ? props.theme.spaces[3] : props.theme.spaces[4]};
+          padding-block: ${props.theme.spaces[1]};
         `;
       default:
         return css`
           padding-inline-start: ${props.$hasLeftAction ? props.theme.spaces[3] : props.theme.spaces[4]};
           padding-inline-end: ${props.$hasRightAction ? props.theme.spaces[3] : props.theme.spaces[4]};
+          padding-block: ${props.theme.spaces[2]};
```
remidej, line 249: Putting the padding outside of the input element creates a problem, as not the visual space dedicated to the input can focus it, which is an accessibility issue:

CleanShot 2024-07-31 at 11 36 23@2x

In terms of aesthetics it also doesn't look great with the autofill highlight from password managers:
CleanShot 2024-07-31 at 11 29 11@2x

remidej (commented):
LGTM for popovers and buttons, just have some comments for the Input

packages/design-system/src/components/Field/Field.tsx
remidej: In my opinion the way to get a11y right is to put all the padding on the input element.

I think we should be able to render the endActions using position absolute on top of the input, and the reserve the space underneath by adapting the input's paddingRight based on the presence or absence of an endAction

madhurisandbhor: I am not sure to have `absolute` position here as `endAction` could be anything and not just an icon

remidej: For the paddingRight value, I think you can compute it, by measuring the width of the endAction using a ref, and adding a small gap so the content can't be stuck to it

Then you can add a div wrapper around endAction, set set this div to absolute, and `top: 50%; transformY(-50%)` to ensure vertical centering regardless of the size

madhurisandbhor: right dynamically 👍

packages/design-system/src/components/Field/Field.tsx
```diff
@@ -154,13 +155,27 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
...
+    const inputElementRef = React.useRef<HTMLInputElement>(null);
+    const inputRef = useComposedRefs(inputElementRef, ref);
 
     const handleChange: React.ChangeEventHandler<HTMLInputElement> = (e) => {
       if (!disabled && onChange) {
         onChange(e);
       }
     };
 
+    React.useEffect(() => {
```
remidej, line 168: I think this is a good use case for useLayoutEffect instead of useEffect, see this guide: https://react.dev/reference/react/useLayoutEffect#measuring-layout-before-the-browser-repaints-the-screen

So that we can have the right width from the very first render

madhurisandbhor:
@remidej Thank you for the improvements!

remidej (commented):
I'm seeing a paddingRight a bit too big on the right side when there's no endAction (should be 16px not 24):

CleanShot 2024-08-01 at 09 43 46@2x

Also it would be nice to add a field with end action example in storybook

remidej (approved):

PR #1773, opened 2024-08-09
fix: combobox with no options and autocomplete none

madhurisandbhor:
### What does it do?
Combobox with no options and `autocomplete: none` , no options value should be displayed.

### Related issue(s)/PR(s)

https://github.com/strapi/design-system/issues/1626


remidej (approved):
👍

PR #1774, opened 2024-08-13
fix: accordian style

madhurisandbhor:
### What does it do?

Fix accordian style, https://github.com/strapi/strapi/issues/20789


lucasboilly:
Hey @madhurisandbhor, here are the remaining issues:

- After clicking on it, if my pointer leaves the hover state disappears but a blue border remains
https://github.com/user-attachments/assets/3aa90f7b-a74f-4c0b-ae7d-e7b4760a6abe
- When hovering a disabled one, a blue border appears but it shouldn't be the case
https://github.com/user-attachments/assets/83df1601-40ca-49be-9813-f8e209dfc078
- We shouldn't have icons on the "L" variant, because we'll need to remove them everywhere in the settings
CleanShot 2024-08-13 at 14 55 12@2x
- The icon buttons in this variant don't have the right colors, maybe is there an override?
CleanShot 2024-08-13 at 14 55 54@2x
- In a group we would need a border top or bottom on each accordion to separate them
CleanShot 2024-08-13 at 14 56 57@2x

Feranchz:
@Lucas, should be ready now. Just a few comments:

> * After clicking on it, if my pointer leaves the hover state disappears but a blue border remains
>   https://github.com/user-attachments/assets/3aa90f7b-a74f-4c0b-ae7d-e7b4760a6abe

I think this is right. It's true that your pointer leaves but the focus is still on the element, the border disappear when you change the focus to another part, I think it's the right way.

> * The icon buttons in this variant don't have the right colors, maybe is there an override?

Yes, it's a override that only happens on the Storybook, is not happening at the CMS so I think is fine.

remidej (dismissed):
Code changes LGTM, but let's wait for Lucas' design review

remidej (approved):

alexandrebodin:
@remidej I think you can go ahead and merge since madhuri isn't here anymore

remidej:
Yes @Feranchz took over the work, I just want Lucas to confirm the PR fixes the issue

lucasboilly:
Hey @Feranchz :)
- The behavior was different in v4 but maybe it's better like this
- Ok!

@remidej that's all good, let's merge it 👍

PR #1779, opened 2024-08-28
fix: button icons position and width

HichamELBSI:
### What does it do?

Fix the alignment of the start and end icons.

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

https://github.com/strapi/strapi/issues/20986

### DIFF
--- a/packages/design-system/src/components/Button/Button.tsx
+++ b/packages/design-system/src/components/Button/Button.tsx
@@ -67,7 +67,7 @@ const Button = forwardRef(
         {...props}
       >
         {(startIcon || loading) && (
-          <Flex tag="span" aria-hidden flex="1 0 1.2rem">
+          <Flex tag="span" aria-hidden>
             {loading ? <LoaderAnimated /> : startIcon}
           </Flex>
         )}
@@ -77,7 +77,7 @@ const Button = forwardRef(
         </Typography>
 
         {endIcon && (
-          <Flex tag="span" aria-hidden flex="1 0 1.2rem">
+          <Flex tag="span" aria-hidden>
             {endIcon}
           </Flex>
         )}


alexandrebodin (commented):
Thanks for the PR. Aren't we missing a margin right & left for start & end icons now ?

HichamELBSI:
> Thanks for the PR. Aren't we missing a margin right & left for start & end icons now ?

If this fix is what we wanted to do visually @lucasboilly, there is no need to add a margin, there is a 8px gap between icons and the text. Now that the text is taking 100% of the remaining space, the text will be centered by default. I don't understand why there was a flex 1 on the icons (cc: @simotae14 @remidej if you have any context about it 🤷🏽‍♂️ ) ⬇️ Screenshots of the use cases with the new update.

Start icon - normal size
Screenshot 2024-08-29 at 11 13 10

Start icon - fullwidth (which is used in the CM list view filters)
Screenshot 2024-08-29 at 11 13 06

End icon - normal size
Screenshot 2024-08-29 at 11 15 25
Screenshot 2024-08-29 at 11 15 30

Start + End icon
Screenshot 2024-08-29 at 11 12 52
Screenshot 2024-08-29 at 11 12 58

HichamELBSI:
> From @lucasboilly comment here https://www.notion.so/strapi/The-button-in-the-filters-of-the-list-view-is-broken-029d9bac6fa64046ae56782fce9496c0?d=7eb956cf8eea434199d2dfcf6483ca65 I think we are expecting the icon to always be attached to the text even in fullwidth so I think we have to remove the flex & not sure the gap will still work but I let you handle that :D

You're right, my bad. It should be good now, the gap seems to be working properly. 
Screenshot 2024-08-29 at 11 28 56
Screenshot 2024-08-29 at 11 28 52
Screenshot 2024-08-29 at 11 28 45
Screenshot 2024-08-29 at 11 28 40
Screenshot 2024-08-29 at 11 28 36
Screenshot 2024-08-29 at 11 28 23

lucasboilly:
It looks great, thanks @HichamELBSI!

simotae14 (approved):

alexandrebodin (approved):
LGTM 🚀

alexandrebodin:
@HichamELBSI I think you need to add a changeset so this can trigger a new release ?

PR #1788, opened 2024-09-20
fix: props table not showing

Bassel17:
### What does it do?

- Upgrade Storybook packages to 8.3.1

- remove @storybook/blocks patch version from the root package.json

- switch TableType component with ArgTypes in storybook docs

### Why is it needed?

- `TypeTable` doesn't exist anymore no props table is generated in the Storybooks
- The resolution patch version of `@storybook/blocks` doesn't seem to be needed, we were already using `8.1.6` version

### How to test it?

- run storybook, check the Docs, you will see props tables for a lot of the components

***Note***
Will be working on a strategy to improve the docs better, as for some component `ArgTypes` did not manage to automatically generate a table of the props


butcherZ (approved):
LGTM! :D

remidej (approved):
so nice to have this working 🫶

Issue #1791, opened 2024-09-18
[Design] Media Library missing cursor pointers on the filters


yanniskadiri:
The cursor pointers are missing on the hover state of both the checkbox and pre-filtered list

![Image](https://github.com/user-attachments/assets/b32b2678-e69a-45e4-81fc-37a6ef94ce07)

hanpaine:
Fix in progress with community member

PR #1793, opened 2024-09-27
fix: added cursor pointer for checkbox and pre-filled list

Sam-Phillemon9493:
What does it do?
Fixed the Media Library's missing cursor pointers issue on the filters

Why is it needed?
To fix issue https://github.com/strapi/design-system/issues/1791: Media Library missing cursor pointers on the filters

How to test it?

- Open the Media Library
- Hover over the filters, it should show the cursor pointers.

Related issue(s)/PR(s)
https://github.com/strapi/design-system/issues/1791

### DIFF
--- a/.changeset/sweet-timers-help.md
+++ b/.changeset/sweet-timers-help.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': major
+---
+
+patch for hovering effect
--- a/.changeset/sweet-timers-help.md
+++ b/.changeset/sweet-timers-help.md
@@ -1,5 +1,5 @@
 ---
-'@strapi/design-system': major
+'@strapi/design-system': patch
 ---
 
 patch for hovering effect
--- a/packages/design-system/src/components/Checkbox/Checkbox.tsx
+++ b/packages/design-system/src/components/Checkbox/Checkbox.tsx
@@ -74,6 +74,7 @@ const CheckboxRoot = styled(Checkbox.Root)`
   align-items: center;
   // this ensures the checkbox is always a square even in flex-containers.
   flex: 0 0 2rem;
+  cursor: pointer;
 
   &[data-state='checked'],
   &[data-state='indeterminate'] {
--- a/packages/design-system/src/components/Checkbox/Checkbox.tsx
+++ b/packages/design-system/src/components/Checkbox/Checkbox.tsx
@@ -51,26 +51,15 @@ const CheckboxEl = React.forwardRef<CheckboxElement, CheckboxElProps>(
     const composedRefs = useComposedRefs(checkboxRef, forwardedRef);
 
     return (
-      // <CheckboxWrapper>
       <CheckboxRoot ref={composedRefs} checked={checked} onCheckedChange={setChecked} {...props}>
         <Checkbox.Indicator style={{ display: 'inline-flex' }}>
           {checked === true ? <CheckIcon width="1.6rem" fill="neutral0" /> : null}
           {checked === 'indeterminate' ? <Minus fill="neutral0" /> : null}
         </Checkbox.Indicator>
       </CheckboxRoot>
-      // </CheckboxWrapper>
     );
   },
 );
-
-// const CheckboxWrapper = styled.div`
-//   position: relative;
-//   width: 44px;
-//   height: 44px;
-//   display: inline-flex;
-//   justify-content: center;
-//   align-items: center;
-// `;
 
 const CheckboxRoot = styled(Checkbox.Root)`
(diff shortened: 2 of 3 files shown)


Sam-Phillemon9493:
Hi @remidej , As discussed in this PR [21460](https://github.com/strapi/strapi/pull/21460), here’s the new PR for the issue. Please let me know if you have any feedback on this change.

packages/design-system/src/components/Checkbox/Checkbox.tsx
```diff
@@ -74,6 +74,7 @@ const CheckboxRoot = styled(Checkbox.Root)`
   align-items: center;
   // this ensures the checkbox is always a square even in flex-containers.
   flex: 0 0 2rem;
+  cursor: pointer;
```
remidej, line 77: I think this style is too high up, it applies the pointer cursor to the whole area and not just to the checkbox:

CleanShot 2024-09-27 at 12 38 46

remidej (commented):
thanks for moving the PR 🙌

packages/design-system/src/components/Checkbox/Checkbox.tsx
Sam-Phillemon9493: @remidej , I’ve implemented the changes,  and the hover effect now applies exclusively to the checkbox.

Aurelsicoko: @remidej Should the label be clickable, too? It prevents the user from being too precise

remidej: Yes the label is clickable, but I don't think it should have the pointer cursor, so I like the way it looks in this PR

Aurelsicoko: LGTM 👍

packages/design-system/src/components/Checkbox/Checkbox.tsx
```diff
@@ -96,6 +98,7 @@ const CheckboxRoot = styled(Checkbox.Root)`
     height: 100%;
     min-width: 44px;
     min-height: 44px;
+    pointer-events: none;
```
remidej, line 101: We can't do that because it's not just a visual change about what the cursor looks like, it makes the area not clickable. And as indicated in the comment above we do want to increase the touch target to make it easier to click, especially on mobile.

So I would recommend changing the approach and removing `cursor: pointer` from the root, and instead only applying it to the icon. Then you wouldn't need to change this pseudo element

packages/design-system/src/components/Select/SelectParts.tsx
remidej, line 1: I checked with @lucasboilly and he also wants cursor pointer on the select's options, could you add that too?

packages/design-system/src/components/Checkbox/Checkbox.tsx
Sam-Phillemon9493: @remidej , I tried applying `cursor: pointer` to the icon, but it doesn’t show in the corners of the checkbox. To resolve this, I added it to the Checkbox Indicator, which ensures that it now encompasses the entire area.

Please let me know if you have any additional thoughts or suggestions!

packages/design-system/src/components/Select/SelectParts.tsx
Sam-Phillemon9493: @remidej , I have added the `cursor: pointer` for the `single select`. Could you please let me know if I should apply this to the `multiple select` as well?

remidej:
Sorry I meant to ping @lucasboilly for a final check

lucasboilly:
Hey @Sam-Phillemon9493, it looks great! Thanks for your contribution!

remidej (approved):

remidej (commented):
I forgot this just needs a changeset (something we use for auto geenrated releases and changelogs), could you add one?

`yarn release:add` and mark it as a patch

Sam-Phillemon9493:
> I forgot this just needs a changeset (something we use for auto geenrated releases and changelogs), could you add one?
> 
> `yarn release:add` and mark it as a patch

Hi @remidej , I've added a changeset. Please let me know if anything else is needed.

.changeset/sweet-timers-help.md
```diff
@@ -0,0 +1,5 @@
+---
+'@strapi/design-system': major
```
remidej, line 2:
```suggestion
'@strapi/design-system': patch
```

this would have released v3.0.0 :)

remidej (approved):

PR #1794, opened 2024-10-07
Optional iconProps can be passed to Accordions

jhoward1994:
### What does it do?

Adds an optional prop 'iconProps' to the Accordion component

### Why is it needed?

So we can modify the icon, e.g. add a fill color to it

This came up when converting the SEO plugin to v5 through https://github.com/strapi/strapi-plugin-seo/pull/76

### Related issue(s)/PR(s)

DX-1693

### DIFF
--- a/.changeset/stale-apes-approve.md
+++ b/.changeset/stale-apes-approve.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': minor
+---
+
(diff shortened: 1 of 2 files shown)


butcherZ (approved):
LGTM. I'm able to change the icon's colour with iconProps now.

PR #1369, opened 2023-09-26
fix: inaccurate floating-point calculations

bilal-auz:
### What does it do?

I used the big.js library to overcome the floating-point error that occurs in the NumberInput component when using the increment and decrement functions. The new implementation ensures that arithmetic operations are accurately performed, especially when dealing with decimal numbers.
image

### Why is it needed?
These changes address the issues raised in [Strapi/Strapi repo issue#18138]( https://github.com/strapi/strapi/issues/18138).
In the main project, when trying to fill the restaurant fields, a bug occurs when selecting an averagePrice of 0.06 using the arrow controls in the restaurant fields. The arrows provide inaccurate precision, resulting in 0.0600000001 instead of 0.06 when reaching this value. This issue also randomly affects other values.
See the issue [Strapi/Strapi repo issue#18138]( https://github.com/strapi/strapi/issues/18138) for better understanding

#### The issue demonstration:-
AveragePrice-Bug

### How to test it?
I haven't completed the test yet due to the discovery of some missing testing packages, which can be seen as another issue. As a result, I've chosen to divide the task into two separate steps.

### Related issue(s)/PR(s)
The related issue is mentioned in the main repo of strapi: https://github.com/strapi/strapi/issues/18138

PR #1795, opened 2024-10-09
fix: floating point precision errors

butcherZ:
### What does it do?

- Round resulting value in `increment` and `decrement` functions used in `NumberInput` component
- Pass `onValueChange` function in `NumberInput` story, as it's required for displaying input value

### Why is it needed?

_Because javascript_

`0.1`+`0.2` = `0.30000000000000004`

### How to test it?

- set `step` of `NumberInput` to `0.1`
- increment/decrement the input value using arrow keys
- incrementing `0.1` two times, you should get `0.3` instead of `0.30000000000000004`

### Related issue(s)/PR(s)

#1369 
[#18138](https://github.com/strapi/strapi/issues/18138)
DX-1674


packages/design-system/src/components/NumberInput/NumberInput.tsx
```diff
@@ -78,8 +78,9 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
       const parsedValue = numberParserRef.current.parse(inputValue);
 
       const newValue = isNaN(parsedValue) ? step : parsedValue + step;
+      const roundedValue = parseFloat(newValue.toFixed(10));
```
remidej, line 81: Just wondering where the 10 value comes from?

butcherZ: Yeah, I put 10 as I wasn't sure about the use cases for how many decimals I should keep, but you are right, I will derive the precision from the `step`!

packages/design-system/src/components/NumberInput/NumberInput.tsx
```diff
@@ -68,6 +68,19 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>(
       }
     };
 
+    /**
+     * Conver step to string to get accurate decimal place
+     * then round the parsed value accordingly
+     */
+    const getDecimalPlaces = (step: number): number => {
+      const stepString = step.toString();
+      if (stepString.indexOf('.') === -1) return 0;
```
remidej, line 77: I personally find `.includes('.')` nicer to read for this check

jhoward1994 (approved):
"0.1+0.2 = 0.30000000000000004"

angry-comedy-bang-bang-gif

remidej (approved):

PR #1802, opened 2024-10-28
chore: remove Status showBullet option

remidej:
### What does it do?

Remove the showBullet option from the Status component. It was a leftover from the CMS v4 designs that is not in use anymore. This v2 RC is our chance to get rid of it (see the comment that was in the code)

### Why is it needed?

Cleaning up dead code

### DIFF
--- a/docs/stories/Status.stories.tsx
+++ b/docs/stories/Status.stories.tsx
@@ -13,12 +13,12 @@ type Story = StoryObj<typeof Status>;
 export const Base = {
   render: () => (
     <Flex direction="column" alignItems="stretch" gap={3}>
-      <Status variant="success" showBullet={false}>
+      <Status variant="success">
         <Typography>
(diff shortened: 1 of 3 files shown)


lucasboilly:
@remidej all good, this option is not relevant anymore in v5 :)

markkaylor (commented):
Looks good to me. 

It looks like in the CMS we are always setting it to false but I guess we need a follow up PR over there to remove the prop entirely. Do we need to communicate / align with cloud on this change?

remidej:
Let me check. @HichamELBSI is this okay for the cloud team? I searched your code and it looks like you always pass `showBullet={false}`. So it won't impact your UI but it means you'll have to remove that prop when upgrading to the latest DS RC

HichamELBSI:
Unfortunately this is used in several places in the Cloud. Do we really need to remove this prop 🤔

Issue #1803, opened 2024-10-29
[bug]: Combobox always renders the "No results found" at the end of the options when "autocomplete" is set to "none"


HichamELBSI:
### What version of design-system are you using?

- Strapi Design System: 2.0.0-rc.11

### What's Wrong?

The Combobox is always rendering the "No results found" at the end of the options when "autocomplete" is set to "none"
This can be seen in the storybook.

### To Reproduce

- Render a Combobox component
- Set the "autocomplete" prop to "none"
- The "No results found" will be rendered at the end of the option list

Can be reproduced in the storybook `Combobox` component.

### Expected Behaviour

Since there is no filter with the `autocomplete: none`, the `No results found` shouldn't be rendered. 

![Screenshot 2024-10-29 at 15 35 49](https://github.com/user-attachments/assets/f0954271-62f4-442f-a3ce-c5f554cf2016)

PR #1816, opened 2024-11-07
fix: combobox primitive empty state when autocomplete is none

HichamELBSI:
### What does it do?

This PR fixes the existing issue that always renders the `No result found` (empty state) of the combobox when autocomplete is set to off.

### Why is it needed?

Avoid rendering the `No result found` at the end of the option list (when `autocomplete = none`) which is the case in all Strapi Cloud combobox.

### How to test it?

Added new unit tests to handle this use case

### Related issue(s)/PR(s)

https://github.com/strapi/design-system/issues/1803

### DIFF
--- a/.changeset/blue-humans-sin.md
+++ b/.changeset/blue-humans-sin.md
@@ -1,0 +1,5 @@
+---
+'@strapi/ui-primitives': major
+---
+
+fixes the combobox primitive empty state when autocomplete is set to none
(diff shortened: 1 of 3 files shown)


simotae14 (approved):
looks good to me, well done 👏🏽

mazzucchelli (approved):
🚀

PR #1834, opened 2024-12-12
Prevent tooltip from rendering with empty label

mazzucchelli:
### Why is it needed?

The `label` prop for the `<Tooltip />` component is optional, but it doesn't prevent the tooltip from rendering when it's missing.

Screenshot 2024-12-12 alle 12 13 55

### How to test it?

Create a tooltip with an empty `label` prop

### DIFF
--- a/.changeset/three-eels-clap.md
+++ b/.changeset/three-eels-clap.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': patch
+---
+
+Prevent tooltip from rendering when the label property is not passed
--- a/packages/design-system/src/components/Tooltip/Tooltip.test.tsx
+++ b/packages/design-system/src/components/Tooltip/Tooltip.test.tsx
@@ -1,0 +1,46 @@
+import { render as renderHarness } from '@test/utils';
+
+import { Button } from '../Button';
+
+import { Tooltip, type TooltipProps } from './Tooltip';
+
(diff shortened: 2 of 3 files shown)


HichamELBSI:
I'm wondering what's the point of using the tooltip to not show the label tooltip ? 🤔

mazzucchelli:
> I'm wondering what's the point of using the tooltip to not show the label tooltip ? 🤔

To handle scenarios where the tooltip message is `string | undefined`. Consumers of the design system are forced to handle this conditional rendering on their side:

```
export const ButtonWithTooltip = ({
  tooltip,
  tooltipPosition,
  Component = Button,
  ...rest
}) => {
  const button = <Component {...rest} />;

  return tooltip ? (
    <Tooltip label={tooltip} side={tooltipPosition}>
        {button}
    </Tooltip>
  ) : (
    button
  );
};
```

The same applies to menuitems, links, etc..

If `label` is optional then we should also handle the case where it can be `undefined`

HichamELBSI (approved):

mazzucchelli:
> Should this PR be opened @mazzucchelli?

Yes, my bad, I didn't notice it was on draft state

lucasboilly:
Hey @mazzucchelli, I don't know if there's a way to check this from a design point of view but it seems like a good improvement :)

PR #1844, opened 2025-01-29
Register lodash as dependency to externalize it

jorrit:
### What does it do?

I noticed that the dist files of the design system package contain the entire lodash source.

See: https://www.npmjs.com/package/@strapi/design-system/v/2.0.0-rc.14?activeTab=code
image

The design-system package uses lodash directly in extendTheme.ts but doesn't list this as a dependency in package.json. lodash happens to be present because it is a sub-sub-subdependency:

```
> npm ls lodash
design-system@1.0.0 .\strapi-design-system
`-- @strapi/design-system@2.0.0-rc.14 -> .\packages\design-system
  `-- @strapi/pack-up@5.0.0
    `-- yup@0.32.9
      `-- lodash@4.17.21
```

By making it a direct dependency, vite will not bundle it in the dist files. Besides that, it is more correct to not rely on sub-dependencies.

### Why is it needed?

When investigating the dist code of the strapi-plugin-navigation module I noticed that it contains the entire lodash code. I investigated this and it lead to the conclusion that it is actually the design-system package that provides this.
See https://github.com/VirtusLab-Open-Source/strapi-plugin-navigation/issues/509

### How to test it?

Build the design-system package before and after this change and observe the file size decrease.

_It may be the case that there are more dependencies of the code that could be externalized_


HichamELBSI:
@jorrit Thanks for your contribution, in order to merge the PR, you will need to generate changeset. You can run `yarn release:add` to generate it. Feel free to follow the contributing guide.

jorrit:
My apologies for not following the guidelines. I'll update the PR after the weekend.

jorrit:
The commit has been updated. I found a couple of additional improvements using the `depcheck` tool.

HichamELBSI (approved):
Thank you!

PR #1850, opened 2025-02-05
feat(Menu): add start and end icons to MenuItem

markkaylor:
### What does it do?

Add start and end icons to Menu.Item

### Why is it needed?

The CMS uses icons in Menu.Item's often

### How to test it?

There are unit tests
Otherwise it can be tested locally in the SimpleMenu.stories.tsx file

### DIFF
--- a/packages/design-system/src/components/SimpleMenu/Menu.tsx
+++ b/packages/design-system/src/components/SimpleMenu/Menu.tsx
@@ -178,7 +178,13 @@ const MenuItem = ({ onSelect, disabled = false, isLink, startIcon, endIcon, isEx
   return (
     <DropdownMenu.Item asChild onSelect={onSelect} disabled={disabled}>
       {isLink || isExternal ? (
-        <OptionLink color="neutral800" {...props} isExternal={isExternal ?? false}>
+        <OptionLink
+          color="neutral800"
+          startIcon={startIcon}
+          endIcon={endIcon}
+          {...props}
+          isExternal={isExternal ?? false}
+        >
           <Typography>{props.children}</Typography>
         </OptionLink>
       ) : (
--- a/packages/design-system/src/components/SimpleMenu/Menu.tsx
+++ b/packages/design-system/src/components/SimpleMenu/Menu.tsx
@@ -174,11 +174,11 @@ type ItemProps<TComponent extends React.ComponentType = typeof BaseLink> =
   | ItemInternalLinkProps<TComponent>
   | ItemExternalLinkProps;
 
-const MenuItem = ({ onSelect, disabled = false, isLink, ...props }: ItemProps) => {
+const MenuItem = ({ onSelect, disabled = false, isLink, startIcon, endIcon, isExternal, ...props }: ItemProps) => {
   return (
     <DropdownMenu.Item asChild onSelect={onSelect} disabled={disabled}>
-      {isLink || props.isExternal ? (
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/SimpleMenu/Menu.tsx
```diff
@@ -180,8 +182,19 @@ const MenuItem = ({ onSelect, disabled = false, isLink, ...props }: ItemProps) =
...
-          <Typography>{props.children}</Typography>
+        <OptionButton
+          cursor="pointer"
+          color="neutral800"
+          background="transparent"
+          borderStyle="none"
+          gap={2}
+          {...props}
+        >
+          {props?.startIcon ? props.startIcon : null}
```
remidej, line 193: i don't think you want the ? (same for endIcon)

remidej (approved):

packages/design-system/src/components/SimpleMenu/Menu.tsx
```diff
@@ -180,8 +182,19 @@ const MenuItem = ({ onSelect, disabled = false, isLink, ...props }: ItemProps) =
...
       ) : (
-        <OptionButton cursor="pointer" color="neutral800" background="transparent" borderStyle="none" {...props}>
-          <Typography>{props.children}</Typography>
+        <OptionButton
+          cursor="pointer"
+          color="neutral800"
+          background="transparent"
+          borderStyle="none"
+          gap={2}
+          {...props}
```
HichamELBSI, line 191: Can you remove startIcon and endIcon from the props object to avoid passing them to the `OptionButton` component please, in the component function definition above => `({ onSelect, disabled = false, isLink, startIcon, endIcon, ...props }: ItemProps)`.

packages/design-system/src/components/SimpleMenu/Menu.tsx
```diff
@@ -180,8 +182,19 @@ const MenuItem = ({ onSelect, disabled = false, isLink, ...props }: ItemProps) =
...
+        >
+          {props.startIcon}
```
HichamELBSI, line 193: By doing this, you are authorising everything to be rendered. I won't ask you to add strong validation on the props but since we are assuming that user should only render non-interactive elements (in this use case, icons). I'd like to see a forced aria-hidden here. Same for the endIcon.

```suggestion
          {startIcon && <Box aria-hidden>{startIcon}</Box>}
```

HichamELBSI (changes requested):

packages/design-system/src/components/SimpleMenu/Menu.tsx
```diff
@@ -172,16 +174,41 @@ type ItemProps<TComponent extends React.ComponentType = typeof BaseLink> =
   | ItemInternalLinkProps<TComponent>
   | ItemExternalLinkProps;
 
-const MenuItem = ({ onSelect, disabled = false, isLink, ...props }: ItemProps) => {
+const MenuItem = ({ onSelect, disabled = false, isLink, startIcon, endIcon, isExternal, ...props }: ItemProps) => {
   return (
     <DropdownMenu.Item asChild onSelect={onSelect} disabled={disabled}>
-      {isLink || props.isExternal ? (
-        <OptionLink color="neutral800" {...props} isExternal={props.isExternal ?? false}>
```
markkaylor, line 179: @HichamELBSI  Following the same thinking as your last comment I thought it would make sense to also destructure isExternal from props as well. Let me know though if I overlooked something.

Issue #1849, opened 2025-02-05
[feat]: Use the new XS IconButton to clear the Searchbar


lucasboilly:
### A clear and concise description of what the feature is

Replace the current clear cross icon with the new XS variant of the `IconButton`.

### Why should this feature be included?

- It increases the click area and makes it more accessible.
- It adds hover and clicked states directly inherited from `IconButton`.

### Please provide an example for how this would work

It's supposed to look like this. Please make sure that the IconButton is vertically centered and 4px away from the right edge of the `Searchbar`.

https://github.com/user-attachments/assets/fc8e76f7-da9f-4b7e-88c3-575f6baa0fa3

Issue #1837, opened 2025-01-03
[feat]: Use the new XS IconButton to clear the combobox


lucasboilly:
### A clear and concise description of what the feature is

Replace the current clear cross icon with the new XS variant of the `IconButton`.

### Why should this feature be included?

- It increases the click area and makes it more accessible.
- It adds hover and clicked states directly inherited from `IconButton`.

### Please provide an example for how this would work

It's supposed to look like this. Please make sure that the IconButton is vertically centered because that's currently not the case with the clear cross icon.

https://github.com/user-attachments/assets/07e8de9f-f175-4312-a660-f3617536688f

PR #1878, opened 2025-03-10
feat: use xs iconbutton for combobox, datepicker and searchbar

lucasboilly:
#### Combobox
- Capitalized the first letter of the value of `clearLabel` 
- Replaced `IconBox` by `IconButton`

#### DatePicker
- Replaced `IconBox` by `IconButton`

#### Searchbar
- Wrapped `Cross` icon inside of `IconButton`

#### Field
- Changed the `right` property of `EndAction`

### Why is it needed?

See https://github.com/strapi/design-system/issues/1837, https://github.com/strapi/design-system/issues/1876 and https://github.com/strapi/design-system/issues/1849.

### How to test it?

You'll need to chec the preview.

### Related issue(s)/PR(s)

https://github.com/strapi/design-system/issues/1837
https://github.com/strapi/design-system/issues/1876
https://github.com/strapi/design-system/issues/1849

### DIFF
--- a/packages/design-system/src/components/Field/Field.tsx
+++ b/packages/design-system/src/components/Field/Field.tsx
@@ -264,7 +264,7 @@ const InputElement = styled.input<{
 
 const EndAction = styled(Flex)`
   position: absolute;
-  right: ${({ theme }) => theme.spaces[1]};
+  right: ${({ theme }) => theme.spaces[4]};
   top: 50%;
   transform: translateY(-50%);
 `;
--- a/packages/design-system/src/components/Field/Field.tsx
+++ b/packages/design-system/src/components/Field/Field.tsx
@@ -264,7 +264,7 @@ const InputElement = styled.input<{
 
 const EndAction = styled(Flex)`
   position: absolute;
-  right: ${({ theme }) => theme.spaces[4]};
+  right: ${({ theme }) => theme.spaces[1]};
   top: 50%;
   transform: translateY(-50%);
 `;
--- a/.changeset/famous-elephants-confess.md
+++ b/.changeset/famous-elephants-confess.md
@@ -1,0 +1,5 @@
+---
+"@strapi/design-system": patch
+---
+
+feat: use xs iconbutton for combobox, datepicker and searchbar
--- a/.changeset/famous-elephants-confess.md
+++ b/.changeset/famous-elephants-confess.md
@@ -1,5 +1,5 @@
 ---
-"@strapi/design-system": patch
+'@strapi/design-system': patch
 ---
 
 feat: use xs iconbutton for combobox, datepicker and searchbar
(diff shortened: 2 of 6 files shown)


packages/design-system/src/components/Field/Field.tsx
```diff
@@ -264,7 +264,7 @@ const InputElement = styled.input<{
 
 const EndAction = styled(Flex)`
   position: absolute;
-  right: ${({ theme }) => theme.spaces[4]};
```
HichamELBSI, line 267: This will have a strong impact on the design (we will basically need to update all the field with an endAction in CMS and Cloud and DS users will be impacted as well in their plugins). The `endAction` is not always used to display an IconButton which means that some input that have a different endAction like a loader or a label (e.g. The UID field in the CMS) will need to be fixed as well 🤷🏽.

Screenshot 2025-03-11 at 12 09 36

Screenshot 2025-03-11 at 12 09 46

HichamELBSI: Is the team aware of this?

HichamELBSI (changes requested):

packages/design-system/src/components/Field/Field.tsx
lucasboilly: @HichamELBSI won't it just move the `endAction` a few pixels closer from the edge of the input? If the other components (UID, loader) are displayed inside of `endAction` won't they just move as well?

What I wanted to fix here is that the cross was too far from the right edge of the input, like in the `Searchbar` for example:

CleanShot 2025-03-11 at 13 01 47@2x

lucasboilly: > Is the team aware of this?

It was a personal initiative so I'm OK to postpone this for later if it's too complex, the main purpose of this PR was to use the XS `IconButton` 👍

HichamELBSI: > @HichamELBSI won't it just move the `endAction` a few pixels closer from the edge of the input? If the other components (UID, loader) are displayed inside of `endAction` won't they just move as well?

Yes, this is the issue, the endAction will be really close to the right border. It will create padding size mismatches with the start icon. That's why we will need an update everywhere we use the endAction.

Screenshot 2025-03-11 at 15 15 20

lucasboilly: @HichamELBSI okay, maybe we could compensate by adding a right padding in `endAction`  but we'll see this later once we'll know all the impacts 👍

I just pushed pushed a new update that restores the previous code for this line 🙂

HichamELBSI (dismissed):

lucasboilly:
@HichamELBSI can we merge it or do I need to change something else?

HichamELBSI:
Definitely yes. I'm updating the PR and I'll merge it.

lucasboilly:
Perfect, thanks!

HichamELBSI (approved):

Issue #1815, opened 2024-11-06
[bug]: Visual bug on the modal's animation


lucasboilly:
### What version of design-system are you using?

Browser: Arc

### What's Wrong?

The modal's animation isn't fluid as it should be. I recorded the Dialog animation as well for comparison.

https://github.com/user-attachments/assets/52ff5e24-2a3e-469d-ad03-e89aaca7ff1f

### To Reproduce

Open any modal (ex: in the Content-Type Builder or in the Media Library).

### Expected Behaviour

The animation should be as fluid as the one of the Dialog.

ShatilKhan:
looking into it

PR #1896, opened 2025-04-01
fix: modal component open and close animations

HichamELBSI:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

https://github.com/strapi/design-system/issues/1815

### DIFF
--- a/packages/design-system/src/components/Modal/Modal.tsx
+++ b/packages/design-system/src/components/Modal/Modal.tsx
@@ -4,7 +4,6 @@ import * as Dialog from '@radix-ui/react-dialog';
 import { Cross } from '@strapi/icons';
 import { styled } from 'styled-components';
 
-import { setOpacity } from '../../helpers/setOpacity';
 import { Flex, type FlexComponent, type FlexProps } from '../../primitives/Flex';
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/Modal/Modal.tsx
```diff
@@ -42,18 +41,18 @@ interface ContentProps extends Dialog.DialogContentProps {}
...
 const Overlay = styled(Dialog.Overlay)`
-  background: ${(props) => setOpacity(props.theme.colors.neutral800, 0.2)};
+  background-color: ${(props) => props.theme.colors.neutral800};
   position: fixed;
   inset: 0;
   z-index: ${(props) => props.theme.zIndices.overlay};
+  opacity: 0.2;
 
   @media (prefers-reduced-motion: no-preference) {
     animation: ${ANIMATIONS.overlayFadeIn} ${(props) => props.theme.motion.timings['200']}
```
remidej, line 58: Maybe we can use will-change here too?

HichamELBSI: You are right! Added

PR #1905, opened 2025-04-10
feat: creation item always visible and sticky

simotae14:
### What does it do?

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/docs/stories/03-inputs/Combobox.stories.tsx
+++ b/docs/stories/03-inputs/Combobox.stories.tsx
@@ -150,7 +150,6 @@ export const Creatable: Story = {
 
     return (
       <Combobox
-        aria-describedby="creatable combobox"
         placeholder="My favourite fruit is..."
         value={value}
         onChange={setValue}
@@ -171,7 +170,6 @@ export const Creatable: Story = {
       source: {
         code: outdent`
       <Combobox
-        aria-describedby="creatable combobox"
         placeholder="My favourite fruit is..."
         value={value}
         onChange={setValue}
@@ -194,11 +192,12 @@ export const CreatableVisible: Story = {
   render: () => {
     const [value, setValue] = React.useState<string | undefined>('');
 
-    const onCreateOption = () => {};
+    const onCreateOption = () => {
+      console.log('Created option');
+    };
 
     return (
       <Combobox
-        aria-describedby="creatable combobox"
         placeholder="My favourite fruit is..."
         value={value}
         onChange={setValue}
@@ -226,7 +225,6 @@ export const CreatableVisible: Story = {
       source: {
         code: outdent`
       <Combobox
-        aria-describedby="creatable combobox"
         placeholder="My favourite fruit is..."
         value={value}
         onChange={setValue}
--- a/docs/stories/03-inputs/Combobox.stories.tsx
+++ b/docs/stories/03-inputs/Combobox.stories.tsx
@@ -205,7 +205,7 @@ export const CreatableVisible: Story = {
         onCreateOption={onCreateOption}
         creatable="visible"
         creatableStartIcon={<Plus fill="neutral500" />}
-        createMessage={() => 'Create a relation'}
+        createMessage={() => 'Create a fruit'}
       >
         {options.map(({ name, value }) => (
           <ComboboxOption key={value} value={value}>
--- a/docs/stories/03-inputs/Combobox.stories.tsx
+++ b/docs/stories/03-inputs/Combobox.stories.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 import { Meta, StoryObj } from '@storybook/react';
 import { Combobox, ComboboxOption, Field, Flex, Typography } from '@strapi/design-system';
-import { Link as LinkIcon } from '@strapi/icons';
+import { Link as LinkIcon, Plus } from '@strapi/icons';
 import { default as outdent } from 'outdent';
 
 const options = [
@@ -200,9 +200,10 @@ export const CreatableAlwaysVisible: Story = {
         placeholder="My favourite fruit is..."
         value={value}
         onChange={setValue}
-        creatable
         createItemAlwaysVisible
+        creatableStartIcon={<Plus fill="neutral500" />}
         createMessage={() => 'Create a relation'}
+        onCreateOption={() => {}}
       >
         {options.map(({ name, value }) => (
           <ComboboxOption key={value} value={value}>
@@ -227,7 +228,6 @@ export const CreatableAlwaysVisible: Story = {
         placeholder="My favourite fruit is..."
         value={value}
         onChange={setValue}
-        creatable
         createItemAlwaysVisible
         createMessage={() => 'Create a relation'}
       >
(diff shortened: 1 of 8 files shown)


docs/stories/03-inputs/Combobox.stories.tsx
```diff
@@ -189,6 +190,64 @@ export const Creatable: Story = {
...
+    const [value, setValue] = React.useState<string | undefined>('');
+
+    return (
+      <Combobox
+        aria-describedby="creatable combobox"
+        placeholder="My favourite fruit is..."
+        value={value}
+        onChange={setValue}
+        creatable
+        createItemAlwaysVisible
```
HichamELBSI, line 204: I would like to avoid adding a new property for that. This will turn into a props hell in the future. 

Can't we do something like this ⬇️  
```suggestion
        creatable="sticky"
```

So "creatable" could be typed as `creatable: boolean | 'sticky'` to avoid breaking changes, handle our new sticky use case and still be flexible in the future.

packages/design-system/src/components/Combobox/Combobox.tsx
```diff
@@ -290,6 +281,25 @@ const Combobox = React.forwardRef<ComboboxInputElement, ComboboxProps>(
...
+                  <OptionBox>
+                    <Flex gap={2} justifyContent="space-between">
+                      <Flex gap={2}>
+                        <Plus fill="neutral500" />
+                        <Typography>{createMessage(internalTextValue ?? '')}</Typography>
+                      </Flex>
+                    </Flex>
+                  </OptionBox>
+                </ComboboxPrimitive.CreateItem>
+              </CreateItemContainer>
```
HichamELBSI, line 301: We could have something a bit more flexible here by giving the choice for the startIcon for example:

```suggestion
              <CreateItemContainer>
                <ComboboxPrimitive.CreateItem
                  onPointerUp={handleCreateItemClick}
                  onClick={handleCreateItemClick}
                  createItemAlwaysVisible={createItemAlwaysVisible}
                  asChild
                >
                  <OptionBox>
                    <Flex gap={2} justifyContent="space-between">
                      <Flex gap={2}>
                        {creatableStartIcon && (
                          <Box tag="span" aria-hidden>
                            {creatableStartIcon}
                           </Box>
                         )}
                        <Typography>{createMessage(internalTextValue ?? '')}</Typography>
                      </Flex>
                    </Flex>
                  </OptionBox>
                </ComboboxPrimitive.CreateItem>
              </CreateItemContainer>
```

HichamELBSI (changes requested):

docs/stories/03-inputs/Combobox.stories.tsx
simotae14: I decided to use "visible" because at the end we want to show the create button always visible

docs/stories/03-inputs/Combobox.stories.tsx
```diff
@@ -456,10 +514,11 @@ export const ComboboxProps = {
...
     },
     creatable: {
-      control: 'boolean',
-      description: 'If true, allows creating new options',
+      control: 'radio',
+      options: [false, true, 'visible'],
+      description: 'If true, allows creating new options. If "visible", always shows the create option.',
       table: {
-        type: { summary: 'boolean' },
+        type: { summary: 'boolean | "visible"' },
```
HichamELBSI, line 521: I'm afraid that “visible” is understood as the basic visible state. This could lead to the belief that if it's not set to “visible”, then it's “invisible”. Don't you think “sticky” would be more appropriate?

docs/stories/03-inputs/Combobox.stories.tsx
```diff
@@ -189,6 +190,63 @@ export const Creatable: Story = {
...
+        code: outdent`
+      <Combobox
+        aria-describedby="creatable combobox"
+        placeholder="My favourite fruit is..."
+        value={value}
+        onChange={setValue}
+        onCreateOption={onCreateOption}
+        creatable="visible"
+        creatableStartIcon={<Plus fill="neutral500" />}
+        createMessage={() => 'Create a relation'}
```
HichamELBSI, line 236: Please don't mention any CMS features in the DS.

HichamELBSI: especially when the combobox is about Fruits actually (placeholder, options, etc) 😅

packages/design-system/src/components/Combobox/Combobox.tsx
```diff
@@ -400,6 +414,12 @@ const Content = styled(ComboboxPrimitive.Content)`
   }
 `;
 
+const CreateItemContainer = styled(Box)`
```
HichamELBSI, line 417: The sticky item can't be reached with the keyboard. We need to fix this for the sake of accessibility.

packages/design-system/src/components/Combobox/Combobox.test.tsx
```diff
@@ -251,6 +251,24 @@ describe('Combobox', () => {
...
+      expect(getByRole('option', { name: /Create/ })).toBeInTheDocument();
+
+      await user.type(getByRole('combobox'), 'Hamb');
+
+      expect(getByRole('option', { name: 'Create "Hamb"' })).toBeInTheDocument();
+
+      await user.type(getByRole('combobox'), 'invalid value');
+
+      expect(getByText('No results found')).toBeInTheDocument();
+    });
```
HichamELBSI, line 270: Please add a text to handle the click on the sticky creatable element. And another one with the keyboard.

packages/primitives/src/components/Combobox/Combobox.tsx
```diff
@@ -1278,7 +1294,7 @@ const ComboboxCreateItem = React.forwardRef<ComboboxItemElement, CreateItemProps
     };
   }, [textValue, subscribe, getItems]);
 
-  if (!textValue || !show) {
+  if ((!textValue || !show) && !context.visible) {
```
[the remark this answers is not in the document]
simotae14: the old use case used to add the value typed in the list of options even if the callback onCreateOption is not passed, it is because of that "no results" is never shown

packages/design-system/src/components/Combobox/Combobox.tsx
simotae14: Are you sure it is not accessible? I can click on the button using the Up arrow

https://github.com/user-attachments/assets/df9cf0e7-b850-46b5-8440-243fb8af526a

HichamELBSI: Yes but this isn't how it should work unfortunately. How can keyboard users understand that they have to press Up to focus the footer 😬. It is not accessible by pressing Down after the last option same with tab, we cannot reach the footer.

HichamELBSI:
Is it me or the comboboxes are always open now. When I load the page, it is open by default and even after selecting a value, it is still open 🤔 

Screenshot 2025-04-14 at 16 19 23

simotae14:
https://github.com/strapi/design-system/pull/1905#issuecomment-2801880814 I have forgot a condition true I used to fix some ui problems, now it is fixed

packages/design-system/src/components/Combobox/Combobox.tsx
simotae14: Please check now

HichamELBSI (approved):
Well done @simotae14

markkaylor (approved):

PR #1907, opened 2025-04-15
fix: Combobox creatable visible button needs to close the combobox when clicked

simotae14:
### What does it do?

It fixes some small errors we introduced with the refactoring of the Combobox

### Why is it needed?

it doesn't close the combobox when the creatable button is clicked with creatable="visible"

### How to test it?

Try to click on the creatable button in the Creatable Visible scenario in the storybook (with mouse and keyboard) and check if the combobox is then closed

### Related issue(s)/PR(s)

### DIFF
--- a/packages/primitives/src/components/Combobox/Combobox.tsx
+++ b/packages/primitives/src/components/Combobox/Combobox.tsx
@@ -1132,7 +1132,7 @@ const ComboboxItemImpl = React.forwardRef<ComboboxItemImplElement, ItemImplProps
       {...restProps}
       id={id}
       ref={composedRefs}
-      onClick={composeEventHandlers(restProps.onClick, handleSelect)}
+      onPointerUp={composeEventHandlers(restProps.onPointerUp, handleSelect)}
     />
   );
 });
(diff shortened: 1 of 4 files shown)


packages/primitives/src/components/Combobox/Combobox.tsx
```diff
@@ -1132,7 +1132,7 @@ const ComboboxItemImpl = React.forwardRef<ComboboxItemImplElement, ItemImplProps
       {...restProps}
       id={id}
       ref={composedRefs}
-      onPointerUp={composeEventHandlers(restProps.onPointerUp, handleSelect)}
+      onClick={composeEventHandlers(restProps.onClick, handleSelect)}
```
markkaylor, line 1135: Why change from onPointerUp to onClick?

simotae14: because of a strange behaviour with keyboard navigation, when I clicked the button it triggered the handler three times, one for pointer up, one for pointer down and one for click

simotae14: btw I reverted back the onPointerUp

remidej (approved):
All good to me

PR #1910, opened 2025-04-30
fix: Combobox fix background issue in the creatable button on hover

simotae14:
### What does it do?

Fix an issue when you hover the creatable button with different options in the Combobox

### Why is it needed?

To solve this issue
Screenshot 2025-04-30 alle 10 04 36

### How to test it?

Check the Combobox creatable option in the storybook and scroll the different options and hover the creatable button to check if the issue is gone

### Related issue(s)/PR(s)

CS-1409

### DIFF
--- a/packages/design-system/src/components/Combobox/Combobox.tsx
+++ b/packages/design-system/src/components/Combobox/Combobox.tsx
@@ -426,7 +426,7 @@ const ComboboxCreateItem = styled(ComboboxPrimitive.CreateItem)`
   }
   &&:hover,
   &&[data-highlighted] {
-    background-color: transparent;
+    background: ${({ theme }) => theme.colors.neutral0};
   }
   && > div {
     padding: ${({ theme }) => theme.spaces[2]} ${({ theme }) => theme.spaces[4]};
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/Combobox/Combobox.tsx
```diff
@@ -426,7 +426,7 @@ const ComboboxCreateItem = styled(ComboboxPrimitive.CreateItem)`
   }
   &&:hover,
   &&[data-highlighted] {
-    background-color: transparent;
+    background: ${({ theme }) => theme.colors.neutral0};
   }
```
remidej, line 430: I wonder if you can just remove this block? There's a neutral0 bg defined above already that maybe would be enough. Otherwise this is fine of course

simotae14: If I remove this line I have this result
Screenshot 2025-04-30 alle 11 49 52
with my solution, instead, we have this background
Screenshot 2025-04-30 alle 11 50 03

remidej (approved):

PR #1911, opened 2025-04-30
feat: add disabled state to the create item option

simotae14:
### What does it do?

Add a new disabled state, and style, to the Combobox Create item option

### Why is it needed?

We want to show the combobox options and disable the combobox create item button in case the user has no permission to create an item.

### How to test it?

In the storybook you can find a new Combobox Creatable disabled section to check the solution provided.

### Related issue(s)/PR(s)

CS-1405

### DIFF
--- a/docs/stories/03-inputs/Combobox.stories.tsx
+++ b/docs/stories/03-inputs/Combobox.stories.tsx
@@ -245,6 +245,65 @@ export const CreatableVisible: Story = {
   },
 };
 
+export const CreatableDisabled: Story = {
+  render: () => {
+    const [value, setValue] = React.useState<string | undefined>('');
+
+    const onCreateOption = () => {
+      console.log('Created option');
+    };
+
+    return (
+      <Combobox
+        placeholder="My favourite fruit is..."
+        value={value}
+        onChange={setValue}
+        onCreateOption={onCreateOption}
+        creatable="visible"
+        creatableDisabled={true}
+        creatableStartIcon={<Plus fill="neutral500" />}
+        createMessage={() => 'Create a fruit'}
+      >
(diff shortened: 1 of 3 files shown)


docs/stories/03-inputs/Combobox.stories.tsx
remidej, line 1: I think you should include this story in the storybook docs page too

packages/design-system/src/components/Combobox/Combobox.tsx
```diff
@@ -47,6 +47,7 @@ interface ComboboxProps
     Omit<ComboboxPrimitive.TextInputProps, 'required' | 'disabled' | 'value' | 'onChange' | 'size'> {
   clearLabel?: string;
   creatable?: boolean | 'visible';
+  creatableDisabled?: boolean;
```
remidej, line 50: screenshot 2025-05-05 at 11 11 25@2x

fyi the props table doesn't receive the type properly

packages/design-system/src/components/Combobox/Combobox.tsx
```diff
@@ -428,13 +435,25 @@ const ComboboxCreateItem = styled(ComboboxPrimitive.CreateItem)`
   &&[data-highlighted] {
     background-color: transparent;
   }
+  &&[data-disabled] {
+    color: ${({ theme }) => theme.colors.neutral600};
+    background: ${({ theme }) => theme.colors.neutral150};
+    cursor: not-allowed;
+  }
```
remidej, line 442: there's a small visual issue with the border radii
screenshot 2025-05-05 at 11 12 13@2x

remidej (approved):

Issue #1902, opened 2025-04-04
[bug]: The content of the modals is cropped


lucasboilly:
### What version of design-system are you using?

- Browser: Arc

### What's Wrong?

The content of the modals is cropped.

![Image](https://github.com/user-attachments/assets/43394fde-5833-431d-9dec-dc3e5c8813ad)

### To Reproduce

Open a modal that has an input close to the edge and click on it. The blue border will be cropped.

### Expected Behaviour

It shouldn't crop anything.

PR #1913, opened 2025-05-05
Fix: input box shadows cropped

simotae14:
### What does it do?

Fix the issue we had on the left and right of focused input elements inside modals

### Why is it needed?

the left and right shadows of input elements where sometimes cropped

### How to test it?

Try to test the issue in the Storybook in the Modal component

### Related issue(s)/PR(s)

fixes https://github.com/strapi/design-system/issues/1902

### DIFF
--- a/packages/design-system/src/themes/utils.ts
+++ b/packages/design-system/src/themes/utils.ts
@@ -13,11 +13,6 @@ export const inputFocusStyle =
     transition-property: border-color, box-shadow, fill;
     transition-duration: 0.2s;
 
-    /* Add a margin left and right to reserve space for the box shadow */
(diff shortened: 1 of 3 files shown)


packages/design-system/src/themes/utils.ts
```diff
@@ -13,6 +13,11 @@ export const inputFocusStyle =
     transition-property: border-color, box-shadow, fill;
     transition-duration: 0.2s;
 
+    /* Add a margin left and right to reserve space for the box shadow */
+    ${rootElement} {
+      margin: 0 2px;
```
remidej, line 18: doesn't a margin affect the positioning of an element on the page?

PR #1920, opened 2025-06-03
fix: datetime fields responsive

Adzouz:
### What does it do?

Improve the responsive of the datetime picker fields on small containers.

https://github.com/user-attachments/assets/e4816e78-ceaf-484d-bb1d-473d817e01f3

_I also took the liberty to unify the size between the two elements (most of the time it's 50/50 in the admin interface)_

### Why is it needed?

Improve the experience on smaller screens or when including a datetime field inside a modal / small container.

### How to test it?

- Go to storybook > Inputs > DateTimePicker
- Try to resize the window

### Related issue(s)/PR(s)

Resolves [#1853](https://github.com/strapi/design-system/issues/1853)

🚀

Issue #1853, opened 2025-02-06
[bug]: No property available to stack date and time picker fields vertically [column instead of row]


TMSchipper:
### What version of design-system are you using?

"@strapi/strapi": "^5.2.0",
"@strapi/design-system": "2.0.0-rc.11",

### What's Wrong?

The date and time picker components in the Strapi Design System currently render in a single row, with no available property or option to display them stacked vertically. This causes layout issues, particularly in narrow side panels (such as the content manager sidebar), where the elements overflow beyond the container.

![Image](https://github.com/user-attachments/assets/60c4d814-4391-4599-8d52-9184effb6c78)

### To Reproduce

Use the <DateTimePicker /> component in a small container (e.g., a sidebar).
Notice that the date and time fields are displayed inline (horizontally), leading to overflow issues.
There is no documented property to switch to a vertical layout.

### Expected Behaviour

There should be a property (e.g., stacked or layout="vertical") to allow developers to control whether the date and time fields are displayed:

Horizontally (default behavior)
Vertically (one field below the other)
This would improve usability and allow for better responsive layouts.

HichamELBSI:
Thanks again for the issue @TMSchipper, we will handle it as soon as possible.

Adzouz:
Hi @TMSchipper,
I tried to reduce the width to 130px but the thing is that we have a clear button that takes space and will crop part of the content unfortunately 😢 See my screenshot attached:
![Image](https://github.com/user-attachments/assets/220d649a-a29e-4e3e-9b94-3e43916433e6)

Adzouz:
Hi @TMSchipper 👋
Good news, in the end I was able to check if the onClear is present or not and if not, it will shrink at 130px 😊
I also applied the changes directly on DatePicker and TimePicker fields so that it would also work by implementing them separately.
The new version of the design-system was released already and it will be part of the next release of Strapi on Wednesday!
Have a great day 🚀

TMSchipper:
Hi @Adzouz, lovely! 

Thank you for your work & explantion! 🫶🏼

PR #1922, opened 2025-06-05
Fix/datetime min-width

### LINKED ISSUE #1882: [bug]: The width of the DateTimePicker changes unexpectedly
### What version of design-system are you using?

@strapi/design-system@2.0.0-rc.18

### What's Wrong?

The input width changes depending on whether the clear button is present.

### To Reproduce

https://design-system.strapi.io/?path=/story/inputs-datetimepicker--base

### Expected Behaviour

The input width does not change due to any actions.

Adzouz:
### What does it do?

I previous PR: https://github.com/strapi/design-system/pull/1920 I updated the responsive of the DateTimePicker but I just had the idea to move the min-width logic inside DatePicker and TimePicker themselves instead to be sure that even if used separated, they'd still behave the same.
I also added a condition to decrease the min-width in the case that onClear is not provided (so that the clear icon won't appear and leave more space).

### Why is it needed?

Improve the visual of both DatePicker and TimePicker so they take less space.

### How to test it?

- Go to Inputs > DatePicker / DateTimePicker and TimePicker
- Resize the window to see them shrinking

### Related issue(s)/PR(s)

Resolves #1882 and #1853

🚀

### DIFF
--- a/.changeset/funny-bulldogs-kneel.md
+++ b/.changeset/funny-bulldogs-kneel.md
@@ -2,4 +2,4 @@
 '@strapi/design-system': patch
 ---
 
-Min-width of DatePicker and TimePicker at wrong place
+fix: min-width of DatePicker and TimePicker
--- a/.changeset/swift-cats-join.md
+++ b/.changeset/swift-cats-join.md
@@ -2,4 +2,4 @@
 '@strapi/design-system': patch
 ---
 
-Wrong color of placeholders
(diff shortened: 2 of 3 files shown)


simotae14 (approved):
The code looks good and it works, can I ask you to modify your changeset to use the structure I suggested

simotae14 (approved):
well done!

PR #1930, opened 2025-06-13
fix: wrapping content of dialog in overlay to restore pointer-events

Adzouz:
### What does it do?

Wraps the content of a dialog inside the overlay to restore the pointer-events and allow user to scroll in select that would be included in a dialog.

### Why is it needed?

The user can't scroll in a select inside a dialog.

### How to test it?

- Go to the admin interface > Content manager
- Pick a localized collection
- Create or edit an item
- Create multiple languages variations of the item (at least 5)
- Create again a new language variation
- Click on the globe icon on top right (Fill in from another locale)

### Related issue(s)/PR(s)

Resolves https://github.com/strapi/strapi/issues/23205

🚀

### DIFF
--- a/.changeset/tricky-plants-sip.md
+++ b/.changeset/tricky-plants-sip.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': patch
+---
+
+fix: wrapping content of dialog in overlay to restore pointer-events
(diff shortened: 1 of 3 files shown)


markkaylor (approved):
LGTM

Issue #1924, opened 2025-06-12
[bug]: Combobox's height changes depending on its content


lucasboilly:
### What version of design-system are you using?

- Browser: Arc

### What's Wrong?

Combobox's height changes depending on its content.

https://github.com/user-attachments/assets/0b2e3ef3-1c68-4aee-b33f-8dadefbd2ed6

### To Reproduce

See [Combobox](https://design-system.strapi.io/?path=/docs/inputs-combobox--docs) in Strapi's design system.

### Expected Behaviour

Its height should be persistent.

PR #1932, opened 2025-06-13
fix: wrong padding on IconButton causing glitch on input heights

### LINKED ISSUE #1925: [bug]: DatePicker's height changes depending on its content
### What version of design-system are you using?

- Browser: Arc

### What's Wrong?

Combobox's height changes depending on its content.

https://github.com/user-attachments/assets/80bc6fe1-7732-4e65-9057-5f868cbd729a

### To Reproduce

See [DatePicker](https://design-system.strapi.io/?path=/docs/inputs-datepicker--docs) in Strapi's design system.

### Expected Behaviour

Its height should be persistent.

Adzouz:
### What does it do?

Updated the padding on IconButton that was causing a glitch of height on inputs / fields that use them.

Rules:
The line-height of the content is 22px.
The actual padding of the IconButton is 5px, the icon 16px = 23px.
It should be 22px if we don't want the field to glitch.

### Why is it needed?

It's shifting the content in the CMS and it's not good.

### How to test it?

- Go to Storybook
- You can try any of the input with an IconButton (Combobox, DatePicker, DateTimePicker, Select, TextInput or TimePicker)
- Write / select a value to make the clear icon appearing and glitching the height.

### Related issue(s)/PR(s)

Resolves #1925 and #1924 by @lucasboilly

🚀

### DIFF
--- a/.changeset/brave-years-yell.md
+++ b/.changeset/brave-years-yell.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': patch
+---
+
+fix: glitch of height on inputs caused by wrong padding on IconButton
--- a/docs/stories/03-inputs/Combobox.stories.tsx
+++ b/docs/stories/03-inputs/Combobox.stories.tsx
@@ -317,6 +317,15 @@ export const Size: Story = {
   args: {
     size: 'S',
   },
+  render: (props) => (
+    <Combobox {...props}>
+      {options.map(({ name, value }) => (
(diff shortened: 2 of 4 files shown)


simotae14 (approved):
The code looks good and I tested your solution on the generated storybook https://64f9707bb212b1f1332bfd97-bbicesdiai.chromatic.com/?path=/story/inputs-datetimepicker--base&args=value:!date(2025-06-15T22:00:02.520Z) and it works, well done

PR #1933, opened 2025-06-14
Allow vertically resizing Textarea

### LINKED ISSUE #1916: [feat]: Textarea should be able to resized vertical

In the system the textarea has a hard coded not configurable height of 10.5rem.
For editors is that very uncomfortable for longtext inputs with more than 3 lines of content.

The feature allows adjusting the height of textarea during editing additional to scrolling with a simple change of two properties.

For editors is a fixed height of 10.5rem that very uncomfortable for longtext inputs with more than 3 lines of content. The authors should be able to adjust the size during editing instead of scrolling in a 3 lines height area.

The proposed change is:

1. replace the height property in the textarea wrapper with a minHeight of the same value
2. give the textarea the resize attribute with the value vertical

spruce:
### What does it do?

Changed the height attribute to min-height and allowed the textarea to be resized.

### Why is it needed?

As described in https://github.com/strapi/design-system/issues/1916 it would be nice to resize Textareas for better editing. In a later PR it would be nice to add the ability to specify another min-height's for each textarea input

### How to test it?

Create a new multiline text input in any entity and then go to the content view of that entity and see that the textarea is resizeable

### Related issue(s)/PR(s)

Fixes #1916

### DIFF
--- a/packages/design-system/src/components/Textarea/Textarea.tsx
+++ b/packages/design-system/src/components/Textarea/Textarea.tsx
@@ -54,13 +54,13 @@ const Textarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
 );
 
 const Wrapper = styled<BoxComponent>(Box)<{ $hasError?: boolean }>`
-  height: 10.5rem;
+  min-height: 10.5rem;
   ${inputFocusStyle()}
 `;
 
 const TextareaElement = styled<BoxComponent<'textarea'>>(Box)`
   border: none;
-  resize: none;
+  resize: vertical;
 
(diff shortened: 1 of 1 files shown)


lucasboilly:
@Adzouz could we make it possible to disable resizing?

Adzouz:
> @Adzouz could we make it possible to disable resizing?

Done ✅

spruce:
Nice. Works as I need it. Thank you very much. 👍

Adzouz:
Can one of you @jhoward1994 or @butcherZ can take a look at this one please? 🙏

PR #1939, opened 2025-06-18
feat(Popover): export Popover.Anchor

markkaylor:
### What does it do?

Exports Anchor along with the other Popover parts from radix

### Why is it needed?

We need it for a feature in the CMS

### How to test it?

No need really, but you can use the anchor just like the Trigger but control the open state yourself

```
<Popover.Root open={true}>
    <Popover.Anchor><div>I'm an anchor</div></Popover.Anchor>
    <Popover.Content side="top" align="center">
      	Hey I'm content!
    </Popover.Content>
  </Popover.Root>
```

### DIFF
--- a/.changeset/dirty-readers-listen.md
+++ b/.changeset/dirty-readers-listen.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': minor
+---
+
+export Popover.Anchor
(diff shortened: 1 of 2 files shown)


simotae14 (approved):
👍🏽

PR #1949, opened 2025-08-04
fix: dark theme support in stories

Adzouz:
### What does it do?

Fix the dark mode support in the stories and the Color page that was broken since last update of the Design System.

Screenshot 2025-08-04 at 17 50 55

Screenshot 2025-08-04 at 17 50 35

### Why is it needed?

After bumping to the latest version of dependencies (storybook v9 and using @vueless/storybook-dark-mode), the dark mode switch in Storybook wasn't working anymore.
In addition, it was breaking the Foundation > Color page because it was using a hook that is not allowed in the stories components.

### How to test it?

- Go to Storybook
- Click on the little Sun / Moon icon in the top bar to switch between light and dark mode
- Go to Foundation > Color
  -> The page should display the list of colors correctly without any error

🚀


simotae14 (approved):
I tested your solution and it works, as you can see in this video

https://github.com/user-attachments/assets/0f8fd07a-5df3-4b49-b13d-418a74183eae

@butcherZ please take a look at this bugfix to have more context on the issue we had with the last huge PR merged in the design system

Thanks

simotae14 (dismissed):
looks even better now

PR #1951, opened 2025-08-06
Revert "chore: upgrade to storybook v9 for using design addon (#1947)"

Adzouz:
This reverts commit f2729b00290c8c1c805e5d00e1a7a4a340be70d8.

### What does it do?

Revert the migration of Storybook v9 causing a few issues.

### How to test it?

The Color page and the theme switching should be fixed with the revert.


Adzouz:
We don't need to revert since we succeed making it work on Storybook v9 (thanks to @simotae14 too 🙏)

PR #1958, opened 2025-09-25
Grid item column fallback

HichamELBSI:
### What does it do?

This PR changes the default `Grid.Item` column size fallback. Right now, the code always falls back to `12` (a full-width span) unless a breakpoint-specific value (`$s`, `$m`, `$col`) is provided. For example:

At `xs`, if you don’t pass `$xs`, it defaults to 12.
At `s`, if you don’t pass `$s`, it defaults to `$xs` (or `12` if `$xs` wasn’t set).
At `m`, if you don’t pass `$m`, it defaults to `$s` → `$xs` → `12`.
At large, it cascades: `$col` → `$m` → `$s` → `$xs` → `12`.

This means that if you set `col={6}` but don’t define `xs`, `s`, or `m`, the smaller breakpoints will expand the item to `12` columns instead of inheriting `6`.

### Why is it needed?

### How to test it?

### Related issue(s)/PR(s)

### DIFF
--- a/docs/stories/Grid.stories.tsx
+++ b/docs/stories/Grid.stories.tsx
@@ -47,6 +47,21 @@ export const BaseGrid = {
   name: 'base grid',
 } satisfies Story;
 
+export const GridItemColSizeFallback = {
+  render: () => (
+    <Grid.Root gap={5}>
+      {Array(12)
+        .fill(null)
+        .map((_, i) => (
+          <Grid.Item key={i} background="success200" col={6}>
+            <Typography>Column {i + 1}</Typography>
+          </Grid.Item>
+        ))}
+    </Grid.Root>
+  ),
+  name: 'Column size fallback to col',
+} satisfies Story;
+
 export const ComplexGrid = {
   render: () => (
     <Grid.Root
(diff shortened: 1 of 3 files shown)


Adzouz:
LGTM, good catch 🚀
There's just the changeset missing, if you can add it please 🙏
Also, would we want to create a specific story in Storybook to cover the use case where there's only the parameter col defined without any other breakpoint?
Also (number 2), when this will be released, we have to make sure to fix its usage on the CMS otherwise we'll have some regression on mobile and tablet breakpoints on a few elements.

Adzouz (dismissed):
LGTM, good catch 🚀
There's just the changeset missing, if you can add it please 🙏
Also, would we want to create a specific story in Storybook to cover the use case where there's only the parameter col defined without any other breakpoint?
Also (number 2), when this will be released, we have to make sure to fix its usage on the CMS otherwise we'll have some regression on mobile and tablet breakpoints on a few elements.

docs/stories/Grid.stories.tsx
```diff
@@ -47,6 +47,21 @@ export const BaseGrid = {
   name: 'base grid',
 } satisfies Story;
 
+export const GridItemColSuzeFallback = {
```
Adzouz, line 50: Little typo here 😬

PR #2020, opened 2026-06-10
chore: check changeset presence + block major

Adzouz:
### What does it do?

- Adds a check for changeset presence in every PR
- Block merging the PR if changeset flagged as major (overridable by a dedicated `flag: allow-major` label)

### Why is it needed?

It's happening way too much that we forget to add changesets for every PR. This PR is so that we won't forget.
We also almost released a major version (fortunately that failed) so adding just an additional security to be sure this won't happen again.

### How to test it?

- Create a new PR with some changes but no changeset
→ The merge should be blocked and the pipeline should fail at step `changeset-check`

- Create a new PR with some changes and a changeset with a major version bump
→ The merge should be blocked and the pipeline should fail at step `check-major-changeset`

🚀

### DIFF
--- a/.github/scripts/check-changeset.cjs
+++ b/.github/scripts/check-changeset.cjs
@@ -1,0 +1,76 @@
+'use strict';
+
+const { execSync } = require('node:child_process');
+const fs = require('node:fs');
+
+// folder name under packages/ -> published npm package name
+const PUBLISHED_PACKAGES = {
+  'design-system': '@strapi/design-system',
+  icons: '@strapi/icons',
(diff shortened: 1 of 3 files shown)


Adzouz:
Hey @mathildeleg, I put some dummy files to test the PR. I'll remove them as soon as you're ok with the provided solution 😊

mathildeleg (dismissed):
Looks good (as long as you don't forget to remove your tests haha)

mathildeleg (approved):

PR #2024, opened 2026-06-11
chore: drop React 17 from supported peer dependency range

unrevised6419:
## What

Removes all traces of React 17 from the repo, mirroring [strapi/strapi#26194](https://github.com/strapi/strapi/pull/26194).

- Tighten `react` / `react-dom` **peerDependencies** to `^18.0.0` (was `^17.0.0 || ^18.0.0`) in:
  - `@strapi/design-system`
  - `@strapi/icons`
  - `@strapi/ui-primitives`
- Remove the React 17 client-only fallback (`|| (() => undefined)`) and its stale comment from the `useId` hook. Behavior under React 18 is identical — `React.useId` is always defined.
- Add changeset (patch bump for all three packages).

## Why

The packages already build and test exclusively against `react@18.3.1` (devDependencies) and rely on React 18-only APIs. The `^17` peer range was stale metadata. This is a peer-range correction, not a new behavioral break.

In practice React 17 was already impossible to use: the Strapi Admin entry point itself mounts via the React 18-only `createRoot` API ([`render.ts`](https://github.com/strapi/strapi/blob/main/packages/core/admin/admin/src/render.ts)):

```ts
import { createRoot } from 'react-dom/client';
// ...
createRoot(mountNode).render(app.render());
```

So any consumer of this design system is already on React 18. The `^17` peer range only advertised support that could never be exercised.

## Notes

`devDependencies` were already pinned to `18.3.1` — unchanged. No `ReactDOM.render` / version-check code paths exist in this repo (design-system, not the admin app), so the upstream PR's `createRoot`/batching notes don't apply to this repo's source.

### DIFF
--- a/.changeset/drop-react-17.md
+++ b/.changeset/drop-react-17.md
@@ -4,6 +4,4 @@
 "@strapi/ui-primitives": patch
 ---
 
-Drop React 17 from the supported peer dependency range.
-
-The `react` and `react-dom` peer dependencies are now `^18.0.0` (previously `^17.0.0 || ^18.0.0`), matching the React 18 APIs already in use. The `useId` hook's React 17 client-only fallback has also been removed.
+chore: drop React 17 from supported peer dependency range
--- a/.changeset/drop-react-17.md
+++ b/.changeset/drop-react-17.md
@@ -1,0 +1,9 @@
+---
+"@strapi/design-system": patch
+"@strapi/icons": patch
+"@strapi/ui-primitives": patch
+---
+
+Drop React 17 from the supported peer dependency range.
+
+The `react` and `react-dom` peer dependencies are now `^18.0.0` (previously `^17.0.0 || ^18.0.0`), matching the React 18 APIs already in use. The `useId` hook's React 17 client-only fallback has also been removed.
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -59,8 +59,8 @@
   },
   "peerDependencies": {
     "@strapi/icons": "^2.0.0 || ^2.0.0-beta || ^2.0.0-alpha",
-    "react": "^17.0.0 || ^18.0.0",
-    "react-dom": "^17.0.0 || ^18.0.0",
+    "react": "^18.0.0",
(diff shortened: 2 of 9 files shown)


nclsndr:
@unrevised6419 we currently have a npm issue slowing down releases of the DS. I'll get back to this PR whenever I get a green light.

nclsndr:
@HichamELBSI is there anything we must be careful of here?

.changeset/drop-react-17.md
```diff
@@ -0,0 +1,9 @@
+---
+"@strapi/design-system": patch
+"@strapi/icons": patch
+"@strapi/ui-primitives": patch
+---
+
+Drop React 17 from the supported peer dependency range.
```
Adzouz: Can you use conventional commit for the message format please? https://www.conventionalcommits.org/en/v1.0.0/

.changeset/drop-react-17.md
```diff
@@ -0,0 +1,9 @@
...
+
+The `react` and `react-dom` peer dependencies are now `^18.0.0` (previously `^17.0.0 || ^18.0.0`), matching the React 18 APIs already in use. The `useId` hook's React 17 client-only fallback has also been removed.
```
Adzouz: You can remove that part and keep a short message

Adzouz (changes requested):
Looks fine to me! I tested locally and everything seems to work as expected. I think we can move forward @nclsndr what do you think?
I can't be 100% sure about the impact on old projects / plugins but I think that's safe to merge.
Just a tiny comment to fix the changeset message 😊

Adzouz (approved):

nclsndr:
Thank you for the review @Adzouz!

nclsndr (approved):
The team is aligned on removing React 17 support and keep 18,19 for now
cc @derrickmehaffy

nclsndr:
@Adzouz can I let you merge this one? I'm interested to get pinged on release to follow up on https://github.com/strapi/strapi/pull/26194

PR #2025, opened 2026-06-16
fix: export color scheme from theme

Adzouz:
### What does it do?

Adds a `colorScheme` property to the `StrapiTheme` so consumers can tell which built-in scheme a theme is based on.

- Exports a new `ColorScheme` type (`'light' | 'dark'`) from `packages/design-system/src/themes`.
- Adds `colorScheme: ColorScheme` to the `StrapiTheme` interface.
- Sets `colorScheme: 'light'` on `lightTheme` and `colorScheme: 'dark'` on `darkTheme`.

The value is exposed through the `styled-components` theme, so it can be read in any styled component via `theme.colorScheme`.

### Why is it needed?

Some styles genuinely need to branch on light vs dark. Until now there was no reliable way to know the active scheme from the theme itself — consumers had to infer it or thread a separate prop. Exposing `colorScheme` on the theme gives a single source of truth.

> Note: prefer semantic color tokens whenever one exists; `colorScheme` is the escape hatch for the cases where branching is unavoidable.

### How to test it?

1. Launch Storybook:
   ```sh
   yarn install
   yarn build
   yarn develop
   ```
2. In any component, read and log the scheme from the theme:
   ```tsx
   import { useTheme } from 'styled-components';

   const { colorScheme } = useTheme();
   console.log('colorScheme:', colorScheme);
   ```
3. Open that component's page in Storybook and use the dark-mode toggle in the toolbar to switch the scheme.
4. Watch the browser console: the logged value flips between `'light'` and `'dark'` as you toggle.

### Related issue(s)/PR(s)

🚀

### DIFF
--- a/.changeset/spicy-days-breathe.md
+++ b/.changeset/spicy-days-breathe.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': patch
+---
+
+chore: add colorScheme to theme
--- a/packages/design-system/src/themes/__tests__/extendTheme.test.ts
+++ b/packages/design-system/src/themes/__tests__/extendTheme.test.ts
@@ -23,7 +23,8 @@ describe('extendTheme', () => {
       import { lightTheme, extendTheme } from '@strapi/design-system';
 
       const myCustomTheme = extendTheme(lightTheme, {
-          colors: /* put the overrides for the colors key */,
+          colorScheme: /* put the overrides for the colorScheme key */,
+      colors: /* put the overrides for the colors key */,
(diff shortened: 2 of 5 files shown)


mathildeleg (dismissed):
LGTM
Don't forget to add the changeset though :)

Adzouz:
> LGTM Don't forget to add the changeset though :)

hahaha dammit, well at least now I can see it 🤣

mathildeleg (dismissed):

mathildeleg (approved):

PR #2028, opened 2026-06-24
fix: upping node version to 20.19

cache-your-dreams:
### What does it do?

upping node version to 20.19

### Why is it needed?

There is a node version incompatibility needed for Trusted publishing that we want to put in place


mathildeleg (dismissed):

innerdvations (dismissed):
This is fine for now, but very soon we will bump it to 24

innerdvations (approved):

PR #2031, opened 2026-07-03
fix: forward crossOrigin on Avatar.Item

giu1io:
### What does it do?

`Avatar.Item` accepts a `crossOrigin` prop and forwards it to both `` elements it renders — the visible thumbnail (`AvatarImage`) and the hover-tooltip preview (`PreviewImg`).

Previously `ItemProps` only picked `onLoadingStatusChange | src | alt` from the underlying Radix image props, so `crossOrigin` was silently dropped even though `Avatar.AvatarImageProps` already supports it. This adds `crossOrigin` to the `Pick` and threads it through to both images.

### Why is it needed?

Companion change to a CMS-side fix for a CORS bug on private S3 buckets / signed URLs (strapi/strapi#26581). AWS S3 doesn't send `Vary: Origin` and only returns `Access-Control-Allow-Origin` when the request carries an `Origin` header. Chrome/Safari then cache a plain (no-CORS) response and reuse it for a later `crossOrigin="anonymous"` request, which fails the CORS check.

The Media Library list-view thumbnail renders via `Avatar.Item`, so with `crossOrigin` unforwardable it always loaded plain — poisoning the shared browser cache and breaking a subsequent cross-origin preview. Letting consumers pass `crossOrigin` makes every load use one consistent mode, so the browser caches a single CORS-valid response.

### How to test it?

- `yarn test:unit` — new tests in `Avatar.test.tsx` assert `crossorigin="anonymous"` reaches both the thumbnail and the hover-preview ``.
- Passing `crossOrigin` is opt-in and defaults to `undefined`, so existing usages are unchanged.

### Related issue(s)/PR(s)

- Closes the design-system gap for strapi/strapi#26581 (reference only — not closed by this PR).
- Follow-up required in `strapi/strapi`: bump `@strapi/design-system` and pass `crossOrigin` from `PreviewCell.tsx`.

### DIFF
--- a/.changeset/quick-tigers-reply.md
+++ b/.changeset/quick-tigers-reply.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': minor
+---
+
+Added crossOrigin support to the Avatar.Item component
--- a/.changeset/quick-tigers-reply.md
+++ b/.changeset/quick-tigers-reply.md
@@ -1,5 +1,5 @@
 ---
-'@strapi/design-system': minor
+'@strapi/design-system': patch
 ---
 
 fix: added crossOrigin support to the Avatar.Item component
--- a/.changeset/quick-tigers-reply.md
+++ b/.changeset/quick-tigers-reply.md
@@ -2,4 +2,4 @@
 '@strapi/design-system': minor
 ---
 
-Added crossOrigin support to the Avatar.Item component
+fix: added crossOrigin support to the Avatar.Item component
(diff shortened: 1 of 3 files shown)


.changeset/quick-tigers-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+'@strapi/design-system': minor
+---
+
+Added crossOrigin support to the Avatar.Item component
```
Adzouz: Can you use conventional commits format for the message in the changelog please? 🙏

Adzouz (changes requested):
Hi @giu1io, thanks a lot for the PR 🙌
I tested and it's working as expected. Just a small comment if you can address it before merging please 🙏

.changeset/quick-tigers-reply.md
giu1io: done! Thanks 🙏

Adzouz (approved):
Perfect, let's merge it!

PR #2033, opened 2026-07-08
chore: migrate from @radix-ui/* to unified radix-ui package

unrevised6419:
## What

Replace individual `@radix-ui/*` dependencies with the unified [`radix-ui`](https://www.radix-ui.com/primitives/docs/overview/releases) package (pinned at `1.6.7`) in the `design-system` and `primitives` workspaces.

> [!IMPORTANT]
> **Stacked on #2048.** That PR bumps the individual `@radix-ui/*` packages to their `radix-ui@1.1.0`-era versions without switching to the unified package, isolating the version-bump breakage from the import rewrite. Its commit rides along here until it merges.
>
> To review only the migration work, excluding the precursor:
> [`compare/2e5d0c82...94a98a70`](https://github.com/strapi/design-system/compare/2e5d0c82...94a98a70)

## This is a runtime upgrade, not just an import rewrite

The individual packages were pinned to their 1.0.x-era releases; `radix-ui@1.6.7` bundles the current primitives. Roughly two years of Radix behaviour lands at once across every overlay, menu and dialog. Changes that reach components in this repo include:

- **Checkbox / Switch / RadioGroup** — programmatic value changes inside a `<form>` no longer dispatch a `click` from the hidden bubble input that propagated to ancestor `onClick` handlers; values now update on form reset; bubble inputs render through `Primitive`
- **DropdownMenu / Tabs** — items and triggers no longer intercept `Space`/`Enter` originating from focusable descendants; menus no longer stay open after the window loses focus
- **Tooltip** — `Tooltip.Content` children were being mounted to the DOM twice; `skipDelayDuration={0}` now behaves as documented
- **Dialog / AlertDialog** — broken ARIA references when title/description are not rendered; dev-only warnings removed
- **ScrollArea** — `--radix-scroll-area-corner-{width,height}` now reset to `0`; `data-state` added to scrollbars
- **DismissableLayer** — nested popovers no longer close all layers on outside click; `pointer-events: none` no longer stranded on `body`
- **Avatar** — loading-state edge cases; zero-sized images now report `error` rather than hanging in `loading`

The Avatar change required a test fix in this PR (`Avatar.test.tsx`), since Radix now derives loaded state from `image.complete`/`naturalWidth` rather than the `load` event alone. Consumers should re-verify these components rather than treating this as a routine dependency bump.

## Changes

- Components import from `radix-ui`, internal utilities from `radix-ui/internal`
- Inline `clamp` (was `@radix-ui/number`) and `usePrevious` (was `@radix-ui/react-use-previous`) — neither is shipped by the unified `radix-ui` package
- Adapt to bundled newer primitive versions: `React.ComponentPropsWithoutRef`, `useControllableState` v1.2 (required `defaultProp`/`caller`)
- Pin `@types/react` to `18.3.3` via root `resolutions` — see below
- Document `radix-ui/internal` as an unstable surface at both import sites, so a future `radix-ui` bump re-verifies those imports

### Why `@types/react` needs a resolution (temporary — pre-existing issue)

**This duplicate already exists on `main`, independently of this PR.** `main`'s `yarn.lock` contains both `"@types/react@npm:*"` resolving to `19.1.9` and `"@types/react@npm:18.3.3"`. The cause is `@types/react-dom@18.3.0`, which declares a real (non-peer) dependency `"@types/react": "*"` that resolves to 19.x.

The migration does not introduce the duplicate — it makes it bite. Radix's published `.d.ts` files import React types, so the stray 19.x copy starts leaking into typechecking where previously nothing consumed it. The `resolutions` block here is a workaround to keep this PR green.

The proper fix is upstream and already released: from `@types/react-dom@18.3.3` the dependency became a peer constrained to `^18.0.0`, making a nested 19.x copy impossible. Bumping `@types/react-dom` to `18.3.7` removes the need for any resolution. That is handled separately in #2049, against `main`. Once it lands, the `resolutions` block will be dropped from this PR and the lockfile regenerated.

## Review feedback

Addressed from @Adzouz's review:

- Replaced the unsound `as string` cast on Combobox's `textValue` default with an explicit `useControllableState<string | undefined>` — `undefined` is a meaningful state there, as the placeholder rendering checks for it
- Documented the `radix-ui/internal` unstable surface at both import sites
- Moved `const` aliases that were interleaved with import blocks below the last import (DatePicker, DesignSystemProvider, Combobox, Select). This made the import blocks contiguous and surfaced pre-existing `import/order` violations in those same files, fixed here via `eslint --fix` — note `yarn lint` runs oxlint, which has no `import/order` rule, so these are only visible to editors using the leftover ESLint config

Still open: widening the public `onValueChange`/`onFilterValueChange` types to `string | undefined`. Both are genuinely unsound today — several internal call sites pass `undefined` — but they are public `RootProps` changes, so they need a check against usage in the Strapi monorepo first.

## Why

**Easier upgrade path.** With individually-versioned `@radix-ui/*` packages, nothing guarantees that the versions you install work together — in theory they can drift out of sync. The unified `radix-ui` package updates every primitive at once as a single, mutually-compatible set, so upgrades are one bump instead of a coordination exercise across dozens of packages. Single dependency, single React tree, fewer version-skew bugs.

**Precursor to a possible Base UI migration.** In the current shadcn/ui ecosystem, Radix is no longer the default primitive library, and there's a real risk it gets left behind over time — [shadcn himself has signalled the shift toward Base UI](https://x.com/shadcn/status/2073021571342520343). Consolidating onto the unified package now is a natural stepping stone toward [Base UI](https://base-ui.com/) if we choose to move later — Base UI ships codemods for Radix and is close to 1:1 API parity, so the eventual switch would be low-friction.

## Test plan

- [x] `yarn test:ts` — 6/6 tasks
- [x] `yarn lint` — 6/6 tasks, 0 errors
- [x] `yarn test:unit` — 36 suites, 322 tests, 33 snapshots
- [x] ESLint clean on all changed files
- [ ] Non-regression testing on the components listed above (maintainer-side, per review)


packages/primitives/src/components/Combobox/Combobox.tsx
```diff
@@ -107,6 +113,7 @@ interface RootProps {
   disabled?: boolean;
   locale?: string;
   onOpenChange?(open: boolean): void;
+  // TODO: This might need to be `value: string | undefined`
   onValueChange?(value: string): void;
```
Adzouz, line 123: This TODO can be resolved in this PR — the answer is yes, the type should be `onValueChange?(value: string | undefined): void`.

Evidence that `undefined` is genuinely delivered at runtime (pre-existing behavior, also on `main`):
- The internal context already types it that way (`onValueChange(value: string | undefined)`, line 80).
- The Escape-clear paths call `context.onValueChange(undefined)` (lines 547 and 679), which flows through `setValue` → `useControllableState.onChange` → the consumer's `onValueChange`.

So the public `string` type has been lying since before this PR; the new v1.2 generics just forced the cast that exposes it. Widening the prop type here:
- lets you delete the `as ((value: string | undefined) => void) | undefined` cast on line 218,
- automatically propagates to the design-system `Combobox` (its `onChange` is typed as `ComboboxPrimitive.RootProps['onValueChange']`),
- won't break consumers' typechecks — method-style signatures are bivariant, so existing `(value: string) => void` handlers still assign.

Suggest doing it now rather than merging a known-incorrect type with a TODO.

packages/primitives/src/components/Combobox/Combobox.tsx
```diff
@@ -115,6 +122,7 @@ interface RootProps {
   value?: string;
   defaultFilterValue?: string;
   filterValue?: string;
+  // TODO: This might need to be `value: string | undefined`
   onFilterValueChange?(value: string): void;
```
Adzouz, line 132: This TODO can also be resolved now. Unlike `onValueChange`, none of the 8 internal `context.onFilterValueChange(...)` call sites ever pass `undefined` — they all pass strings. So strictly, `string` is accurate today.

That said, the state is declared `useControllableState<string | undefined>` and the context types it `(value: string | undefined) => void` (line 95), so I'd still widen the public type to `string | undefined` for consistency — it lets you drop the cast on line 230 and matches the actual state domain (`filterValue` can legitimately be `undefined`, see the `context.filterValue === undefined` checks). Either way, the TODO should not ship.

packages/primitives/src/components/Combobox/Combobox.tsx
```diff
@@ -200,23 +208,27 @@ const Combobox = (props: RootProps) => {
...
     prop: valueProp,
     defaultProp: defaultValue,
-    onChange: onValueChange,
+    onChange: onValueChange as ((value: string | undefined) => void) | undefined,
+    caller: COMBOBOX_NAME,
   });
   const [textValue, setTextValue] = useControllableState({
     prop: textValueProp,
-    defaultProp: allowCustomValue && !defaultTextValue ? valueProp : defaultTextValue,
+    defaultProp: (allowCustomValue && !defaultTextValue ? valueProp : defaultTextValue) as string,
```
Adzouz: This `as string` cast is unsound: `valueProp` and `defaultTextValue` are both optional, so the whole expression can be `undefined` — and `undefined` is a *meaningful* state here (placeholder rendering checks `context.textValue === undefined`, e.g. the `data-placeholder` logic).

Suggest typing it explicitly like the sibling states instead of casting the value:

```ts
const [textValue, setTextValue] = useControllableState<string | undefined>({
  prop: textValueProp,
  defaultProp: allowCustomValue && !defaultTextValue ? valueProp : defaultTextValue,
  onChange: onTextValueChange as ((value: string | undefined) => void) | undefined,
  caller: COMBOBOX_NAME,
});
```

(And if `onTextValueChange` gets the same `string | undefined` widening as the other two callbacks, that cast disappears too.)

.changeset/unified-radix-ui.md
```diff
@@ -0,0 +1,6 @@
+---
+"@strapi/design-system": patch
```
Adzouz, line 2: The import rewrite is mechanical, but `radix-ui@1.6.0` bundles the *current* primitive versions while the individual packages were pinned to ~2023 versions. It's ~2 years of Radix behavior changes landing at once — focus handling, pointer-events on dismissable layers, positioning — i.e. a real runtime upgrade for every overlay/menu/dialog in consumers, not just a dependency rename.

The description should also say this explicitly: Radix primitives are upgraded from the 1.0.x era to the versions bundled in `radix-ui@1.6.0`, and the radix install nests React 19 types — consumers on React 18 may want an `@types/react` resolution like the one added to this repo's root `package.json`.

packages/primitives/src/components/Select/Select.tsx
```diff
@@ -13,33 +13,42 @@
...
+  DismissableLayer as DismissableLayerPrimitive,
+  FocusGuards,
+  FocusScope as FocusScopePrimitive,
+  Popper as PopperPrimitive,
+  Primitive,
+  composeEventHandlers,
+  useComposedRefs,
+  useControllableState,
+  useLayoutEffect,
+} from 'radix-ui/internal';
```
Adzouz, line 43: Worth a note (here and in `Combobox.tsx`): `radix-ui/internal` is Radix's explicitly-unstable surface — no semver guarantees, unlike the individually-versioned public packages these imports replace. Since this package is a fork of Radix Select anyway the dependency on internals is somewhat unavoidable, and the exact `1.6.0` pin protects us. But I'd add a short comment at the import site saying bumps of `radix-ui` must re-verify the `radix-ui/internal` imports, so a future "routine" dependency bump doesn't silently hit a moved/removed internal API.

packages/design-system/src/components/DatePicker/DatePicker.tsx
```diff
@@ -16,8 +16,10 @@ import {
   parseAbsolute,
   getLocalTimeZone,
 } from '@internationalized/date';
-import { useFocusGuards } from '@radix-ui/react-focus-guards';
-import { FocusScope } from '@radix-ui/react-focus-scope';
+import { FocusGuards, FocusScope as FocusScopePrimitive } from 'radix-ui/internal';
+
+const FocusScope = FocusScopePrimitive.FocusScope;
+const useFocusGuards = FocusGuards.useFocusGuards;
```
Adzouz: Nit: these `const` re-exports are declared in the middle of the import block (imports continue below on line 23). Valid JS since imports hoist, and oxlint doesn't flag it, but it reads oddly. Suggest either moving the aliases below the last import, or destructuring in place where possible. Same pattern in `Combobox.tsx`, `Select.tsx`, `Collection.tsx`, and `DesignSystemProvider.tsx`.

Adzouz (changes requested):
Hi @unrevised6419, thanks a lot for the PR, your efficiency and many contributions for Strapi, it really means a lot for us 🙌
About this one, it's a bigger change than just merging all packages into one since we're also bumping the version (moving from ~2 years old packages to the latest ones) we have to run non-regression tests to be sure that the impacted components are still working as before.
In the meantime, here's a few comments that would need to be addressed while we're testing on our side 🙏

packages/primitives/src/components/Combobox/Combobox.tsx
unrevised6419: This could be a type "breaking change". I will need first to do some checks in Strapi monorepo to see how is used there.

unrevised6419:
Precursor: #2048 bumps the individual `@radix-ui/*` packages to their `radix-ui@1.1.0`-era versions without switching to the unified `radix-ui` package. Worth landing that first — it isolates the version-bump breakage from the unified-package migration itself.

The two type fixes it carries: `@radix-ui/react-primitive` dropped `ComponentPropsWithoutRef`/`PropsWithoutRef` in its 1.0.3 → 2.0.1 major bump, and `Menu.tsx` referenced `DropdownMenu.MenuItemProps`, which `@radix-ui/react-dropdown-menu` does not export (at 2.0.6 and 2.1.5 alike it is a local, non-exported alias — the exported name is `DropdownMenuItemProps`). Both previously resolved to `any`, which is why neither errored before.

_(Edited: an earlier version of this comment described the second item as a rename between versions. That was incorrect — the type was never exported.)_

PR #2034, opened 2026-07-10
fix(design-system): externalize @codemirror/state in JSONInput bundle

### LINKED ISSUE #2032: [bug]: JSON input crashes with "Unrecognized extension value... multiple instances of @codemirror/state" in production builds (regression of #1706)
### What version of @strapi/design-system are you using?

2.2.0 (also present in 2.2.1, the current latest). Strapi 5.48.

### What's wrong?

Any admin edit view that renders the **JSON input** (a `json` field, or a component containing one — e.g. the SEO component's `structuredData`) crashes the whole page:

```
Error: Unrecognized extension value in extension set ([object Object]).
This sometimes happens because multiple instances of @codemirror/state are
loaded, breaking instanceof checks.
```

React Router's error boundary catches it, so the entry becomes uneditable.

This is a **regression of #1706** (fixed for 2.0.0-beta.2 in 2024, back in 2.2.x).

### Root cause

`@strapi/design-system`'s published `dist/index.mjs` loads **two different instances of `@codemirror/state`**:

1. It **inlines** `@codemirror/state` into its own bundle — the `Facet`/`StateField`/`StateEffect` classes are compiled directly into `dist/index.mjs` (the "Unrecognized extension value" throw site lives inside the design-system bundle itself).
2. It **imports `@uiw/react-codemirror` and `@codemirror/lang-json` as externals** — and those resolve the **standalone** `@codemirror/state` from `node_modules`.

The JSON input builds a decoration `StateField` (the `xp` local in the minified dist) with the **inlined** copy, then passes it in `extensions: [json(), xp]` to the `@uiw/react-codemirror` editor, which runs on the **standalone** copy. CodeMirror validates extensions via `instanceof` against its own `@codemirror/state`, so the field built from the *other* instance is rejected → throw.

Confirmed in 2.2.0's `dist/index.mjs`:
- inlined `@codemirror/state`: **1** copy (the error-string / Facet classes are in the bundle)
- external imports: `@uiw/react-codemirror` and `@codemirror/lang-json`

Because both copies are baked into the **published** `dist`, consumers can't fix it with npm `overrides`, `resolve.dedupe`, or lockfile changes — the duplication is inside the shipped artifact.

### Why it only shows up in production

It reproduces only in **production builds** (e.g. Dockerized `strapi build` on linux). In local `strapi develop` the app installs against a lockfile that dedupes design-system to a single copy and the dev bundler happens to collapse the instances, so it doesn't throw there — which makes it easy to ship unnoticed.

### To Reproduce

1. Strapi 5.48 project with a `json` field (or a component with one, e.g. SEO `structuredData`).
2. `strapi build` + `strapi start` (production), ideally in a linux/alpine container.
3. Open the entry's edit view → crash with the error above.

### Expected

The JSON input mounts and is editable.

### Suggested fix

Mark `@codemirror/*` (`@codemirror/state`, `@codemirror/view`, `@codemirror/language`, …) as **external** in design-system's build config, so design-system shares the single `@codemirror/state` instance used by `@uiw/react-codemirror` / `@codemirror/lang-json` instead of inlining its own copy.

### Workaround (for others hitting this)

`patch-package` to drop the inlined-copy extension from the JSON editor's extension set:

```diff
- extensions: [Oh(), xp],
+ extensions: [Oh()],
```

`Oh()` is `@codemirror/lang-json` (same standalone instance as the editor, so JSON highlighting is kept); `xp` is the design-system decoration `StateField` built from the inlined copy — the only value crossing instance boundaries. Verified this makes the JSON input render and stay editable in a production alpine build.

innerdvations:
### What does it do?

Declares singleton dependencies in `@strapi/design-system` so the Vite build **externalizes** them instead of inlining copies into `dist/index.mjs`:

- `@codemirror/state@6.7.1` and `@codemirror/view@6.43.0` — used by `JSONInput` error-line highlighting
- `@tanstack/react-virtual@^3.10.8` — used by Combobox `VirtualizedList`

Adds a **bundle contract test** (`bundle-codemirror.test.ts`) that fails when those packages are inlined into the published dist.

Fix #2032

### Why is it needed?

Production admin edit views can crash when a **JSON field** mounts (or a component containing one, e.g. SEO `structuredData`):

```
Unrecognized extension value in extension set ([object Object]).
This sometimes happens because multiple instances of @codemirror/state are loaded, breaking instanceof checks.
```

**In plain terms:** the JSON editor and a custom highlight helper inside design-system were each using a *different copy* of CodeMirror. CodeMirror checks “is this extension from *my* copy?” — when the answer is no, it throws and React Router blocks the whole edit view.

This is a **regression of #1706 / #1709** after the packup → Vite migration: `@codemirror/state` and `@codemirror/view` were imported in source but not declared in `package.json`, so `vite-plugin-externalize-deps` inlined them into dist while `@uiw/react-codemirror` still loads CodeMirror from `node_modules`.

The same undeclared-import → inlined pattern affected `@tanstack/react-virtual` in Combobox (already correctly external in `@strapi/ui-primitives`).

**Why it surfaced recently:** the bundling bug has been latent since DS ≥ 2.1.0, but **`@codemirror/state@6.7.1` (July 5)** changed duplicate-instance extensions from silently ignored to an explicit throw — so fresh lockfile installs started crashing without a Strapi release.

Also reported as [strapi/strapi#26951](https://github.com/strapi/strapi/issues/26951).

#### Affected versions

| What matters | Detail |
|--------------|--------|
| **Broken** | `@strapi/design-system` **≥ 2.1.0** through **2.2.2** |
| **Not broken** | `@strapi/design-system` **2.0.0** and earlier |
| **Strapi** | Any Strapi 5.x that resolves to an affected DS version. Rolling back Strapi alone does **not** fix it if the broken DS bundle is still installed. |

### How to test it?

1. `yarn install && yarn build --filter=@strapi/design-system`
2. `cd packages/design-system && npx jest -c jest.config.mjs --testPathPattern=bundle-codemirror` — should pass
3. Sanity check on dist:

   ```bash
   grep -c "Unrecognized extension value" packages/design-system/dist/index.mjs
   # expect 0

   grep "@codemirror/state" packages/design-system/dist/index.mjs
   grep "@tanstack/react-virtual" packages/design-system/dist/index.mjs
   # expect external import lines
   ```

4. After release: bump `@strapi/design-system` in strapi/strapi, production build (`yarn build && yarn start`), open a Content Manager edit view with a **json** field — should load without console errors.

### Related issue(s)/PR(s)

- Fixes #2032
- Regression of #1706 / #1709
- [strapi/strapi#26951](https://github.com/strapi/strapi/issues/26951)

### DIFF
--- a/.changeset/fix-codemirror-bundle-externals.md
+++ b/.changeset/fix-codemirror-bundle-externals.md
@@ -2,8 +2,4 @@
 '@strapi/design-system': patch
 ---
 
-fix(design-system): externalize singleton deps in published bundle
-
-Declare @codemirror/state, @codemirror/view, and @tanstack/react-virtual as
-dependencies so the Vite build externalizes them instead of inlining copies
-into dist. Fixes production admin crashes (strapi/strapi#26951, #2032).
+fix: externalize singleton deps in published bundle
--- a/.changeset/fix-codemirror-bundle-externals.md
+++ b/.changeset/fix-codemirror-bundle-externals.md
@@ -1,0 +1,7 @@
+---
+'@strapi/design-system': patch
+---
+
+fix(design-system): externalize @codemirror/state and @codemirror/view in JSONInput bundle
+
+Fixes regression of #1706 where the Vite build inlined @codemirror/state while @uiw/react-codemirror uses the package from node_modules, causing production admin crashes (strapi/strapi#26951, #2032).
--- a/.changeset/fix-codemirror-bundle-externals.md
+++ b/.changeset/fix-codemirror-bundle-externals.md
@@ -2,6 +2,8 @@
 '@strapi/design-system': patch
 ---
 
-fix(design-system): externalize @codemirror/state and @codemirror/view in JSONInput bundle
+fix(design-system): externalize singleton deps in published bundle
 
-Fixes regression of #1706 where the Vite build inlined @codemirror/state while @uiw/react-codemirror uses the package from node_modules, causing production admin crashes (strapi/strapi#26951, #2032).
+Declare @codemirror/state, @codemirror/view, and @tanstack/react-virtual as
+dependencies so the Vite build externalizes them instead of inlining copies
+into dist. Fixes production admin crashes (strapi/strapi#26951, #2032).
--- a/package.json
+++ b/package.json
@@ -77,6 +77,9 @@
     "typescript": "^5.1.3"
   },
   "packageManager": "yarn@3.6.4",
+  "resolutions": {
+    "@codemirror/state": "6.7.1"
+  },
   "dependencies": {
     "@vueless/storybook-dark-mode": "9.0.6"
   }
--- a/package.json
+++ b/package.json
@@ -77,9 +77,6 @@
     "typescript": "^5.1.3"
   },
   "packageManager": "yarn@3.6.4",
-  "resolutions": {
-    "@codemirror/state": "6.7.1"
-  },
   "dependencies": {
     "@vueless/storybook-dark-mode": "9.0.6"
   }
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -18,6 +18,8 @@
   ],
   "dependencies": {
     "@codemirror/lang-json": "6.0.1",
+    "@codemirror/state": "6.7.1",
+    "@codemirror/view": "6.43.0",
     "@floating-ui/react-dom": "2.1.0",
     "@internationalized/date": "3.5.4",
     "@internationalized/number": "3.5.3",
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -41,6 +41,7 @@
     "@radix-ui/react-tooltip": "1.0.7",
     "@radix-ui/react-use-callback-ref": "1.0.1",
     "@strapi/ui-primitives": "2.2.2",
+    "@tanstack/react-virtual": "^3.10.8",
     "@uiw/react-codemirror": "4.22.2",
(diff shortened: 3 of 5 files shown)


.changeset/fix-codemirror-bundle-externals.md
```diff
@@ -0,0 +1,7 @@
+---
+'@strapi/design-system': patch
+---
+
+fix(design-system): externalize @codemirror/state and @codemirror/view in JSONInput bundle
+
+Fixes regression of #1706 where the Vite build inlined @codemirror/state while @uiw/react-codemirror uses the package from node_modules, causing production admin crashes (strapi/strapi#26951, #2032).
```
Adzouz: Keep it short with conventional commit convention, you can remove this part

.changeset/fix-codemirror-bundle-externals.md
```diff
@@ -0,0 +1,7 @@
+---
+'@strapi/design-system': patch
+---
+
+fix(design-system): externalize @codemirror/state and @codemirror/view in JSONInput bundle
```
Adzouz: No need to precise the package here, it's precised above in the changeset header and will be automatically moved to the right changelog

Adzouz (changes requested):
Thanks for the reactivity @innerdvations, just two small comments and we can merge it 🙏

innerdvations:
Addressed both changeset comments in c6909c8b — short `fix:` summary, no package scope or long body.

Adzouz (approved):
Let's merge that baby

PR #2036, opened 2026-07-13
chore: bump Node toolchain to 22

jhoward1994:
requires Node 22+


Bassel17:
shall we bump the design-system to use 22-26 instead ?
it's mostly a tooling update so perhaps we can just do that

jhoward1994:
Does that diff seem okay now ?

Bassel17 (approved):
LGTM!

PR #2040, opened 2026-08-02
fix: scrollbars and native form controls ignore dark theme

### LINKED ISSUE #947: [feat]: Use css color-scheme to improve dark theme

The css `color-scheme` property allows to set some native browser features as dark or light themed when supported. 

The usecase I came across is the Chrome scrollbar. 

It is obvisouly not as perfect as designing custom things but that works much better than having the light scroll in our dark theme for almost no cost. 

It might be something we can set directly in the product rather than in the DS but putting it outhere to open the discussion and see if there are drawbacks to doing it. 

```css
color-scheme: dark;
```

emmyKcodes:
## What does it do?

Fixes #947.

The Design System sets a `dark` theme via `styled-components`, but never tells
the browser about it. Because of that, browser-native UI — scrollbars,
`<select>` dropdowns, `<input type="date">`, `<input type="color">`, etc. —
keeps rendering with the OS/browser default (usually light), even when the
app is running in dark mode. This creates a jarring visual mismatch where the
app chrome is dark but native form/browser elements stay light.

The fix applies the CSS `color-scheme` property to `body`, driven by the
theme's existing `colorScheme` token (`'light' | 'dark'`), which was already
present in the theme definition but never actually consumed anywhere:

\`\`\`css
body {
  ...
  color-scheme: ${({ theme }) => theme.colorScheme};
}
\`\`\`

This lets the browser theme its own native elements to match the active
Strapi theme automatically, using a browser-native mechanism rather than
custom overrides.

## How to test it?

1. Run `yarn develop` and open Storybook.
2. Toggle the theme switcher to **dark**.
3. Open DevTools → Elements → select `<body>` → check the Computed panel:
   `color-scheme` should read `dark`.
4. Look at the page's scrollbar (or open a story with a native `<select>` /
   date input) — it should now render dark instead of the default light
   style.
5. Toggle back to **light** and confirm `color-scheme` on `<body>` reverts to
   `light`, and native elements go back to their light appearance.

**Before:** scrollbars/native inputs stay light even in dark theme.
**After:** scrollbars/native inputs match the active theme automatically.

No automated test was added since this is a global CSS-only change with no
existing test coverage for `global.ts`, and it isn't meaningfully testable
via unit tests — verification is visual, as described above.

### DIFF
--- a/.changeset/nice-gifts-cheer.md
+++ b/.changeset/nice-gifts-cheer.md
@@ -1,0 +1,5 @@
+---
+'@strapi/design-system': patch
+---
+
+Apply color-scheme so native browser UI (scrollbars, form controls) matches the active theme
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,4 @@ CLAUDE.local.md
 
 # Codex
 .agents/
+docs/%SystemDrive%/
--- a/packages/design-system/src/styles/global.ts
+++ b/packages/design-system/src/styles/global.ts
@@ -15,6 +15,7 @@ ${css`
   html {
     /* Sets 1rem === 10px */
     font-size: 62.5%;
+    color-scheme: ${({ theme }) => theme.colorScheme};
   }
 
   body {


innerdvations:
Thanks for the PR — wiring `theme.colorScheme` into global styles is a clean way to address #947.

Could you tidy up the branch before we merge?

1. **Please remove** the accidental Windows cache files under `docs/%SystemDrive%/ProgramData/Microsoft/Windows/Caches/` (the `.db` files). They look unrelated to the fix and shouldn’t ship in the repo.
2. **Please move** `color-scheme` from `body` to the existing `html` rule. Viewport scrollbars follow the root element, so putting it on `html` better matches the scrollbar behaviour called out in the issue (descendant form controls still inherit either way).

Optional polish: the changeset text currently reads like the bug statement; a short “fix …” / “apply color-scheme so …” phrasing would read better in the changelog. Also noting there’s an older overlapping attempt in #2007 — once this lands, we can close that one as superseded.

Happy to take another look after those updates.

_AI-assisted comment._

emmyKcodes:
Thanks for the review! Updated:
- Removed the accidental Windows cache files and added a .gitignore entry
- Moved `color-scheme` to the `html` rule as suggested
- Reworded the changeset summary
Good catch on #2007, happy to have this supersede it once merged.

emmyKcodes:
> Thanks for the PR — wiring `theme.colorScheme` into global styles is a clean way to address #947.
> 
> Could you tidy up the branch before we merge?
> 
> 1. **Please remove** the accidental Windows cache files under `docs/%SystemDrive%/ProgramData/Microsoft/Windows/Caches/` (the `.db` files). They look unrelated to the fix and shouldn’t ship in the repo.
> 2. **Please move** `color-scheme` from `body` to the existing `html` rule. Viewport scrollbars follow the root element, so putting it on `html` better matches the scrollbar behaviour called out in the issue (descendant form controls still inherit either way).
> 
> Optional polish: the changeset text currently reads like the bug statement; a short “fix …” / “apply color-scheme so …” phrasing would read better in the changelog. Also noting there’s an older overlapping attempt in #2007 — once this lands, we can close that one as superseded.
> 
> Happy to take another look after those updates.
> 
> _AI-assisted comment._

Thanks for the review! Updated:

Removed the accidental Windows cache files and added a .gitignore entry
Moved color-scheme to the html rule as suggested
Reworded the changeset summary
Good catch on https://github.com/strapi/design-system/pull/2007, happy to have this supersede it once merged.

PR #2042, opened 2026-08-05
fix: keep scroll position when focusing a cell by pointer

Adzouz:
### What does it do?

`RawTable` moved roving focus to the active cell on **every** `rowIndex`/`colIndex` change — including pointer and programmatic ones reported through `setTableValues`.

It now tracks a `shouldSkipFocusRef`:

- `setTableValues` (the pointer/programmatic path) sets it, so the focus effect **skips** `focusFocusable` — the element the pointer focused keeps focus.
- `handleKeyDown` clears it, so keyboard navigation still moves focus to the roving cell (and scrolls it into view).

Also adds `RawTable.test.tsx` (keyboard nav moves focus; pointer focus does not) and a changeset (patch).

### Why is it needed?

Clicking a control (e.g. a row's checkbox) far down a scrolled `Table` refocused the roving cell — the first header cell once the index had settled back to `(0, 0)` — and the browser scrolled that cell into view, snapping the list back to the top. The selection applied correctly, but the scroll position was lost on every tick, making bulk-select down a long list impractical.

`RawTable` is shared, so this affects any sufficiently long Strapi table (found during Media Library QA; the Content Manager list view is equally affected).

### How to test it?

Automated:

```
cd packages/design-system
yarn test:unit src/components/RawTable/RawTable.test.tsx
```

Manual (in a consumer — e.g. Strapi's new Media Library **Table view**, or the Content Manager list):

1. Open a table with enough rows to scroll well past one viewport.
2. Scroll to the bottom.
3. Tick a row's checkbox with the **mouse**.

**Expected:** the row is selected and the scroll position stays put. Keyboard arrow navigation still scrolls the focused cell into view.

### Related issue(s)/PR(s)

fix CMS-1567

🚀

### DIFF
--- a/packages/design-system/src/components/RawTable/RawTable.test.tsx
+++ b/packages/design-system/src/components/RawTable/RawTable.test.tsx
@@ -1,0 +1,68 @@
+import { render as renderRTL } from '@test/utils';
+import { fireEvent, screen } from '@testing-library/react';
+
+import { RawTd, RawTh } from './RawCell';
+import { RawTable } from './RawTable';
+import { RawTbody } from './RawTbody';
+import { RawThead } from './RawThead';
+import { RawTr } from './RawTr';
+
+import { focusFocusable } from './focusFocusable';
+
+jest.mock('./focusFocusable', () => ({
+  focusFocusable: jest.fn(),
+}));
+
+const focusFocusableMock = focusFocusable as jest.Mock;
+
+const render = () =>
+  renderRTL(
+    <RawTable colCount={1} rowCount={3}>
+      <RawThead>
+        <RawTr>
+          <RawTh>
+            <button type="button">Header</button>
+          </RawTh>
+        </RawTr>
+      </RawThead>
+      <RawTbody>
+        <RawTr>
+          <RawTd>
+            <button type="button">Row one</button>
+          </RawTd>
+        </RawTr>
+        <RawTr>
+          <RawTd>
+            <button type="button">Row two</button>
+          </RawTd>
+        </RawTr>
+      </RawTbody>
+    </RawTable>,
+  );
+
+describe('RawTable roving focus', () => {
+  beforeEach(() => {
+    focusFocusableMock.mockClear();
+  });
+
+  it('moves the roving focus when the index changes via keyboard navigation', () => {
+    render();
+    focusFocusableMock.mockClear(); // ignore mount
+
+    fireEvent.keyDown(screen.getByRole('grid'), { key: 'ArrowDown' });
+
+    expect(focusFocusableMock).toHaveBeenCalled();
+  });
+
+  it('does not move the roving focus when a cell is focused by pointer', () => {
+    // A pointer focus reports its coords through `setTableValues`; re-focusing
+    // the roving cell there would scroll it into view and jump a scrolled table
+    // to the top. The element the user clicked already holds focus.
+    render();
+    focusFocusableMock.mockClear(); // ignore mount
+
+    fireEvent.focus(screen.getByRole('button', { name: 'Row two' }));
+
+    expect(focusFocusableMock).not.toHaveBeenCalled();
+  });
+});
--- a/packages/design-system/src/components/RawTable/RawTable.test.tsx
+++ b/packages/design-system/src/components/RawTable/RawTable.test.tsx
@@ -54,10 +54,11 @@ describe('RawTable roving focus', () => {
     expect(focusFocusableMock).toHaveBeenCalled();
   });
 
-  it('does not move the roving focus when a cell is focused by pointer', () => {
-    // A pointer focus reports its coords through `setTableValues`; re-focusing
-    // the roving cell there would scroll it into view and jump a scrolled table
-    // to the top. The element the user clicked already holds focus.
+  it('does not move the roving focus when a cell reports focus through `setTableValues`', () => {
+    // A cell reports focus through `setTableValues` whenever the focus did not
+    // come from the table's own keyboard navigation — a click or a tab both land
+    // here. Re-focusing the roving cell would scroll it into view and jump a
+    // scrolled table to the top; the element that took focus already holds it.
     render();
     focusFocusableMock.mockClear(); // ignore mount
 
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/RawTable/RawTable.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+import { RawTd, RawTh } from './RawCell';
+import { RawTable } from './RawTable';
+import { RawTbody } from './RawTbody';
+import { RawThead } from './RawThead';
+import { RawTr } from './RawTr';
+
+import { focusFocusable } from './focusFocusable';
+
+jest.mock('./focusFocusable', () => ({
```
giu1io, line 12: **The test mocks the very thing the bug was about, so it can't catch a regression.**

`jest.mock('./focusFocusable')` means both tests assert on "was this function called", not on where focus actually ended up. The reported bug was a *scroll jump* caused by real focus moving to the header cell. A regression that moved focus by some other path — or that called `focusFocusable` with the wrong node — would still pass both tests.

I checked: without the mock, jsdom tracks this fine. Asserting on `document.activeElement` works today and tests the real behaviour:

```js
fireEvent.focus(screen.getByRole('button', { name: 'Row two' }));
expect(document.activeElement).toHaveTextContent('Row two'); // stays put
```

```js
screen.getByRole('button', { name: 'Header' }).focus();
fireEvent.keyDown(screen.getByRole('grid'), { key: 'ArrowDown' });
expect(document.activeElement).toHaveTextContent('Row one'); // focus followed
```

Worth noting this is the only `jest.mock` of an internal module in the package, so it also cuts against local convention. Not blocking — the fix itself is right — but the tests would be worth more.

_Generated by Claude, edited by me._

packages/design-system/src/components/RawTable/RawTable.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+  it('moves the roving focus when the index changes via keyboard navigation', () => {
+    render();
+    focusFocusableMock.mockClear(); // ignore mount
+
+    fireEvent.keyDown(screen.getByRole('grid'), { key: 'ArrowDown' });
+
+    expect(focusFocusableMock).toHaveBeenCalled();
+  });
+
+  it('does not move the roving focus when a cell is focused by pointer', () => {
```
giu1io: **`fireEvent.focus` doesn't distinguish pointer from keyboard, so the test name overstates what's covered.**

The test is called "focused by pointer", but `RawCell` listens for the plain `focus` event — it never sees the input modality. Tabbing to that button fires the same event and takes the same branch.

That's fine, and I confirmed keyboard nav still works (the `handleKeyDown` reset covers it). But the name suggests a pointer-vs-keyboard distinction the code doesn't actually make. Maybe "when a cell reports focus through `setTableValues`" — describes the real trigger.

_Generated by Claude, edited by me._

giu1io (dismissed):
QA looks good! ✅ 

Added 2 small comments but they are not blocking.

packages/design-system/src/components/RawTable/RawTable.test.tsx
Adzouz: Fair point — `RawCell` listens on a plain `focus` event, so it genuinely can't tell a click from a tab. Renamed to your suggestion in 0fb69c37, and reworded the comment above it since it carried the same misleading framing ("a pointer focus reports its coords…" → a click *or* a tab both land there).

_Reply drafted by AI, reviewed by @Adzouz._

packages/design-system/src/components/RawTable/RawTable.test.tsx
Adzouz: The convention point is right — this is the only `jest.mock` of an internal module in the package. But I tried the suggested rewrite before applying it, and it doesn't hold up: I ran your snippets unmocked, then reverted the fix (`RawTable.tsx:45`, `if (mountedRef.current && !shouldSkipFocusRef.current)` → `if (mountedRef.current)`) to see which tests actually fail when the bug is back.

| Test | With fix | Fix reverted |
| --- | --- | --- |
| Your pointer assertion (`fireEvent.focus` + `toHaveTextContent('Row two')`) | pass | **pass** |
| Your keyboard assertion (`toHaveTextContent('Row one')`) | pass | **pass** |
| Same pointer test tightened to `toHaveFocus()` | **fail** | pass |
| Current mocked pointer test | pass | **fail** |

Why:

1. `fireEvent.focus` doesn't move DOM focus in jsdom — `document.activeElement` stays `<body>`, and `toHaveTextContent` matches body's aggregate text, so that assertion passes unconditionally.
2. Tightening it to `toHaveFocus()` *inverts* the guard — green only when the bug is present. Without the fix, `focusFocusable` focuses the first `[tabindex="0"]` node, which in a single-button cell is the button that was just focused.
3. Underlying reason: in this fixture the focused element *is* the `[tabindex="0"]` node once the index settles, so `focusFocusable` refocuses the same element and the regression is invisible to `activeElement`. The real symptom — index reverting to `(0, 0)`, header cell focused, browser scrolling it into view — needs layout and scroll, which jsdom doesn't have.

That said, your point does have a working form. A cell with **two** buttons makes real focus observable without any mock, because `focusFocusable` grabs the first `[tabindex="0"]`:

```tsx
<RawTr>
  <RawTd>
    <button type="button">Select row two</button>
    <button type="button">Edit row two</button>
  </RawTd>
</RawTr>
```

```tsx
it('does not move the roving focus when a cell reports focus through `setTableValues`', () => {
  render();
  act(() => {
    screen.getByRole('button', { name: 'Edit row two' }).focus();
  });

  expect(screen.getByRole('button', { name: 'Edit row two' })).toHaveFocus();
});
```

Verified: passes with the fix, fails with it reverted. `act()` is needed because a real `.focus()` fires `setIsActive` outside act.

I've left the mock in place here — it's a working regression guard and the PR is otherwise ready — but happy to swap in the two-button fixture, either on this PR or a follow-up. Your call.

_Reply drafted by AI, reviewed by @Adzouz._

giu1io (approved):
Awesome job, thanks 🙏

PR #2044, opened 2026-08-09
feat(primitives): add modal prop to Combobox Root

### LINKED ISSUE #1032: [feat]: Combobox from ui-primitives

Please add "modal" prop to Root of the Combobox primitive which exist on Radix primitives

Because radix-ui has these props, they give you more control over the component.

```
The modality of the popover.
When set to true, interaction with outside elements
will be disabled and only popover content will be visible to screen readers.
```

gaurav-init:
## What

Adds a `modal` prop (default `true`) to `Combobox.Root` in the primitives package, mirroring the same prop on Radix's Popover and Select primitives.

Closes #1032.

## Why

When `modal` is `true` (the current default and the backward-compatible behavior):
- `hideOthers` aria-hides everything outside the trigger + content — screen readers see only the combobox
- `RemoveScroll` prevents page scrolling while the listbox is open
- `FocusScope` traps keyboard focus inside the trigger

These defaults are correct for most cases, but can conflict when:
- The combobox is placed inside a dialog that already manages aria-hiding and focus trapping
- The product needs pointer events outside the listbox to remain active (e.g. a floating combobox over a canvas)

Setting `modal={false}` disables all three behaviors so the parent context stays in control.

## Changes

- `Combobox.tsx`: `modal` added to `RootProps` (with JSDoc), `ComboboxContextValue`, and the `Combobox` component. `hideOthers`, `FocusScope trapped`, and `RemoveScroll` are each gated on `context.modal`.
- `Combobox.test.tsx`: three new tests covering open, Escape dismiss, and item selection with `modal={false}`.

## Verification

- All 50 existing unit tests pass unchanged (default `modal={true}` behavior is unaffected).
- 3 new tests for `modal={false}` pass.
- `oxlint` and `tsc --noEmit` clean.

### DIFF
--- a/.changeset/combobox-modal-prop.md
+++ b/.changeset/combobox-modal-prop.md
@@ -1,0 +1,16 @@
+---
+'@strapi/ui-primitives': minor
+---
+
+**`Combobox`**: add `modal` prop to `Root`
+
+The `Combobox.Root` component now accepts a `modal` prop (default `true`).
+
+When `modal={false}`, the three modal-mode behaviours are disabled:
+- `aria-hidden` is not applied to elements outside the trigger and content
+- Page scroll is not locked while the listbox is open
+- Focus is not trapped inside the trigger
+
+This is useful when the combobox is placed inside a dialog or other container that already manages focus trapping and aria-hiding, or when pointer events outside the open listbox must remain active.
+
+The default (`modal={true}`) is unchanged, so existing usage is unaffected.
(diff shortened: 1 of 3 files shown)


HichamELBSI (approved):

gaurav-init:
@HichamELBSI thanks for the approval!

Quick note on the failing `chromatic` check — the log ends with:

> `Error: ✖ Missing project token`
> `Set your project token as the CHROMATIC_PROJECT_TOKEN environment variable`

This isn't a snapshot regression; it's the standard GitHub Actions behavior where repository secrets aren't exposed to workflows triggered from a fork PR, so `CHROMATIC_PROJECT_TOKEN` is unset when this branch runs. Re-running the workflow with `pull_request_target` or from a branch on the upstream repo should produce real snapshots, and the diff should be additive (new stories/behavior only, no existing story visuals change — this only adds an optional `modal` prop with the same default behavior as before).

Happy to rebase or tweak anything if it helps. Otherwise ready to merge whenever suits.

PR #2050, opened 2026-08-21
[Next] Initialise Tailwind + shadcn

ricardofrancoli:
### What does it do?

It adds Tailwind CSS v4 and shadcn at a new import path, `@strapi/design-system/next`. The old import path does not change, and both systems work on the same page.

The package ships two stylesheets, each with a different role:
- `styles.css` is the prebuilt stylesheet. It holds the compiled utilities of this package.
- `theme.css` holds the design tokens. It ships uncompiled, from `src`. It includes both semantic colours and overwritten scale steps: the 62.5% root font size rule makes 1rem equal 10px, so each type and container step is the stock Tailwind value multiplied by 1.6. `--text-2xs` is new, because Tailwind has no `2xs` step and the old system needs one.

`DesignSystemProvider` writes the dark class on the root element, so the new components follow the theme. A `useColorScheme` can be used for an application that uses only the new components.

### Why is it needed?

First step towards a new Design System!

### How to test it?

```
yarn install
cd packages/design-system
yarn build
yarn test:unit
yarn test:ts
```

To check the consumer path, in an application that runs Tailwind v4:
1. In the application's entry, import the prebuilt stylesheet. It styles the components of this package, and it declares the colour custom properties that the semantic classes read: `import '@strapi/design-system/next/styles.css';`
2. In the application's own Tailwind entry, reference the tokens:
    ```
    @import 'tailwindcss';
    @reference '@strapi/design-system/next/theme.css';
    ```
3. Write `class="bg-primary text-2xs p-4"` in the application's own source. All three compile.
4. Remove the `@reference` line and rebuild. `bg-primary` and `text-2xs` stop compiling, while `p-4` still compiles. That difference is the gap `theme.css` closes.

### Related issue(s)/PR(s)

Issue EE-149


packages/design-system/src/next/components/ui/button.test.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+
+import { Button } from '@strapi/design-system/next';
+import { render, screen } from '@testing-library/react';
+
+describe('Button', () => {
```
valfur03, line 7: I feel like both of those are testing React more than the actual button. Not blocking though, I don't think there is much to test on a button.

packages/design-system/src/next/lib/utils.ts
valfur03, line 1: What do you think about exporting this function from the design-system package? Given that it almost became the norm to use it paired with Tailwind classes, it would prevent us and any third-party package from re-declaring it in their own codebase.

packages/design-system/package.json
valfur03: Claude mentioned to replace this `sideEffects` entry to:

```suggestion
  "sideEffects": ["**/*.css"],
```

I think it's used by webpack, which we don't don't use in this repo, so this would not break our current setup, it would only be for correctness.

packages/design-system/src/utilities/DesignSystemProvider.tsx
```diff
@@ -4,6 +4,7 @@ import { DefaultTheme, ThemeProvider } from 'styled-components';
 import { LiveRegions } from '../components/LiveRegions';
 import { createContext } from '../helpers/context';
 import { useIsomorphicLayoutEffect } from '../hooks/useIsomorphicLayoutEffect';
+import { useColorScheme } from '../next/hooks/useColorScheme';
```
valfur03: It feels quite odd to import a hook from the `/next` directory inside the production code no?

The solution I would consider is to have a second `DesignSystemProvider` inside the `next/` directory so that the production one (this one) works independently of the experimental path, but experimental one still has access to this provider.

packages/design-system/src/next/components/ui/button.test.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+
+import { Button } from '@strapi/design-system/next';
```
valfur03: 💅 nitpick

```suggestion
import { Button } from './button';
```

packages/design-system/src/next/components/ui/button.tsx
```diff
@@ -0,0 +1,58 @@
...
+    defaultVariants: {
+      variant: 'default',
+      size: 'default',
+    },
+  },
+);
+
+// The shadcn output assumes React 19, where ref is a plain prop.
+// We add this `React.forwardRef` wrapper in order to keep this functionality. It can be removed at React 19
+const Button = React.forwardRef<HTMLElement, ButtonPrimitive.Props & VariantProps<typeof buttonVariants>>(
```
valfur03:
```suggestion
const Button = React.forwardRef<HTMLButtonElement, ButtonPrimitive.Props & VariantProps<typeof buttonVariants>>(
```

packages/design-system/package.json
```diff
@@ -43,18 +69,26 @@
     "@strapi/ui-primitives": "2.2.4",
     "@tanstack/react-virtual": "^3.10.8",
     "@uiw/react-codemirror": "4.22.2",
+    "class-variance-authority": "0.7.1",
+    "clsx": "2.1.1",
     "lodash": "4.18.1",
-    "react-remove-scroll": "2.5.10"
+    "lucide-react": "1.31.0",
+    "react-remove-scroll": "2.5.10",
+    "tailwind-merge": "3.6.0"
```
valfur03, line 77: I noticed that react and styled-components are declared as `peerDependencies`, so I am wondering whether it should be the case for some of the new ones (like `lucide-react`).

I don't think the ones in `devDependencies` are important though, as they won't be shipped to users' packages.

packages/design-system/README.md
```diff
@@ -36,6 +36,20 @@ export default App;
...
+
+Import the stylesheet yourself. No JavaScript import brings it, and no future change may bring it.
+
+```jsx
+import '@strapi/design-system/next/styles.css';
+import { Button } from '@strapi/design-system/next';
+```
+
+The package builds this stylesheet, so you need no CSS tools of your own.
+
```
valfur03, line 54: I feel like we could add a warning callout here, like:

```md
> [!WARNING]
> This path is experimental …
```

valfur03 (commented):
Nice one! I have some minor comments that should be easily fixable.

On a more general note, I'd raise only those two concerns:
- I feel like this PR adds a lot of comments, and some of them are too verbose. An non-exhaustive list:
  - For the new Button component, given that all components will have the same issue, I don't think we should document that here
  - `useColorScheme`, the comment is as big as the function itself, which is already pretty straightforward
  - CSS files
  I do agree that it could make sense to keep track of some decisions (so that AI agents can benefit from key context), but in this case, maybe it might be worth having a single documentation file referencing these inside the `next/` directory? What do you think?
- As a follow up task, I think it might be nice to install some additional dev tools, like linting for TailwindCSS

packages/design-system/src/next/components/ui/button.test.tsx
ricardofrancoli: Agreed that it's a bit borderline whether to test or not. I thought about keeping them due to the `forwardRef` wrapper: it's the one thing we add to the shadcn output. We can rethink them in future versions. Wdyt?

packages/design-system/src/next/lib/utils.ts
ricardofrancoli: Good shout, changed!

packages/design-system/package.json
ricardofrancoli: Added 😄

packages/design-system/src/next/components/ui/button.test.tsx
ricardofrancoli: Changed

packages/design-system/src/next/components/ui/button.tsx
ricardofrancoli: Changed

packages/design-system/src/utilities/DesignSystemProvider.tsx
ricardofrancoli: Good shout, it's an oversight from my part. I've modified it in [this commit](https://github.com/strapi/design-system/pull/2050/changes/1190aa79013dcceb81701299dadaa93d5ebb65f6).

I took a smaller route than a second provider: the hook now lives in `src/hooks/useColorScheme.ts` with the other hooks, and `/next` re-exports it.

The provider imports from `src/hooks`, so nothing in `src/` imports from `src/next/` any more. The `/next` tree already depended on the root for the `ColorScheme` type. A second provider would need the same theme type from the root, and would give us two providers to keep in sync. At DS 3.0 the old tree goes and the hook moves into `/next` for good.

Lemme know what do you think about this direction 👀

packages/design-system/package.json
ricardofrancoli: The rule followed here is that a package goes in `peerDependencies` only when a second copy of it breaks something. Otherwise it goes in `dependencies`, like the new ones you see here. `peerDependencies` have to then be installed by the consumers.

`react` and `styled-components` are peers because a second copy breaks them, e.g. a second React copy breaks hooks and a second `styled-components` copy reads a different `ThemeProvider` context and loses the theme.

On the other hand, these new libraries are stateless, simpler and/or self-contained. E.g. `lucide-react` renders SVGs, `clsx` and `tailwind-merge` are pure functions, so a duplicate costs a (slightly) larger bundle size at most, and we do not want to ask a user to install five packages by hand in order to use the DS.

ricardofrancoli:
> Nice one! I have some minor comments that should be easily fixable.
> 
> On a more general note, I'd raise only those two concerns:
> 
> * I feel like this PR adds a lot of comments, and some of them are too verbose. An non-exhaustive list:
>   
>   * For the new Button component, given that all components will have the same issue, I don't think we should document that here
>   * `useColorScheme`, the comment is as big as the function itself, which is already pretty straightforward
>   * CSS files
>     I do agree that it could make sense to keep track of some decisions (so that AI agents can benefit from key context), but in this case, maybe it might be worth having a single documentation file referencing these inside the `next/` directory? What do you think?
> * As a follow up task, I think it might be nice to install some additional dev tools, like linting for TailwindCSS

Thanks for the comments! I've pushed a few commits that should fix your main concerns. Let me know what you think 🙏

packages/design-system/package.json
```diff
@@ -43,18 +69,26 @@
     "@strapi/ui-primitives": "2.2.4",
     "@tanstack/react-virtual": "^3.10.8",
     "@uiw/react-codemirror": "4.22.2",
+    "class-variance-authority": "0.7.1",
+    "clsx": "2.1.1",
     "lodash": "4.18.1",
-    "react-remove-scroll": "2.5.10"
+    "lucide-react": "1.31.0",
```
Adzouz, line 75: `lucide-react` is a runtime `dependency`, but nothing imports it yet — `git grep lucide-react` outside `package.json` returns no hits. Every consumer installs it for nothing until the first icon-using component lands.

`components.json` sets `iconLibrary: "lucide"`, so it'll be needed soon and this may well be deliberate. If so, ignore me. Otherwise `devDependencies` until the first component actually pulls it in.

_Comment generated by AI, reviewed by @Adzouz._

packages/design-system/src/hooks/useColorScheme.ts
```diff
@@ -0,0 +1,12 @@
+import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
+
+import type { ColorScheme } from '../themes';
+
+/**
+ * Toggles the `dark` class on the root element of the document
+ */
+export const useColorScheme = (colorScheme: ColorScheme) => {
```
Adzouz, line 9: Two edges worth a JSDoc line, since this is exported publicly from `/next`:

1. **No cleanup on unmount** — the `dark` class outlives the hook. Unmount a provider and the document stays dark.
2. **One global class, multiple callers** — two `DesignSystemProvider`s with different themes on the same page fight over `document.documentElement`, and whichever effect runs last wins.

Neither matters for the CMS (one provider, whole page), so I'd not change the behaviour. But a consumer rendering a light-themed subtree will find this surprising, and the current comment ("Toggles the `dark` class on the root element") doesn't hint at it. Documenting the scope costs less than the bug report.

_Comment generated by AI, reviewed by @Adzouz._

turbo.json
```diff
@@ -25,8 +25,14 @@
     "test:unit": {
       "dependsOn": ["^build"]
     },
+    "@strapi/design-system#test:unit": {
```
Adzouz: Correct — `styles.test.ts` compiles Tailwind, so it genuinely needs the build. Flagging the cost rather than the correctness: this makes `test:unit` (previously the fastest feedback loop in the repo) depend on a full self-build, for the sake of one test file.

If that starts to hurt, an alternative is to split the CSS invariant tests into their own script that depends on `build`, and leave `test:unit` build-free. Not worth doing pre-emptively.

_Comment generated by AI, reviewed by @Adzouz._

packages/design-system/src/next/components/ui/button.tsx
```diff
@@ -0,0 +1,56 @@
...
+        ref={ref}
+        data-slot="button"
+        className={cn(buttonVariants({ variant, size, className }))}
+        {...props}
+      />
+    );
+  },
+);
+
+export { Button, buttonVariants };
```
Adzouz, line 56: No story for this `Button` in `docs/stories`, which means the `/next` tree starts with zero visual regression coverage — and for a Tailwind-based component whose whole surface is classes, Chromatic is the only thing that can actually catch a break.

Fine to leave for a follow-up on an init PR that nothing consumes yet. Raising it now because "add stories later" tends to mean the first `/next` regression is found by a consumer.

_Comment generated by AI, reviewed by @Adzouz._

Adzouz (commented):
Hi @ricardofrancoli, so happy to see things moving forward 🙌
I'll test and take a deeper look but in the meantime, I ran a review and left a few comments, thanks a lot 🙏

packages/design-system/package.json
ricardofrancoli: Yup indeed, this is a deliberate choice. It will be relevant once we add components that use them. I decided to keep it as is for now, and knowing it's still an experimental/early stages.

packages/design-system/src/next/components/ui/button.tsx
ricardofrancoli: Good catch. Adding the story has been deferred: I was hitting some walls when implementing it (e.g. font not rendering correctly) so I think it deserves a deeper look. It's one of the next issues in line for this repo. In any case, at the moment, the shadcn button is just the default one, so not super important to surface any regressions right now.

turbo.json
ricardofrancoli: 💡 This is actually not needed any more. The stylesheet test now compiles Tailwind from source, and the `Button` test imports `./button`, so the `/next` tests no longer read `dist`. Removed both `dependsOn: ["build"]` entries. `rm -rf dist && yarn test:ts` and `rm -rf dist && yarn test:unit` both pass.

packages/design-system/src/hooks/useColorScheme.ts
ricardofrancoli: Makes sense. Edited the comment to clarify these bits 👌

packages/design-system/src/next/components/ui/button.test.tsx
valfur03: Yes I don't think we should really spend time on that for now 👍 

I agree to keep them as is.

valfur03 (approved):
Looks good to me! 🚀

packages/design-system/src/utilities/DesignSystemProvider.tsx
valfur03: Yeah it looks good too, I just thought about the additional provider so that the production code would not need to change at all, but this is ok, it's just about adding a class so I don't think it could introduce any issue at all.

PR #2053, opened 2026-09-02
feat: add a single Tailwind consumer entry

ricardofrancoli:
### What does it do?

Adds `@strapi/design-system/next/source.css`: one entry that imports the compiled stylesheet, references the Tailwind and design system themes, then generates the utilities. Automatic scanning stays off, so the consumer adds their own `@source`.

The `dark` variant moves to `theme.css`, which the theme reference carries through to the consumer build. `tailwindcss` becomes an optional peer dependency.

### Why is it needed?

Before this, a consumer had to import `styles.css`, then add a `@reference` to `theme.css`, then set up Tailwind. Three steps, and each one is easy to get wrong.

The utilities also did not work correctly (e.g. `dark:` classes), because the package shipped no entry that generates them. This gives one import instead:

```css
@import '@strapi/design-system/next/source.css';
@source './src/**/*.{ts,tsx}';
```

Most importantly: this will allow us to wire it up to the CMS so we can start using it there.

How to test it?

1. Install the experimental release in a Vite or Next.js app that has Tailwind v4.
2. Put the two lines above in the app stylesheet.
3. Use a `/next` component and a Tailwind utility class in the same file. Both must render.
4. Add the `dark` class to the document root. The `dark:` utilities must apply.

Related issue(s)/PR(s)

Follow-up to #2050.

### DIFF
--- a/.changeset/quiet-otters-listen.md
+++ b/.changeset/quiet-otters-listen.md
@@ -3,8 +3,19 @@
 ---
 
 Add Tailwind CSS and shadcn at `@strapi/design-system/next`, with a Button as the
-first component. Import the stylesheet from `@strapi/design-system/next/styles.css`.
-The old import path does not change.
+first component. The old import path does not change.
+
+Import `@strapi/design-system/next/source.css` to get the stylesheet, the tokens and
+the `dark` variant. Then add an `@source` for your own files:
+
+```css
+@import '@strapi/design-system/next/source.css';
+@source './src/**/*.{ts,tsx}';
+```
+
+The entry keeps automatic scanning off, because Tailwind reads a CSS value in a
+styled-components template literal as a class name. Tailwind resolves the `@source`
+path from the stylesheet that holds it.
 
 `DesignSystemProvider` writes the `dark` class on the document root, so both systems
 share one theme. A `useColorScheme` hook at `/next` does the same without the
--- a/.changeset/quiet-otters-listen.md
+++ b/.changeset/quiet-otters-listen.md
@@ -13,9 +13,9 @@ the `dark` variant. Then add an `@source` for your own files:
 @source './src/**/*.{ts,tsx}';
 ```
 
-The entry keeps automatic scanning off, because Tailwind reads a CSS value in a
-styled-components template literal as a class name. Tailwind resolves the `@source`
-path from the stylesheet that holds it.
+The entry keeps automatic scanning off, because the package cannot know the file paths of your project. Tailwind resolves the `@source` path from the stylesheet that holds it.
+
+The `@strapi/design-system/next/theme.css` export is removed. If you use it with `@reference`, import `source.css` instead.
 
 `DesignSystemProvider` writes the `dark` class on the document root, so both systems
 share one theme. A `useColorScheme` hook at `/next` does the same without the
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -8,7 +8,9 @@
     "url": "git+https://github.com/strapi/design-system.git",
     "directory": "packages/design-system"
   },
-  "sideEffects": ["**/*.css"],
+  "sideEffects": [
+    "**/*.css"
+  ],
   "source": "./src/index.ts",
   "main": "./dist/index.js",
   "module": "./dist/index.mjs",
@@ -27,7 +29,7 @@
       "default": "./dist/next/index.mjs"
     },
     "./next/styles.css": "./dist/next/styles.css",
-    "./next/theme.css": "./src/next/theme.css",
+    "./next/source.css": "./src/next/source.css",
     "./dist/*": {
       "types": [
         "./dist/*",
@@ -39,7 +41,8 @@
   },
   "files": [
     "dist",
-    "src/next/theme.css"
+    "src/next/theme.css",
+    "src/next/source.css"
   ],
   "dependencies": {
     "@base-ui/react": "1.7.0",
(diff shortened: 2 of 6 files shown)


valfur03 (approved):
Nice job! 🎨

packages/design-system/src/next/source.css
```diff
@@ -0,0 +1,14 @@
...
+ * The Tailwind entry for @strapi/design-system/next
+ *
+ * A consumer imports this file and their build must process it.
+ * The bare `tailwindcss/*` imports resolve from the copy of Tailwind that build owns, which is why we declare it as a peer dependency.
+ */
+
+@import '@strapi/design-system/next/styles.css';
+@import 'tailwindcss/theme.css' theme(reference);
+@import './theme.css' theme(reference);
+/* `source(none)` keeps automatic scanning off. Otherwise Tailwind reads a CSS value in a
```
Adzouz: The comment explains `source(none)` in terms of the *DS build's* styled-components problem, but in this file it's the *consumer's* scanning that's being switched off — and the reason is different: the package can't know a consumer's file paths. Worth rewording so a reader doesn't go looking for styled-components in their own app.

The second half ("Tailwind resolves that path from the file that holds the line") is the genuinely useful bit and deserves to stay.

packages/design-system/package.json
```diff
@@ -98,7 +101,13 @@
     "@strapi/icons": "^2.0.0 || ^2.0.0-beta || ^2.0.0-alpha",
     "react": "^18.0.0",
     "react-dom": "^18.0.0",
-    "styled-components": "^6.0.0"
+    "styled-components": "^6.0.0",
+    "tailwindcss": "^4.0.0"
```
Adzouz: `^4.0.0` is wider than what actually gets shipped. `dist/next/styles.css` is compiled by this package's pinned `tailwindcss@4.3.3`, and the consumer recompiles `source.css` with their own copy — so on 4.0.x you'd merge a 4.0-shaped `@layer properties` polyfill with the 4.3.3-shaped one already baked into the dist file.

Mostly benign, but a `^4.3.0` floor costs nothing and keeps both halves of the output on the same emitter.

.changeset/quiet-otters-listen.md
```diff
@@ -3,8 +3,19 @@
 ---
 
 Add Tailwind CSS and shadcn at `@strapi/design-system/next`, with a Button as the
-first component. Import the stylesheet from `@strapi/design-system/next/styles.css`.
-The old import path does not change.
+first component. The old import path does not change.
+
+Import `@strapi/design-system/next/source.css` to get the stylesheet, the tokens and
```
Adzouz, line 8: Worth one line noting that `./next/theme.css` is no longer an export. It fails loudly (the exports map blocks the path) and only affects anyone who followed the `@reference` route from #2050 on the experimental tag, so it isn't a bump-level question — but someone on that tag will hit a resolution error and the changeset is where they'd look.

packages/design-system/package.json
```diff
@@ -8,7 +8,9 @@
     "url": "git+https://github.com/strapi/design-system.git",
     "directory": "packages/design-system"
   },
-  "sideEffects": ["**/*.css"],
+  "sideEffects": [
+    "**/*.css"
```
Adzouz, line 12: Nit: this reformat is unrelated to the PR and adds two lines of diff noise. Fine to keep if a formatter did it, just flagging that it isn't part of the change.

Adzouz (commented):
Hey @ricardofrancoli, thanks for the PR 🙌 I just have a few nits comments from my review 😊

packages/design-system/package.json
ricardofrancoli: Nice catch! Raised it to `^4.3.0`. For future reference, I tested this directly: `4.0.0` to `4.0.8` compile `source.css` without a warning but do not inline theme fallbacks for `theme(reference)` imports, so any stock utility such as `text-red-500` becomes a dead `var()`. This only becomes correct at `4.0.9` and later. Keeping `^4.3.0` as the base anyway as it's cheap and easy to adopt since the only DS consumer with Tailwind today, the CMS admin, is already on `4.3.3`, and plugins (who can adopt it eventually) run inside it.

packages/design-system/src/next/source.css
ricardofrancoli: Reworded. It now names the real reason: the package cannot know a consumer's file paths.

.changeset/quiet-otters-listen.md
ricardofrancoli: Added a line in reference to that 👌

packages/design-system/package.json
ricardofrancoli: While it is true that it's unrelated, this was a formatting issue that dragged from the previous `next` PR, uncaught by the CI. It is reformatted here 😄

Adzouz:
Thanks for the quick turnaround @ricardofrancoli — all four points are addressed 🙌

- **`source.css` comment** — now names the real reason (the package can't know a consumer's file paths) and keeps the `@source` resolution note. 👌
- **`tailwindcss` peer floor** — `^4.3.0`. Really good find on 4.0.0–4.0.8 not inlining theme fallbacks for `theme(reference)` imports; a stock utility silently becoming a dead `var()` is exactly the kind of thing nobody would have traced back to a peer range. Worth remembering when we bump the CMS.
- **Changeset** — the removed `./next/theme.css` export is noted, with the migration hint. 👍
- **`sideEffects` reformat** — fair enough, it's drift from #2050 that CI missed. Happy to have it fixed here.

I pulled the branch and checked the result rather than just reading the diff:

- `build`, `test:ts`, `lint` all clean, `/next` unit suites 9/9
- compiled `source.css` with `@tailwindcss/node` the way a consumer's build would: `dark:bg-red-500` resolves to `.dark\:bg-red-500:is(.dark *)`, and the theme fallbacks are intact (`var(--color-red-500, oklch(63.7% 0.237 25.331))`, `var(--spacing, 0.4rem)`)
- the DS variables are emitted once, from the compiled `dist/next/styles.css`, and consumer utilities land after the DS ones so a consumer class wins on a tie

Tiny one, purely cosmetic, ignore if you like: the changeset reflow left that first paragraph as a single long line while the rest of the file wraps at ~85 chars.

LGTM ✅

Adzouz (approved):

PR #2054, opened 2026-09-04
fix: experimental release working with the `yarn changeset pre exit` command

cache-your-dreams:
### What does it do?
Hopefully fixes experimental releases using Trusted Publishing

### Why is it needed?
So we can release experimental versions of the Design System again

### How to test it?
Once merged, run the experimental action

### DIFF
--- a/.github/workflows/experimental.yml
+++ b/.github/workflows/experimental.yml
@@ -40,7 +40,18 @@ jobs:
       - name: Build
         run: yarn build --filter=\!@strapi/design-system-docs
 
+      - name: Check changeset pre-release mode
+        id: pre-release
+        run: |
+          if [ -f .changeset/pre.json ]; then
+            echo "active=true" >> "$GITHUB_OUTPUT"
+            echo "tag=$(node -p "require('./.changeset/pre.json').tag")" >> "$GITHUB_OUTPUT"
+          else
+            echo "active=false" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Exit changeset pre-release mode
+        if: steps.pre-release.outputs.active == 'true'
         run: yarn changeset pre exit
 
       - run: ./scripts/pre-publish.sh --yes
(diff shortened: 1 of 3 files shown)


ricardofrancoli (commented):
Tested this locally and can confirm the guard works. On a branch without `pre.json` it sets `active=false` and the job skips both steps. However I do wonder if the better move is to completely remove the steps instead of guarding them...

These steps come from when `main` was in pre mode for the 2.0 release candidates. Changesets refuses experimental publishes in pre mode, so the job had to exit pre mode to publish and enter again afterwards. My guess is that, after the rc line ended and `pre.json` was deleted, these steps stayed behind. 🤔

This workflow only does experimental snapshots (`0.0.0-experimental.<sha>`), which don't use pre mode, and no branch has `pre.json` today. So the steps are dead code today. If we ever need this flow (releasing experimentals while in pre-mode) we can revisit.

What I have in mind instead:
- Remove both pre steps and the guard
- Add a push trigger on `next`: we will merge experimental work there for the next few weeks/months

There are also some misc issues in the script itself, e.g. it formats the experimental version as `0.0.0-0.0.0-experimental.<sha>` (double triple 0).

Leaving this guard behind is also not catastrophic though, happy to discuss. What do you think? I can tackle those in EE-175 if you'd prefer!

Bassel17:
I think we should simplify this workflow around its actual goal: publish all public packages from a chosen commit as `0.0.0-experimental.<sha>` under the npm `experimental` tag.

Changesets is useful for normal releases semver, changelogs, and prerelease cycles but adds the wrong behavior here:

- `branch` input is required but never used by checkout.
- No pending changeset means no experimental package is produced.
- `VERSION` is treated as a snapshot tag, producing `0.0.0-0.0.0-experimental.<sha>-<timestamp>` instead of the requested exact version.
- Exit/re-enter prerelease steps are leftover from the 2.0 RC cycle and only mutate the temporary runner.
- Per-branch concurrency can let multiple runs update the same npm tags simultaneously.

I suggest replacing this with a direct publishing flow: checkout requested ref, resolve HEAD SHA, build all packages, set all package and internal dependency versions to `0.0.0-experimental.<sha>`, publish in dependency order using Trusted Publishing, make retries idempotent, then verify all `experimental` tags point to the same version.

The guard in this PR fixes one immediate failure, but it keeps obsolete complexity. I would not merge this as the final experimental-release fix.

PR #2055, opened 2026-09-10
fix: experimental releases fail at changeset pre exit and publish nothing

ricardofrancoli:
### What does it do?

The experimental workflow stops at changeset pre exit, because it finds no `.changeset/pre.json`. 

The changesets snapshot path has three more faults:
- It publishes nothing when no changeset is pending
- It refuses to run in pre mode
- It writes a wrong version string (`0.0.0-0.0.0...`)

With this PR, we remove changesets from the experimental path. The possible flows:
- A push to `next` publishes the three public packages at `0.0.0-experimental.<sha>` under the `experimental` dist-tag
- The "Run workflow" button starts the same publish from any branch
 
One Node script stamps the version, rewrites every internal `@strapi/*` dependency to that exact version, and publishes each package with npm. A rerun of the same commit skips the packages that are already on npm and publishes the rest. Before any publish, the script also reads where the `experimental` tag of each skipped package points. When a package is missing and a skipped package keeps its tag on another commit, the run fails and publishes nothing. So the tag never splits across two commits. If that commit still needs a release, a new push to `next` gives a new version. The concurrency key no longer includes the branch, so runs of this workflow go one at a time. Two runs from different branches cannot publish at the same time and mix the experimental dist-tag between two commits.

The stable release keeps changesets and doesn't change.

*Before the first run*: an npm org admin must bind `strapi/design-system` with `experimental.yml` as a trusted publisher on `@strapi/design-system`, `@strapi/icons` and `@strapi/ui-primitives`.

*After the merge*: merge `main` into `next`, so the push trigger reaches `next`.

### Why is it needed?

So we can easily release experimental snapshots both when pushing to `next` (which is going to happen a lot in the upcoming months) or on demand, via GitHub Workflows.

### How to test it?

1. Run the workflow from this branch with the "Run workflow" button. The run is green and the three packages exist on npm at `0.0.0-experimental.<sha of this branch>`
2. Run it again on the same commit. The run reports three skips and stays green
3. `npm view @strapi/design-system@experimental dependencies` shows `@strapi/ui-primitives` at the same version

### Related issue(s)/PR(s)

- Supersedes #2054
- Closes EE-175


scripts/publish-experimental.mjs
```diff
@@ -0,0 +1,83 @@
...
+  console.log(`stamped ${manifest.name}@${version}`);
+}
+
+// npm, not yarn: npm needs no secret with trusted publishing
+for (const { dir, manifest } of packages) {
+  const spec = `${manifest.name}@${version}`;
+
+  if (isOnNpm(spec)) {
+    console.log(`skipped ${spec}, it is on npm already`);
+    continue;
```
cache-your-dreams: [P2] Retries can leave experimental tags pointing to different commits.
The check only asks whether the version exists. It doesn’t check where the experimental tag points:
1. A publishes primitives, then fails.
2. B publishes all three packages, moving their tags to B.
3. Retrying A finds primitives A already exists and skips it—its tag stays on B.
4. Icons A and design-system A don’t exist, so npm publish --tag experimental publishes them and moves their tags to A.

The result is primitives tagged B, with icons and design-system tagged A. Skipping the existing version leaves its tag unchanged; the subsequent publishes move the other tags.
Allegedly.

ricardofrancoli: Correct, the skip checked only the version, not the tag. The script cannot move a tag onto an existing version. Trusted publishing with OIDC permits `npm publish` only, not `npm dist-tag add`.

So the script now reads two things before any publish: which packages exist at this version, and where the `experimental` tag of each of those points. It exits 1 and publishes nothing when a package is missing and a skipped package keeps its tag on another commit. The retry in your scenario stops there.

A retry of the same commit with no newer run in between still skips what landed and publishes the rest. So the one-click repair stays. A rerun of an older commit that published all three stays green, because it publishes nothing and can split nothing.

Nothing is broken in the mixed case: the three tags still point at the newer commit. If that older commit still needs a release, a new push to `next` gives a new version.

I tested this with dry runs, and the full, skip and mixed cases all behave as described.

valfur03 (approved):
Very nice! 👏

cache-your-dreams (approved):
LGTM

PR #2057, opened 2026-09-18
chore(next): remove lucide-react

ricardofrancoli:
### What does it do?

Removes `lucide-react` from `@strapi/design-system` `dependencies` and drops its `yarn.lock` entry. `components.json` is not changed.

Note: `yarn.lock` also picks up one stale line, the design-system peer range `tailwindcss: ^4.0.0 → ^4.3.0`. #2053 bumped it in `package.json` without refreshing the lockfile.

### Why is it needed?

`lucide-react` was added in the shadcn init (#2050) but no file imports it yet. As a runtime dependency it adds about 35MB to every Strapi app that installs a release built from `next`, with or without the feature flag (see strapi/strapi#27550).

`iconLibrary` in `components.json` stays `lucide` for now. It is a CLI hint only (e.g. for when we run `shadcn add {{ component }}`) and no icon dependency is installed. Replacement library is to be discussed/decided, but it's likely going to be Phosphor, added as a `devDependency` with some mechanisms in order to wire them up to the components that need icons.

### How to test it?

- `grep -rn lucide packages --include='*.json' --include='*.ts' --include='*.tsx'` returns only `components.json:12`
- `grep -c lucide yarn.lock` returns 0
- `yarn build --filter=@strapi/design-system` passes

### Related issue(s)/PR(s)

- #2050 introduced the dependency
- #2053 left the stale `tailwindcss` peer range in the lockfile
- strapi/strapi#27550 raised the install-cost concern


Adzouz (approved):
LGTM, thanks! 🙏

HichamELBSI (approved):

